PR #2, opened 2016-05-18
Create readme

briri:
Added an initial README file

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,15 @@
+h2. DMP Roadmap
+
+h4. Current Release
+v.0.1.0
+
+h4. Summary
+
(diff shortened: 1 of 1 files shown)


mxsstrong:
I see that we are testing here, but we should ensure that merges have helpful messages attached to them.  For example merge <source branch> into <target branch> .  Can you add a message to these merges?

Marisa

briri:
Agreed. 

The message was in part due to 'squashing' the pull request into one commit. Github puts a link to the original commit messages, #2 and #3, in this case. It will be good practice to explain what is being merged to which branch and to re-summarize the changes

PR #5, opened 2016-06-09
Added README content, better seed data, and made a few mods to fix some minor bugs

briri:
Please hold off on merging this until next week when the UC3 team is in Glasgow so that we can walk through this process together to make sure we're all on the same page.


briri:
Conflicts exist. Going to reject this PR and resubmit

PR #8, opened 2016-06-27
Development

briri:
- Changed ownership terminology on README.rdoc, README.md so that it lists both the UC3 and DCC organizations.
- Updated db/seeds.rb so that the seed data is more generic and no longer specific to DCC.
- Fixed some JS issues created during the migration to Rails 4.0
- Uncommented the attr_accessible lines in models/user.rb so that changes are updating the db
- Added missing fontawesome web font to repo
- Fixed issue with undefined variable in projects.js
- Fixed db migration file that referenced an old AdminUser model that no longer exists
- Commented out failing tests
- Added travis.yml and README icon and then updated rakefile to run tests by default


briri:
Ray, Jimmy and Marta, please review and merge onto master if all looks good to you. 

I'm thinking we should have Stephanie and Sarah weigh in on the wording of the README files.

briri:
Added a travis.yml file and commented out some of the existing tests that weren't passing. We can fix these as we work on individual pieces of the system I guess

PR #21, opened 2016-08-12
Updated Gemfile to use Rails 4.2.7, fixed some minor issues with locale initializer

briri:
- Update Gemfile to specify >= Ruby 2.2.2
- Update Rails version to 4.2.7 because of XSS vulnerability: https://groups.google.com/forum/#!topic/rubyonrails-security/I-VWr034ouk
- Rebuilt schema.rb after running latest db migrations
- Updated locale.rb initializer to check for existence of default_language field and whether or not a default language has been defined.


briri:
Also added languages to the seeds.rb file

PR #49, opened 2016-10-18
Bug fixes - Plan and Template pages and Devise authentication

briri:
- Fixed JS issues with create plans page
- Fixed issue with exporting as a PDF
- Moved wicked_pdf config settings from application.rb into an initializer
- Removed devise changes for redirecting users after session timeout because it was causing login issues in certain circumstances
- Removed irrelevant 'guidance' button from template edit page
- Fixed issue with tinymce initialization on edit template page
- Selecting 'Make big changes' when editing a template no longer produces an error. It creates a new version as expected
- Clicking 'Make small changes' when editing a template now properly displays the edit version details section
- User is now redirected to the main landing page after requesting a password reset
- User is now redirected to the main landing page when trying to access a secure page when they are not authenticated


briri:
@raycarrick-ed and @vyruss, here are the bug fixes associated with the last sprint. Please accept and merge them into master if all looks good.

PR #68, opened 2016-11-10
ORCID functionality added to user profile page

briri:
Added the ability to attach the user's account to their ORCID using oauth2.


briri:
This PR also involved rebasing the development branch to what we currently have in master

PR #71, opened 2016-11-21
Latest changes from DCC Roadmap/master

raycarrick-ed:


briri:
The tests for this are failing because the branding.yml looks like it was moved to a local directory on your server. Can you please add it back in and handle instance specific configuration in a separate branch?

If it's helpful, you can take a look at https://github.com/CDLUC3/roadmap/tree/deploy for an example. We're using Capistrano with some steps that bring in our customized branding and handle other instance specific configuration.

PR #82, opened 2017-01-03
Development

briri:
Added unit and functional tests for the data model objects that will not be changed by the upcoming DB refactor


briri:
Canceling this one. Will resubmit once tests are passing.

PR #91, opened 2017-01-24
Added visibility to the Plans model and initial tests

briri:


briri:
meant to submit to DCC fork first

PR #92, opened 2017-01-26
Date question format

benjaminfaure:
Some remarks : 
 - When lauching the server, I had an error on Ability::Cancan, the gem was not included in the Gemfile, hope it's ok.
 - For now, the date format of the datepicker is put directly in application.js, there is maybe a better place for it.
 - The datepicker uses the default theme of jquery-ui, I did some customisation in our instance. If you are interested, I can include the CSS (which match our colors).
- I forked the code a while ago, so it seems the merges from the head repository are included in this pull request, hope it is ok.


briri:
Thank you Benjamin. Your changes look good. We will be sure to remove that Ability model on our end and then remove the need for the gem afterward.

I think placing the JS in application.js is appropriate here since the date selector jquery widget

benjaminfaure:
Thank you.

What I meant about the date format was that, right now the datepicker uses the UK/European date format (dd-mm-yy) and if people want to change it, they have to modify the application.js. 
If thought that this information could be moved to the locales.

I checked and I could use the [I18n-js](https://github.com/fnando/i18n-js) that is included in the Gemfile but it seemed overly complex for something so trivial.

It can be considered if, in the future, we need to use more locales in the js.

vyruss:
@raycarrick-ed @briri is this safe to merge into [`development`](https://github.com/DMPRoadmap/roadmap/tree/development) ?

briri:
Yes I think so. We will just have to remove the CanCan dependency afterward. Benjamin was working off of master which did not have all of our latest development branch changes.

If I remember correctly, he only added it back in because he was getting an error message when trying to run the app due to the presence of the Ability.rb model

vyruss:
@paulwalk mentioned this link: [making Rails internationalisation available to JS components](https://github.com/fnando/i18n-js)

PR #143, opened 2017-03-02
added missing UsersPerm and OrganisationType models

briri:
Fix for 3rd migration step. 2 models were missing but referenced in the migration files.


briri:
Found an additional issue:
```shell
StandardError: An error has occurred, all later migrations canceled:

Unknown primary key for table users_perms in model UsersPerm.
```

Issue #187, opened 2017-03-20
Templates: edit, preview   


stephaniesimms:
1. clicking Edit or View takes me to Template details (screenshot below) where nothing on page works. Most buttons and Generic DMP tab throw "We're sorry" error message. "Edit template details" button takes me to a page (second screenshot) where I still cannot edit anything

![template-errors](https://cloud.githubusercontent.com/assets/12703638/24120715/cdfc9a46-0d72-11e7-9c72-08ce3312592c.png)
![template-loop-errors](https://cloud.githubusercontent.com/assets/12703638/24120723/d23c5178-0d72-11e7-8974-caaa5eaba411.png)

briri:
The following error appears in the log when you click on the 'Generic DMP' phase in the example above:
```shell
App 660 stdout: NoMethodError (undefined method `template' for #<Phase::ActiveRecord_Relation:0x0055d27ce737f0>):
App 660 stdout:   app/policies/phase_policy.rb:17:in `admin_show?'
App 660 stdout:   app/controllers/phases_controller.rb:103:in `admin_show'
```

briri:
related to the following tickets:
- #191 
- #192 
- #193 
- #194

stephaniesimms:
tested and appears to be fixed

PR #190, opened 2017-03-21
Bug fix for policy issue and fixes for tests

briri:
- Fixed the policy issue reported in #187. This was actually an issue with the way we were calling Phase.eager_load. 2 eager_load methods were added to Plan that override the existing ActiveModel eager_load method. I think this caused confusion.
- Fixed issue with no default phase being generated when creating a new template #191 
- Updated a bunch of the tests so that they use FastGettext


briri:
This one fixes the specific policy error we saw earlier. It was actually an issue with Phase.eager_load(params[:id]) on the plans_controller.rb. Since Phase doesn't have its own override of that ActiveRecord method, I just updated it to work with the active record version.

The other stuff is just me switching some of the tests over from I18n to FastGettex

PR #203, opened 2017-03-24
Moved Managing Org definition to branding.yml and fixed tests

briri:
- Switched our old Managing Org definition from the locale.yml magic strings section over to the branding.yml file. **The abbreviation for the org in this file MUST match one in the DB going forward.** I added a check for the test/configuration_test.rb.
- Added branding.yml to the .gitignore and changed the name of branding.yml to branding_example.yml
- Finished fixing broken tests. Most were caused by not knowing the Managing Org or by the switch from I18n and moving locale from the path to the session

Adding @jollopre to help review the I18n changes here. I commented out 2 tests in application_controller_test.rb because when I change the user's language it doesn't update FastGettext.locale (it seems to work ok when running the app so it may be something weird with the devise test helpers)


jollopre:
@briri, when the user signs in the FastGettext.locale is called at https://github.com/DMPRoadmap/roadmap/blob/development/app/controllers/sessions_controller.rb#L25

Could you check whether or not the Devise::Test::IntegrationHelpers is calling the action create at sessions_controller.rb? If so, this will need further investigation. Perhaps the update of the session variable is ignored and consequently the FastGettext.locale never changes.

Issue #221, opened 2017-04-03
Guidance - Question dropdowns are confusing


briri:
The Template-Phase-Section-Question dropdowns on the Guidance page are confusing. They do not distinguish between different versions of a template and do not distinguish between funder templates and the org's customisation of those templates.

![screen shot 2017-04-03 at 11 01 01 am](https://cloud.githubusercontent.com/assets/1204467/24623415/2f0497e0-185d-11e7-8576-a2c56549bfe6.png)

briri:
The code behind the dropdowns is also generating errors in the logs
```shell
App 11762 stdout: F, [2017-04-03T10:59:31.770282 #11762] FATAL -- : 
App 11762 stdout: NoMethodError (undefined method `versions' for #<Template:0x0055dfe1e58a28>):
App 11762 stdout:   app/controllers/guidances_controller.rb:34:in `update_phases'
App 11762 stdout: 
```

vyruss:
Will a display format like _$TEMPLATE - $ORG cust._ do the trick?
e.g. _Horizon2020 DMP - University of Marbella cust._
And of course you should never see non-current-version stuff.

sjDCC:
Will this dropdown not be disappearing anyhow since we'll no longer have guidance by question, just annotations? Would be better if users could annotate the question directly rather than adding guidance separately via this interface. It's always been difficult for users to find relevant questions by going through the dropdowns. 
Having multiple entries for each template seems to be a bug btw - that's not usual

briri:
I agree. I'll go ahead and remove these dropdowns from the page.

briri:
@stephaniesimms and @sjDCC this fix has been deployed to the server. The dropdowns were removed from the guidance edit/add pages.

stephaniesimms:
This looks good sans dropdowns.

sjDCC:
Reopening this as it can't be tested at all now. Guidance page throwing an error altogether

![capture](https://cloud.githubusercontent.com/assets/6169940/24914105/1fb8c706-1ecb-11e7-9715-5659074d43f0.PNG)

briri:
Missed this when removing the dropdowns from the guidance page. The tests did not catch it because our test data doesn't contain any guidance with a question_id.

I did a quick check and there are 120 records that have the old question_id populated and no tie in to a theme.
```sql
select * from guidances where question_id is not null;
```
[...]

Issue #212, opened 2017-03-29
Add functional tests for controllers


briri:
Add basic functional tests for our controllers that test each action for both success and failure.

- [x] answers
- [x] api_projects
- [x] application (base controller)
- [x] contacts (DELETED - Does not get used)
- [x] confirmations
- [x] guidance_groups
- [x] guidances
- [x] home
- [x] notes
- [x] omniauth_callbacks
- [x] omniauth_shibboleth_request
- [x] orgs
- [x] passwords (SKIPPING - Devise gets mocked during testing so difficult to test these overrides)
- [x] phases
- [x] plans
- [x] projects (SKIPPING - This controller is no longer in use)
- [x] questions
- [x] registrations
- [x] roles
- [x] sections
- [x] sessions
- [x] static_pages
- [ ] suggested_answers
- [x] templates
- [ ] token_permission_types
- [ ] user_identifiers
- [ ] users

briri:
Found a nasty issue with TravisCI not liking the way that Perm defined its constants. Switched constants to scope definitions. Seems more appropriate to use that approach for a query anyway.
Old Definition:
```ruby
ADD_ORGS            = Perm.where(name: 'add_organisations').first.freeze
```
New Definition:
```ruby
scope :add_orgs, -> {Perm.find_by(name: 'add_organisations')}
```
Switched all corresponding references to `Perm::CONSTANT` to `Perm.scope`

briri:
tests are in place. we can continue to add/enhance as we make changes in the system

Issue #204, opened 2017-03-27
Provide JS access to .po and .pot localization files


briri:
Investigate using the Gettext-I18n JS repo, https://github.com/webhippie/gettext_i18n_rails_js (recommended by gettext_i18n_rails), to translate the .pot and .po files into json files under assets.

Then update all of the old I18n references in the javascript files so that they use this new approach.

briri:
related to an issue about missing translations on #185

raycarrick-ed:
related to https://github.com/DMPRoadmap/roadmap/issues/182

raycarrick-ed:
when this ticket gets tackled it will need to take into account the i18N of the various statuses (plan, section, answer) which get generated by JS in places. I think there are all in views/answers/update.js.erb.  I have left TODO markers for them

briri:
I added the gem mentioned above and did the following:get checked into source code
- added the include statements to the `application.js` file
- added the current `Fastgettext.locale` to the main `<html>` tag
- replaced all of the I18n references in the js files
- added jquery.timeago text to the pot/po files

We will need to run `rake gettext:po_to_json` whenever we update the pot/po files. This task generates the json equivalents.

Won't be able to thoroughly test though until the recent export and plans fixes are merged in.

briri:
@stephaniesimms and @sjDCC this fix has been deployed to the server.  There were a few places on the plans page that I wasn't able to thoroughly test for this missing translation fix. You can see it working though on the guidance page if you click save without filling everything out. Please open a new ticket if you find further instances of 'missing translation' messages

stephaniesimms:
I can't find any instances of missing translations so closing for now.

PR #227, opened 2017-04-06
Mergining in Guidance fixes and functional controller tests

briri:
Fixes for #204 and #221. This code also includes the majority of the functional controller tests outlined in #212.

All tests should pass before accepting future PRs after this is merged! 

Note that it looked like there were some missing pieces to the plan pages. For example the `views/plans/_answer_form_ro.html.erb` was missing so I just added a stub in. If you have a more accurate copy @raycarrick-ed, please replace this one with yours. The registratations controller also had some typos in it that were preventing some of the various results to fail. 

Also be aware that I touched almost all of the controllers to update/correct how they were processing error conditions. Please pay close attention to those when you're reviewing.


app/policies/plan_policy.rb
```diff
@@ -28,17 +28,25 @@ def export?
     @plan.readable_by?(@user.id)
   end
 
-  def show_export?
-    @plan.readable_by?(@user.id)
-  end
+# TODO: There is no route to this function on the controller
+#  def show_export?
```
jollopre: Uncomment this, otherwise the export page will not be displayed. See the route here https://github.com/DigitalCurationCentre/roadmap/blob/development/config/routes.rb#L201

app/controllers/guidance_groups_controller.rb
```diff
@@ -66,7 +66,7 @@ def admin_update_publish
     if @guidance_group.update_attributes(params[:guidance_group])
       redirect_to admin_index_guidance_path(params[:guidance_group]), notice: _('Guidance group was successfully updated.')
     else
-      render action: "edit"
+      redirect_to admin_index_guidance_path(@guidance_group), notice: generate_error_notice(@guidance_group)
```
raycarrick-ed: We should adopt this as general pattern. Much better than what we've been working with. It might be worth generalising the 'Unable to save your changes' message e.g under plan creation we could do the same but might want 'Unable to Create Plan'. Add as a param maybe?

briri: Yes that's a good idea

app/policies/plan_policy.rb
briri: ah ok. missed that one

app/controllers/phases_controller.rb
```diff
@@ -176,7 +179,8 @@ def admin_update
     if @phase.update_attributes(params[:phase])
       redirect_to admin_show_phase_path(@phase), notice: _('Information was successfully updated.')
     else
-      render action: "admin_show"
+      # Redirecting here because the method loads a lot of stuff. We may lose whatever the user had entered though
```
briri: that's a good point. The redirect would likely lose whatever the user had entered. I'll switch them to renders

app/controllers/plans_controller.rb
```diff
@@ -17,121 +17,98 @@ def index
...
   # GET /plans/new
   def new
-    if user_signed_in? then
-      @plan = Plan.new
-      authorize @plan
-      @funders = Org.funders.all 
+    authorize Plan
+    
+    @plan = Plan.new
+    authorize @plan
```
briri: Ah. good catch. No need for the first one here

app/models/guidance.rb
```diff
@@ -13,12 +13,10 @@ class Guidance < ActiveRecord::Base
   ##
   # Associations
   belongs_to :guidance_group
-#  belongs_to :question
+  belongs_to :question
```
briri: yes. I thought I had it commented out but may have added it back in prior to removing those dropdowns from the guidance pages. I'll remove it.

briri:
Thanks for reviewing. I'll make the changes you have suggested.

briri:
Ok @raycarrick-ed and @jollopre. I made some changes based on your feedback:
- Uncommented show_export method in `plan_policy.rb` and `plan_controller.rb`
- broke the `generate_error_notice` method into 3 separate methods. One for each transaction type - create/update/destroy. Also updated it to require that you provide an object name since its not always equivalent to the class name (e.g. User vs Profile)
- Switched all save failure routes in controllers from redirect to render so that user does not lose the information they have already entered on the form
- Fixed tests to work with new error messaging and routing

PR #229, opened 2017-04-10
Emergency bug fixes for show_export.html.erb and jquery.timeago.js

briri:
- Some bugs that prevented show_export page from loading at all. 
- Bugs in the jquery.timeago.js localisation

I hot fixed the server. Please merge these in when you have a chance to sync everything up


briri:
I'm not completely happy with the jquery.timeago.js changes. Its a bit weird to use internationalization with the way they swap in the %d value. It sounds like we may go a different route though in the future (something other than jquery?) so we can leave it for now I guess since its working.

app/views/plans/show_export.html.erb
```diff
@@ -11,10 +11,10 @@
     <!--Text describing page content-->
     <%= raw _("<p>From here you can  download your plan in various formats. This may be useful if you need to submit your plan as part of a grant application.</br> Select what format you wish to use and click to 'Export'.</p>")%>
 
-    <% if @plan.phases.count == 1 %>
-        <%= render :partial => "/shared/export_links", locals: {plan: @plan, phase: @phan.phases[0]} %>
+    <% if @plan.template.phases.count == 1 %>
```
raycarrick-ed: This is a valid comment but since it isn't actually a bug I will accept the PR for now and then we can discuss how we want to handle things like this.

briri: Ah originally it was throwing an error on `@phan.phases[0]`. My first thought was that it needed to be `.template.phases`. So I updated them all accordingly. When that didn't fix the issue I realized that it was `phan` instead of `plan` so I fixed the typo and everything started working.

briri: As for the shortcuts, I think its just a matter of preference. I am not against them entirely. 

My concern is two-fold.
1. They could be confusing to external people looking at the code. `plan.phases` might make someone think that plans are associated directly with phases. rather than through the `template`. 
2. Creating shortcuts for the UI seems to me to be the reason the original codebase got so messy in the first place. All of the random associations between the tables and improperly placed functions in models and controllers that we had to clean up was largely due to improper implementation of shortcuts. 

Anyway, its not much more to type and I don't think it adds any DB querying overhead (i haven't tested that assumption though) to do `plan.phases` vs `plan.template.phases`

PR #256, opened 2017-04-19
Bug fixes and integration tests

briri:


briri:
agreed on avoiding `.where()` when possible. its too vague. I'm also a proponent of `.pluck()` which Sam started using.

PR #283, opened 2017-04-28
fixed a few tests after rebasing from DMPRoadmap

briri:
It looks like only some of the annotations work made it through. The site seems to be stable  (I was able to create simple templates and plans) but when running the tests I get errors about a missing annotations table. The PR was missing the db migration, so I tried copied it over the one from the DCC annotations branch but running the migration script failed because `delete_table` is invalid. 

I'm guessing that will be resolved when the annotations work is finished. @xsrust please update/add tests for the new table and any models you've changed or deleted before sending the final changes over. Thanks!


vyruss:
Merging with tests failing for now - @briri and I recommend we deploy after merging the @xsrust's changes into `development` and we have all tests passing.

vyruss:
Merging with tests failing for now - @briri and I recommend we deploy after merging the @xsrust's changes into `development` and we have all tests passing.

PR #296, opened 2017-05-08
Feature/sso

vyruss:


briri:
Looks good @vyruss. deployed locally and verified that the issues have been addressed

PR #298, opened 2017-05-08
Template bug fix

briri:
Correction to the templates controller and the way it pulls in funder and org templates and an update to the email notification that gets sent out when a user changes sharing permissions. 

This includes your shib/orcid PR from yesterday.

**This change requires that each templates.dmptemplate_id only have one published version! I had to manually correct some of the records on our test instance, and you may need to do the same if you have not run the current DMPonline_v4 to DMPRoadmap migrations!**


xsrust:
The current migrations dont actually enforce a single published version for each set of dmptemplate_id.  I've written a rake task to enforce this which will be in a future PR and added to the post-migration steps when it reaches the roadmap codebase.

vyruss:
Thanks @xsrust your rake task seems to work fine.

briri:
Excellent. Thanks @xsrust

PR #308, opened 2017-05-12
vyruss/bugfixes

vyruss:
(stuff in review column.)


config/environments/production.rb
```diff
@@ -11,7 +11,7 @@
   config.eager_load = true
 
   # Full error reports are disabled and caching is turned on.
-  config.consider_all_requests_local       = false
+  config.consider_all_requests_local       = true
   config.action_controller.perform_caching = true
```
briri: We shouldn't do this in the shared repo's file.

Issue #293, opened 2017-05-05
Language should default to English in Edit Profile


stephaniesimms:
At present the Language dropdown on the Edit Profile page defaults to Deutsch (the first in the alphabetical list) on first login. It should default to English (GB). Also, clicking on the "x" in the dropdown should delete the selection and/or the "x" should be removed.

![screen shot 2017-05-05 at 1 50 06 pm](https://cloud.githubusercontent.com/assets/12703638/25763936/5ba07318-319a-11e7-8972-5ad7f329b12d.png)
.

briri:
fixing as part of #230

briri:
Fix will be pulled in with new create plan workflow because it relies on jquery accessible dropdown

sjDCC:
Working on staging and DMPonline-test. There's no longer an 'x' by the language choice as it's a proper dropdown but the x against organisation clears the choice. Closing out

Issue #230, opened 2017-04-11
Edit profile fields look wonky


stephaniesimms:
Edit profile fields display in a weird, misaligned manner (screenshot below). Organisation drop-down is clearly shifted to the right. Other fields look oddly spaced too.

![screen shot 2017-04-10 at 8 05 53 pm](https://cloud.githubusercontent.com/assets/12703638/24891234/5b896e68-1e29-11e7-9243-c958a7ee8bf1.png)

briri:
This should be fixed when we replace select2 and replace the use of html tables to format these forms. 
I tried updating some of the css on the `select2-container` but nothing seemed to make it budge.

stephaniesimms:
ok @briri - I'll remove it from MVP-v0.3 and tag w/MVP.

briri:
related to #125

sjDCC:
Fixed on both staging and test. Closing out

Issue #150, opened 2017-03-03
Step 1 to create a plan


stephaniesimms:
As a researcher, I would like to select a template and begin writing a plan. This step involves selecting/confirming a user's org affiliation and/or which funding agency they are applying to.

See UX wireframe @JEK-III : http://h90c3m.axshare.com/#g=1&p=create_plan

* reorder selectors (select2 replacement): first selector is pre-populated w/user org if it exists; can also tick a box to skip it (see functional spec in blue notepad icon in wireframes)
* second selector to choose a funder
* if there are subdivisions for a specific funder (mainly an issue for NSF), a second drop-list appears in same selector to choose the subdivision 
* (optional) third selector for other circumstances where there are multiple templates for specific funder (multiple org templates, CRUK, etc)

Pivotal ticket: https://www.pivotaltracker.com/story/show/120866325
Related to the story for copying a plan:
https://www.pivotaltracker.com/story/show/120864879

Screenshots of existing DMPonline create plan steps: 1) select funder, 2) subsequent steps to select guidance from institution and other sources
![dmponline_plan_create1_funder](https://cloud.githubusercontent.com/assets/12703638/23566082/36dc1c6a-0005-11e7-8825-b862444c92cf.png)

![dmponline_plan_create2_guidance](https://cloud.githubusercontent.com/assets/12703638/23566111/5bf05eb2-0005-11e7-8a2a-670381a2d90c.png)

JEK-III:
Wireframe: http://h90c3m.axshare.com/#g=1&p=create_plan

briri:
![screen shot 2017-04-27 at 4 33 00 pm](https://cloud.githubusercontent.com/assets/1204467/25508496/503f92f8-2b67-11e7-84d0-2cebce16d340.png)

This isn't deployed yet, just wanted to run the screenshot by everyone first to see if the layout, wording, etc. all look good. @stephaniesimms @sjDCC @JEK-III

briri:
same as #261

stephaniesimms:
@briri this looks great to me. let's wait until Monday for @sjDCC to have a look too.

briri:
bringing over @sjDCC comments from the duplicate #261 
"I would also prefer that you could pick guidance to include at this point too, even though you can switch it off later. It would make the optional guidance more prominent because it's very hidden otherwise. Many unis recommend that researchers select the DCC checkbox, so by removing it from this stage, we're requiring all their guidance be rewritten.

[...]

Issue #125, opened 2017-03-01
Format all <form> with <fieldset> instead of <table>


briri:
Replace all of the existing <table> layouts on the forms with <fieldset>. Try to use the new dmproadmap.scss stylesheet.

briri:
related to #230

briri:
(copying over some conversation that happened in Slack)
@briri "I'm working on some of the UI issues that came out of our accessibility test last summer (specifically replacing the html tables that were causing problems).

I decided to try and follow the look of the new create plan page (#150). Let me know if this is a good approach for now.
![screen shot 2017-05-08 at 9 02 32 am](https://cloud.githubusercontent.com/assets/1204467/25860614/e4fc813e-3496-11e7-867e-62841718e5b2.png)

and here's the old look:
![screen shot 2017-05-08 at 9 06 33 am](https://cloud.githubusercontent.com/assets/1204467/25860631/faca3344-3496-11e7-92b0-1b37c89c1a7a.png)

I dropped the cancel button since it wasn't really doing anything besides bringing the user back to the view plans page. We can add it back if necessary. I also added some text below the email asking the user to also enter their password when changing the email address."

Some conversation then began about the workflow for changing one's email address and adding in the account recovery email address field. This particular task though is just about cleaning up the accessibility issues in the html form. We'll create a different ticket to record the desired changes to the edit profile page.

briri:
(copying over some conversation that happened in Slack)
@briri: "and here's an example of one of the updated forms that are part of a larger page (like template/plan editing).
![screen shot 2017-05-08 at 11 02 09 am](https://cloud.githubusercontent.com/assets/1204467/25860828/a43a590e-3497-11e7-9d53-6453090c8fef.png)

@JEK-III: "This looks great. I do recommend moving the field labels to above the form elements (as you did on the profile page) instead of to the left. Although there are plenty of exceptions, top labels are generally considered best practice. In this case, because the labels are of such variable length, “ID,” for example is very far from the associated input field.
If that’s problematic in the short-term, I would right-justify the labels– which would make them harder for a user to scan, but give us a tighter association between the label and the field.

(None of which impacts the original underlying accessibility issue, which should be taken care of by what you’ve done already)"

@briri: "I'll right justify them then. I tried letting them fall above their input boxes originally but the visual transition from the read-only view of this info to editable form was a bit disorienting"

briri:
(copying over some conversation that happened in Slack)
[...]

PR #309, opened 2017-05-12
Create plan page chnages, accessible autocomplete, accessible <form>s

briri:
- Updated plans_controller to work with new `Template.live()` method
- Recreated new plan page based on UX wireframes on #150
- Added accessible submit button to new plan and contact us pages
- Replaced select2 dropdowns on edit profile and new plan pages with jquery autocomplete #124
- Added fontawesome for the accessible autocomplete box's clear button
- Fixed alignment issue with Org selection box on edit profile page #230
- Language dropdown now defaults to the correct language #293
- Removed table and ol from form on the following pages per #125 
  - Contact Us 
  - Edit plan details
  - Plan answers
  - Edit/Add notes to a plan
  - Edit Profile

@xsrust I am not assigning any guidance to the new plans. Can you make sure that its adding them appropriately (if at all) when you push you setup your PR?

Note that there is a new `lib/assets/stylsheets/roadmap-form.scss` that contains styling for the new forms and the create plan page. I want us to move away from the old dual css approach that is currently in place to something simpler. This one is pretty specific to forms but its scope could be broadened easily. 

Also not that there are 2 new partials in `app/views/shared`. One is for adding the new jquery autocomplete box to a page (includes all the js wiring) and the other is an accessible submit button that properly informs a user using a screen reader that the button is disabled when the form is not complete. You can see examples of their use on the `app/views/plans/new_plan.html.erb`


briri:
Seems to be missing some commits. Will send another once resolved

PR #319, opened 2017-05-16
Template publish fix and sharing email fix

briri:
- updated the template controller so that a new version is created when a published template is updated
- updated sharing email notification to use `user.name` function so that it uses the email address when no first and surnames are defined


briri:
The conflict is on the removal of the template versioning code in the template_controller. You can accept and resolve if all looks good

PR #357, opened 2017-05-24
Vyruss/bugfixes

vyruss:
Fix for template editing (missing `question.guidance` killed it)


xsrust:
looks good.   i must have forgotten that file when i did the annotations work

Issue #352, opened 2017-05-23
View plan for an user with read_only permission does not work


jollopre:
If the plan to view includes questions such as radiobutton, checkbox, etc, the following error is raised: 
![snaphost](https://cloud.githubusercontent.com/assets/7393044/26354116/315f4514-3fbb-11e7-890a-fa77a0d9e6f4.png)

sjDCC:
@jollopre Could you share a NERC plan as read-only to s.jones@arts.gla.ac.uk please so I can test this?

jollopre:
done! Please try to see if you can edit anything related to plan's answer

sjDCC:
I can't edit anything in read-only mode (text areas or dropdowns etc) I can leave a note though.

All seems right and no errors are thrown so closing out

PR #360, opened 2017-05-25
Issue352

jollopre:
- Fixed read only problem for plan answers
- Fixed broken dropdown and multiselectbox question formats


vyruss:
Closes #352

PR #365, opened 2017-05-26
updated org-admin and plan create pages to reflect schema change

xsrust:
Added a new boolean 'migrated' to the templates model

This distinguishes between templates created to fill in the back end of a plan, and those templates which reflect the live data from the org-admin or plan creation interfaces

This is accessed in the templates model by a new scope valid.  To get only non-migrated templates, simply being a query with Template.valid and chain your query from there.


vyruss:
**Make sure to run `rake db:migrate` after pulling this.**

Issue #367, opened 2017-05-26
Admin rights not saving


sjDCC:
In trying to test #314 I have encountered another issue. Admin rights are not saving when assigned via super-admin. I tried to give an account rights and this showed as saving correctly but the permissions show as none and when I log into that account I don't have admin rights
![org-admin-right](https://cloud.githubusercontent.com/assets/6169940/26489027/919a71a8-41fd-11e7-9616-e1df9edd1eec.PNG)

sjDCC:
It seems the permission setting are being completely cleared out on saving as I've just tried testing with another account I had. This had superadmin rights and I tried to downgrade it to org-admin (i.e. I removed a few additional rights) Rather than saving as just the org-admin privileges, it's now saying I have no permissions
![admin2](https://cloud.githubusercontent.com/assets/6169940/26489132/0b93a5ce-41fe-11e7-9e51-4e7a593336c1.PNG)

sjDCC:
Working now. Closing out

Issue #345, opened 2017-05-22
Change shibboleth link text under edit profile


sjDCC:
Please change the shibboleth link text back to 'Your account is linked to your institutional credentials. Click here to unlink your institutional credentials' and 'Click here to link to your institutional credentials'. That'll be clearer to end users than a reference to the UK Access Federation

Grabbing a screen capture to point this out has raised another issue - where have the links gone?
![shibboleth-link](https://cloud.githubusercontent.com/assets/6169940/26303452/a18e78be-3edf-11e7-9e03-c41493e551c7.PNG)

vyruss:
How does _UK Access Federation (your institutional credentials)_ sound?

alexstuart:
Since this SP uses the UK federation Central Discovery Service (CDS), and the CDS lists all the IdPs in the UK federation and eduGAIN, it's possible that people can authenticate at an IdP outwith the UK federation and link their federated account with the local account. Something like _Your account has been linked to a federated account (institutional credentials)_ might be more appropriate.

Another point is that the label for this field should not be Shibboleth, as it's not just Shibboleth IdPs that can provide authentication statements to yout SP. Formally, you'd probably want to say _SAML_ but that might not have the name recognition. And whilst the omniauth provider you're using is a shibboleth provider, that's because it links to the Shibboleth **SP** software; it's not that it can only accept authentication from a Shibboleth **IdP**. For example, SURFconext appears to have IdPs which deploy ADFS or SimpleSAMLphp software.

vyruss:
I've been advised by @dsisu that _SAML_, _Shibboleth_ and _UK Access Federation_ all mean almost nothing to our users :)

alexstuart:
LOLOLOL. I take your point about SAML and Shibboleth, but still, I suggest that it's best not to describe the linking as "UK Access Management Federation" when the application can be accessed by researchers who are authenticating at IdPs that are registered outside the UK federation.

vyruss:
Done.

sjDCC:
There's a typo in this - 'your' is duplicated. It's also misleading to suggest they can create institutional credentials
![roadmap](https://cloud.githubusercontent.com/assets/6169940/26489844/ea9d31c0-4200-11e7-8ff0-cab230ea0804.PNG)

Probably best to just have an abbreviated version of what is currently in place on the live site e.g.

Link to your institutional credentials (UK users only)
OR
Unlink your institutional credentials

[...]

Issue #314, opened 2017-05-16
dmponline-test - user keeps admin rights when moving organisation


dsisu:
Hi all

If a user has administrative rights, they should lose them when they transfer to a new organisation as their role there may be different. At the moment this doesn't happen. A user pointed this out and I tested it with a spare account of mine too.

sjDCC:
This works. Closing out

Issue #259, opened 2017-04-20
Clarify permissions to be granted


sjDCC:
For usability, I think it would be good to add some help text to clarify what each permission allows. I've suggested some below but please correct these if wrong:

Grant permissions - this allows the user to assign permissions to other users within the same organisation. Users can only assign permissions they own themselves _(is that right?)_

Modify templates - this allows the user to create new institutional templates, edit existing ones and customise funder templates

Modify guidance - this allows the user to create and edit guidance

Change organisation details - this allows the user to amend the organisation details (name, URL etc) and add basic branding such as the logo 

API rights - this provides the user with an API token and grants rights to harvest information from the tool

These text could display within question mark icon popups or could perhaps be included in the table. Whatever works best display wise. 

I think the table headings should display as single words too, not text strings like "grant_permissions"

![capture](https://cloud.githubusercontent.com/assets/6169940/25233980/998fef8c-25d8-11e7-9550-5628b1663e3a.PNG)

stephaniesimms:
@sjDCC your clarification text might be sufficient to add as tooltips for now. I added a UX tag to this issue so we can confer about it this week. Permissions work is on their roadmap too so I just want to cross-reference.

vyruss:
Added tooltips and strings as described. Table looks ugly because of inherited CSS classes but let's leave that to UX team post-v0.3 :)

sjDCC:
This is fine for now, but I think some kind of question mark or other icon should appear so people know there is a tooltip if they hoverover. We should try to be consistent with how it's handled elsewhere in tool for usability. UX can hopefully make a suggestion
![capture](https://cloud.githubusercontent.com/assets/6169940/26487991/9ff9f0ce-41f9-11e7-93bd-f91b8d32b009.PNG)

sjDCC:
Fine for now. Closing out

PR #368, opened 2017-05-26
Various bugfixes/text changes

### LINKED ISSUE #362: Shouldn't be able to share plan with same user twice
You shouldn't be able to share the same plan with the same user multiple times. I did this by accident when testing with Jose. I was trying to let him see read-only rights, but had already shared the plan as editor so this overrode the second sharing. Two copies of the same plan showed up in his though

capture

vyruss:
Check that user role doesn't already exist before sharing. Fixes #362 
Permission yanking should not apply to super admin. #314 , #367 
Added dotted underline (HTML5 recommendation) for org admin user privileges - #259 
Text changes for #345


app/controllers/roles_controller.rb
```diff
@@ -10,22 +10,26 @@ def create
...
-        else
-          flash[:notice] = failed_create_error(@role, _('role'))
+          if user.nil?
+            registered = false
+            User.invite!(email: params[:user])
+            message = _('Invitation to %{email} issued successfully.') % {email: params[:user]}
+            user = User.find_by(email: params[:user])
+          end
+          @role.user = user
+          if @role.save
```
jollopre: Nice to see statements like this to handle db communications.

Issue #389, opened 2017-05-31
Can't edit existing plans


stephaniesimms:
Related issue for DMPOnline test #385 - I can create a new plan and answer questions but not edit anything for existing plans. Creating a separate issue for now because it's unclear to me whether DMPonline test and Roadmap are in sync.

On My plans page, clicking the plan title or Edit produces this error
![screen shot 2017-05-31 at 1 40 13 pm](https://cloud.githubusercontent.com/assets/12703638/26653117/fd1a62fa-4606-11e7-9c24-670aade01b59.png)

sjDCC:
Can't re-create this on DMPonline test but will leave to check on staging

stephaniesimms:
This is working on roadmap stage now. Closing issue.

Issue #385, opened 2017-05-31
Can't edit my existing plans


dsisu:
Hi, I can create plans but can't go back and edit them. Get 'page unavailable' error.

dsisu:
I thought I'd add a further comment, in case it helps. I can edit some plans but not others. There is no logic to it. The older plans seem to be less accessible than the newer ones. However, I did create a couple of plans today, which I was unable to edit once I saved them. Sam was here and saw that. I'm talking about dmponline-test.

sjDCC:
I have experienced this too. Screengrab with url example in case that helps
![capture](https://cloud.githubusercontent.com/assets/6169940/26646122/5ffef644-4632-11e7-881f-09fdc169a91a.PNG)

sjDCC:
Have tried various plans on various accounts including previous offending ones and all seem to open and edit fine now so closing this out

Issue #375, opened 2017-05-28
Institutional logos are missing from the top banner


sjDCC:
Logos files have not been transferred. Here's the banner for Salford on live and on test.
![salford-live](https://cloud.githubusercontent.com/assets/6169940/26532482/b691ce9a-43f9-11e7-9215-10ee8bfd9d8a.PNG)
![salford-test](https://cloud.githubusercontent.com/assets/6169940/26532485/bb4e5674-43f9-11e7-8d93-15e32eafe305.PNG)

The logo is not showing on the admin interface either under 'org details' though the file upload option is available

vyruss:
Weiwei's addition of logo uploads means we can't use the on-disk filenames anymore and have to re-upload for those orgs.

sjDCC:
I have the files to upload so this can effectively be closed.

Two things to check:
1. Should I upload them all now or will the data be changed / replaced before we push out to test?
2. Do we need to place an image resolution limit as one I uploaded looks very grainy. On the live site it displays much smaller
![sterling-test](https://user-images.githubusercontent.com/6169940/26875241-e74a6066-4b78-11e7-8e4e-341238e8b1d2.PNG)
![sterling-live](https://user-images.githubusercontent.com/6169940/26875247-ecc8122c-4b78-11e7-8c24-fa14c3f7f973.PNG)

vyruss:
This will be down to the header CSS scaling the image to 100 pixels height or something like that. I don't know how we can fix this besides disabling scaling

sjDCC:
I've searched and found a better image to replace this with. It was the only one that caused any grief so far so happy to leave as is, especially as people will be uploading themselves so can check result.

Should I upload all now or will the data change again before testing? I guess it will change due to the suggested answer issue so I should hold off for now?

Closing out as main issue fixed

vyruss:
@sjDCC we will re-migrate so don't upload yet. Cheers

PR #393, opened 2017-06-02
Vyruss/bugfixes

vyruss:
#375 #385 #389


vyruss:
@briri Dragonfly gem needs them to access the logo object (I don't understand why it needs so many columns tbh) =)

briri:
ugh ok. We will probably need to look at replacements then

Issue #337, opened 2017-05-21
Error when adding a question to an existing draft template


sjDCC:
I tried editing a draft template. I could add a section but when I tried to add a question to this I got a 404. Initially I only had text in the question field (the only mandatory one afaik). This wouldn't save so I tried adding text into some and then all other fields. Each time I got this
![capture](https://cloud.githubusercontent.com/assets/6169940/26287808/40287daa-3e7a-11e7-8c78-5fe809348c57.PNG)

Same happened when I tried to add a question to an existing saved section

sjDCC:
All working fine now. Closing out.

PR #404, opened 2017-06-06
Fixed problem when creating a question. Fixed JS to display/hide defa…

jollopre:
…ult answer. Fixed tinymce problem for example answer. DMPRoadmap/roadmap#337

### DIFF
--- a/app/models/question.rb
+++ b/app/models/question.rb
@@ -2,7 +2,7 @@ class Question < ActiveRecord::Base
   ##
   # Associations
   has_many :answers, :dependent => :destroy
-  has_many :question_options, :dependent => :destroy
+  has_many :question_options, :dependent => :destroy, :inverse_of => :question  # inverse_of needed for nester forms
   has_many :annotations, :dependent => :destroy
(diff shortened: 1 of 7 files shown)


app/models/question.rb
```diff
@@ -2,7 +2,7 @@ class Question < ActiveRecord::Base
   ##
   # Associations
   has_many :answers, :dependent => :destroy
-  has_many :question_options, :dependent => :destroy
+  has_many :question_options, :dependent => :destroy, :inverse_of => :question  # inverse_of needed for nester forms
```
vyruss: interesting

lib/assets/javascripts/admin.js
```diff
@@ -30,147 +30,28 @@ $( document ).ready(function() {
...
-				data:  {
-	 				section_id : $('#sections_select').val()
-	 			}
-		 	});
-		 	//$('#phases_select').show();
-		 	//$('#versions_select').show();
-		 	//$('#sections_select').show();
-		 	$('#questions_select').show();
-		 });
-*/
```
vyruss: Thanks for getting rid of this terrible Ajax

PR #413, opened 2017-06-12
fixed logic which lead to guidance not being displayed

xsrust:
fixes a guidance display bug.
also removes a deprecated rake task, unrelated to bug(was just already git add'ed)


sjDCC:
Checked on Tuuli and this works. Closing out

PR #423, opened 2017-06-15
Xsrust/bugfixes

xsrust:
- Funder guidance displays on customised templates
    - before was just displaying the customiser's guidance twice
 - Since annotations display with the org's abbreviation, added logic to drive org admins of orgs without an abbrivation to the show org details page
 - removed fuzzy translations warnings which appear when you run a rails c/raketask


xsrust:
Tests now all pass

PR #443, opened 2017-06-20
Issue 431

jollopre:
The JS located at /lib/assets/javascript/plans.js has been refactored. The issue itself it is far to be closed but this changes are good enough to be merged.


jollopre:
yes, we will need to recompile JS again. And the file config/application.rb also has been modified.

PR #446, opened 2017-06-22
Xsrust/bugfixes

xsrust:
Docx export changes


xsrust:
Forgot pushing would add to PR.

Tiny MCE text boxes now automatically re-size.  can change the bottom padding if desired.  I reduced it from the default.
fixes #448

Issue #431, opened 2017-06-16
JavaScript files cleanup


jollopre:
The JS files have to be reviewed and check for any potential refactoring. 

For instance, lib/assets/javascripts/plans.js has code that is no longer needed.

This ticket is related to #170

briri:
@jollopre can we close this one out?

PR #453, opened 2017-06-23
Using data-toggle attribute to initiliase and trigger any popover.

jollopre:
DMPRoadmap/roadmap#431

### DIFF
--- a/lib/assets/javascripts/bootstrap_listeners.js
+++ b/lib/assets/javascripts/bootstrap_listeners.js
@@ -15,9 +15,10 @@ $(document).ready(function(){
         $(this).parent().find('.minus-laranja').removeClass('minus-laranja').addClass('plus-laranja');
     });
     // Initialises all tooltips present on a page
-    $('.has-tooltip').tooltip({
-        placement: "right",
-        trigger: "focus"
-    });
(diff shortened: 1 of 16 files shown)


lib/assets/javascripts/bootstrap_listeners.js
```diff
@@ -15,9 +15,10 @@ $(document).ready(function(){
         $(this).parent().find('.minus-laranja').removeClass('minus-laranja').addClass('plus-laranja');
     });
     // Initialises all tooltips present on a page
-    $('.has-tooltip').tooltip({
-        placement: "right",
-        trigger: "focus"
-    });
+    $('.has-tooltip').tooltip({ placement: "right", trigger: "focus" });
     $(".help").popover();
+    //Initiliases all popovers on a page
```
briri: looks good @jollopre we'll have to remember to choose this one and remove mine when we merge over the CDL-MVP branch. I solved it in a similar way but used `$("a[rel='popover'])`. Yours is better since it doesn't force the use of an anchor

jollopre: I noticed today when refactoring admin.js that there were multiple classes to trigger the same behaviour. That's the example provided in bootstrap website. Your example is good too but using data-* does not tight to a specific html tag.

Issue #463, opened 2017-06-25
Institutional custom section on a funder template: enable this to display in 1st position


sjDCC:
Manchester uni raised an issue about their custom section display. They’d like to place this in the first position on all templates but it doesn’t seem possible to do that consistently. Sometimes the admin interface displays it there but the front end doesn’t or vice versa. There is also one case where their custom section isn’t displaying at all despite being published.

I’ve gone through all their customisation data and summarised what’s happening below:
**Admin interface**
- 8 templates are showing as having the University of Manchester section as 1st position. These are: BBSRC, EPSRC, ESRC, H2020, NSF, Population Research Committee, STFC, Wellcome Trust
- 4 templates are showing as having the University of Manchester section as 2nd position. These are: AHRC, MRC, NERC & CRUK

**Front end – user interface**
The UoM section displays on all bar the Population Research Committee template. This is a secondary template for Cancer Research UK. Other than that, there’s nothing different about it and the customisation is showing as published so it’s unclear why it’s not displaying.
The position of the UoM section varies and is not consistent with what shows on the admin view. Here are the discrepancies:
- H2020 template suggests 1st position on admin side but displays second to end users
- STFC template suggests 1st position on admin side but displays second to end users
- Wellcome Trust suggests 1st position on admin side but displays as final section to end users
- AHRC, NERC & CRUK all suggest the UoM section is in second position on admin side but it displays first where they want it…

So it does seem possible to get the Uni section in first position, but I don’t know how we can make that consistent. Perhaps this is a data issue?

xsrust:
@sjDCC would allowing customizers to have a section with order of display = 0 be a sufficient solution?  This would allow them to always set their section to come first.  Template owners would have their section numbers start at 1 so there should be no confusion.

sjDCC:
Yes, I think that would work. Currently the dropdown for ordering sections starts at 1.

Only thing to watch is that if we re-start the ordering from 0, we don't want any newly added funder templates to have their sections ordered from 0 too. This should be reserved for customisations. Hope that makes sense!

xsrust:
Yes, that makes perfect sense. Only customizers will be able to reduce a section number to 0, they will start at 1 on own templates.

sjDCC:
Perfect, looks great. Will let Manchester know this is fixed

PR #465, opened 2017-06-26
Xsrust/bugfixes

xsrust:
Adds options for bullets to be square, disk, or circle in TinyMCE editor.  Only Square and circle export to docx due to limitations in the htmltoword gem we are using for this functionality.

Updated Questions controller to properly deal with annotations being updated and assign org to guidance.

### DIFF
--- a/app/controllers/questions_controller.rb
+++ b/app/controllers/questions_controller.rb
@@ -56,9 +56,20 @@ class QuestionsController < ApplicationController
       if guidance.blank?
         guidance = @question.annotations.build
         guidance.type = :guidance
+        guidance.org_id = current_user.org_id
       end
       guidance.text = params["question-guidance-#{params[:id]}"]
       guidance.save
+    end
+    example_answer = @question.get_example_answer(current_user.org_id)
+    if params["question"]["annotations_attributes"]["0"]["id"].present?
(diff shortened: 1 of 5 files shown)


xsrust:
Now addresses #463, allowing customizer's sections to optionally appear before the template creator's sections.

briri:
How does the customizer's section appear before the funder's sections? By selecting zero? Does selecting 3 make it appear at the end even if the funder has 5 sections? We should probably add some tooltips or text somewhere on the page that explains how it works for the end user.

xsrust:
There were complaints from org-admins that when they set their section to 1, it would not always appear first.  By allowing the customizer to select 0, they will see their section first.

As changing this number re-orders the sections in the org-admin interface(and saving refreshes page), I'm not sure how much explanation will be needed, but tooltips could be helpful.

briri:
I guess we can add instructions later on then if we start getting a lot of helpdesk questions. Its a small percentage of the user base so hopefully not a big deal

xsrust:
That sounds reasonable to me.

vyruss:
@briri @xsrust yes, customizers are a small number of people right now, and those who need reordering are even fewer. If we find that more people ask about this issue we can document it within that page

PR #480, opened 2017-06-30
Merge FROM CDL-MVP

briri:
Config/DB changes
-----------------------------------------
- There is a new config setting that will enable the custom shib discovery service page. It is off by default
- There is a new config setting for the default plan visibility. Set to private visibility by default 
- The plans table has a new data_contact_email and data_contact_phone. 
- There is a new org_identifiers table. Its setup and mapped similarly to our user_identifiers table
- The users table now has a new 'prefs' field that stores a hash. This stores the user's email notification preferences 

Overview of code changes
-----------------------------------------
- removed bootstrap
- removed show route for orgs, guidance groups and guidance
- added Shib DS routes to the orgs_controller
- added visibility logic to the plans_controller
- added public DMPs pages and route to plans_controller
- added email notifications to registrations_controller
- updated layout of edit profile page
- added error messaging for when a user tries to attach an orcid that's already used
- removed the old helper logic that built the view plans columns
- added deep_copy to plans and answers
- added an org_identifiers table for use with shibboleth DS and future maDMP stuff (fundref)
- added an owned_by? method to plans
- added data_contact_email and data_contact_phone to plans table
- cleaned out old block-commented lines from plans
- rebuild CSS from scratch (see below)
- restructure the JS (see below)
- Applied new CSS to all but the template writing pages (in progress)

Restructured JS (lib/assets/javascripts)
-----------------------------------------
javascripts
  |-----application.js 
  |-----dmproadmap/    (Generic JS - available to each page)
	|-----locale/        (Fastgettext json)
	|-----vendor/        (jquery, autocomplete, etc.)
	|-----views/         (page specific JS - loaded by specific pages)
	
- dmproadmap/ contains generic JS (e.g. tabs, accordions, helper functions, etc.) These files are loaded as part of the core application.js
- vendor/ contains JS components from outside entities
- views/ contains page specific JS. The file naming conventions follow the app/assets/views hierarchy (e.g. views/plans/edit.js belongs to views/plans/edit.html.erb) These are loaded by their specific pages
- admin.js and plans.js (leftover and will go away once we're confident all of the JS has been accounted for in the new files)

Restructured CSS (lib/assets/stylesheets)
-----------------------------------------
css
  |-----dmproadmap.scss  (This is the new file that can/should be customized per install)
  |-----dmproadmap/      (Generic CSS styles broken apart by topic)
  |-----vendor/          (jquery ui styling)
	
- The new dmproadmap.scss file is meant to be customized for each installation. It allows you to change the colors, fonts, etc. in one spot. You can also control various aspects of the header from here
- The files in dmproadmap/ are broken apart into functional parts like accordions, tabs, tables, forms, etc. The base.scss contains the general page layout and is loaded first.
- If specific tweaks need to be made for a paticular page they should be done within the appropriate scss file at the bottom. For example if we need to restyle tabs for the edit profile page we would place that customization at the bottom of the tabs.scss file.


briri:
Do NOT merge until we've synced up the master branch with what is currently in development!

app/views/contact_us/contacts/new.html.erb
```diff
@@ -1,111 +1,95 @@
-<% javascript "contacts/new_contact.js" %>
+<% javascript "views/contacts/new_contact.js" %>
 
 <h1>
   <%= _("Contact Us") %>
 </h1>
 <p>
-  <%= raw _('%{application_name} is provided by the %{organisation_name}. You can find out more about us on our <a href="%{organisation_url}" target="_blank">website</a>. If you would like to contact us about %{application_name}, please fill out the form below.') % {organisation_name: Rails.configuration.branding[:organisation][:name],
+  <%= raw _('%{application_name} is provided by the %{organisation_name}.<br /> You can find out more about us on our <a href="%{organisation_url}" target="_blank">website</a>. If you would like to contact us about %{application_name}, please fill out the form below.') % {organisation_name: Rails.configuration.branding[:organisation][:name],
```
vyruss: Yes, I think that's the best way right now. Else we will get small fragments of text to translate like "is provided by the" and it will be difficult to understand the context. We can change the `raw` to `sanitize` everywhere with no ill effects though I believe.

app/views/guidances/admin_new.html.erb
```diff
@@ -1,99 +1,64 @@
...
+        <%= f.collection_select(:theme_ids, @themes,
+                  :id, :title, {prompt: false, include_blank: 'None'}, {multiple: true}) %>
+        <div class="inline">
+          <a href="#" data-toggle="popover" rel: "popover" data-html: "true"
+             data-content="<%= _('Select which theme(s) this guidance relates to.') %>">
+          <span class="fa fa-question-circle"></span></a>
+        </div>
+      </div>
+      <div class="form-input">
+        <%= f.label :published %>
```
vyruss: This is related to issue #483 - I'm also unsure how to translate these labels in the frontend if we start using the helpers for this

app/controllers/plans_controller.rb
```diff
@@ -105,6 +114,9 @@ def create
   def show
     @plan = Plan.eager_load(params[:id])
     authorize @plan
+
+    @visibility = @plan.visibility.blank? ? @plan.visibility.to_s : Rails.application.config.default_plan_visibility
```
briri: yes, good catch. otherwise its always returning the default

app/controllers/plans_controller.rb
```diff
@@ -157,8 +196,11 @@ def update
     @plan = Plan.find(params[:id])
     authorize @plan
 
+    attrs = plan_params
+    attrs['visibility'] = Rails.application.config.default_plan_visibility if plan_params['visibility'].blank?
```
briri: Yes, a rake task would work here. We'll have to be sure to record it as a necessary step in the release notes though!

app/controllers/users_controller.rb
```diff
@@ -55,4 +57,28 @@ def admin_update_permissions
     end
   end
 
+  def update_preferences
```
briri: yes indeed. Have create a new ticket #481 to deal with this afterward

app/models/answer.rb
```diff
@@ -13,30 +13,41 @@ class Answer < ActiveRecord::Base
...
 #  # Make sure the plan and question are associated with the same template!
 #  validates :plan, :question, answer_for_correct_template: true
 
+  ##
+  # deep copy the given answer
+  #
+  # @params [Answer] question_option to be deep copied
+  # @return [Answer] the saved, copied answer
+  def self.deep_copy(answer)
+    answer_copy = answer.dup
```
briri: yes. will update to include options

app/models/user.rb
```diff
@@ -26,31 +31,35 @@ def filter(query)
...
   #                :accept_terms, :role_ids, :dmponline3, :api_token,
-  #                :organisation, :language, :language_id, :org, :perms, 
+  #                :organisation, :language, :language_id, :org, :perms,
   #                :confirmed_at, :org_id
 
   validates :email, email: true, allow_nil: true, uniqueness: {message: _("must be unique")}
 
+
+  validates :prefs, presence: true
+  before_validation :create_default_preferences
```
briri: this can probably go away once we start saving only the deviations from default.
We've found that validate? always fires before save (which gets fired after create and update). Since we're validating the presence_of, before_create never fires since the validation step fails.
Going to leave for now. We'll clean up in the new ticket #481

app/views/contact_us/contacts/new.html.erb
briri: I've had similar concerns with the embedded html tags. In part due to it being a potential vulnerability, but if any of the html tags get messed up by the translator the page layout could suffer.

app/views/guidances/admin_index.html.erb
```diff
@@ -1,140 +1,141 @@
...
-			        </td>
-			    <% end %>
-			    <% if guidance.guidance_group.present? then %>
-            <td class="dmp_td_small">
-		          <%= guidance.guidance_group.name %>
+  <!-- List of guidance -->
+  <% if @guidances.length > 0 then%>
+    <table class="dmp_table tablesorter">
+      <thead>
+        <% if @guidances.count > 10 %>
```
briri: done

app/views/guidances/admin_new.html.erb
briri: ah good point. Will add a ticket to go through and swap out labels to the standard html `<label for="guidance_published"><%= _('Published') %></label>`.
See ticket #483

app/views/orgs/shibboleth_ds.html.erb
```diff
@@ -0,0 +1,84 @@
...
+
+<div class="content">
+  <%= form_for 'shibboleth_ds', url: shibboleth_ds_path, html: {class: "shibboleth-ds-form roadmap-form"} do |f| %>
+    <fieldset class="standard padded">
+      <div class="form-input">
+        <label for="org_name">
+          <%= _('Look up your institution here') %>
+        </label>
+    
+        <% if @orgs.count <= 10 %>
```
briri: done

app/views/orgs/shibboleth_ds.html.erb
```diff
@@ -0,0 +1,84 @@
...
+        
+        <%= render partial: 'shared/accessible_submit_button',
+                   locals: {id: 'submit-button', 
+                            val: _('Go'), 
+                            disabled_initially: true,
+                            classes: 'small-input-button',
+                            tooltip: _('Select an institution.'),
+                            classes: 'inline left-indent'} %>
+
+        <% if @orgs.count > 10 %>
```
briri: done

app/views/phases/_answer_form.html.erb
```diff
@@ -29,95 +29,94 @@
...
+      <% if annotations.present? || question_guidances.present? %>
+        <% active_tab = 'guidance_tab' %>
+        <li class="guidance_tab active" role="tab" aria-controls="guidance-question-area-<%= question.id %>">
+          <a href="#guidance-question-area-<%= question.id %>" aria-selected="true"><%= _('Guidance') %></a>
+        </li>
+      <% else %>
+      <% active_tab = 'note_tab' %>
+      <% end %>
+      <li class="note_tab <%= "active" if active_tab == 'note_tab' %>" role="tab" aria-controls="comment-question-area-<%= question.id %>">
+        <% if comments.count > 0 %>
```
briri: also done

app/views/contact_us/contacts/new.html.erb
vyruss: @briri pootle turns red and warns the translator when they miss out brackets, tags, quotes in the translated string as compared to original - so that's a first line of defense. Second line is we can see the diff before we merge the new translation in ;)

app/views/guidances/admin_new.html.erb
jollopre: There is no need to translate to standard html, as I said the other day, FormHelper can help as to detect any syntax error whereas using standard html could fail silently.

<%= f.label(:published, _('Published: ') %> should work. Reference at: http://api.rubyonrails.org/v5.1/classes/ActionView/Helpers/FormHelper.html#method-i-label

briri: Yes, also a good point @jollopre ;) I updated #483 with your suggestion

PR #488, opened 2017-07-03
Fixes based on PR feedback

briri:


xsrust:
It looks like this addresses my comments from the other PR

PR #489, opened 2017-07-05
Xsrust/bugfixes

xsrust:
Bugfix:
- fixes generation of api token when a user is given api permissions through the org-admin interface
Performance:
- Remove selection of roles/plans from default scope for users
- Remove 2 N+1 querry issues on plan edit pages (each generating ~100-200 queries)
- Re-evaluated and updated indexes to reflect queries in the codebase


briri:
indexes for the win! Thanks for cleaning these up @xsrust

Issue #496, opened 2017-07-05
Add aria tags to modal dialogs


briri:
Make sure all modal dialogs are using the proper Aria tags.
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_dialog_role

Search for `modal` across the views to find the appropriate files. Also make sure that they follow the same pattern as the one found in the `views/layout/_signin_signout.html.erb`. 
This uses `data-toggle="modal"` attribute to signify that the link will open a modal and the `href` tag to indicate which div to display as the modal. That div should have the `class="modal"` attribute and a `style="display: none;"`

For example:
```html
<a href="#header-login-form" data-toggle="modal" dialog-width="450px"><%= _('Sign in')%></a> /
<a href="#header-register-form" data-toggle="modal" dialog-width="450px"><%= _('Create account')%></a>

<div id="header-login-form" class="modal" style="display:none">
  <%= render :partial => "shared/login_form" %>
</div>
<div id="header-register-form" class="modal" style="display:none">
  <%= render :partial => "shared/register_form", locals: {extended: false} %>
</div>
```

xsrust:
above commit should address all modals within the site.

stephaniesimms:
Unless @JEK-III is able to test this? i'll close the issue and await next round of accessibility testing.

PR #499, opened 2017-07-06
Xsrust/mvp

xsrust:
First commit adds aria tags according to #496.  Please check that the formatting of these is correct as I am not very familiar with aria.

The second commit is a bit of housekeeping which removes the deprecated 'views/projects' for which we have no controller.


app/views/guidances/admin_edit.html.erb
```diff
@@ -52,6 +52,6 @@
 <%= tinymce :content_css => asset_path('application.css') %>
 
 <!-- alert for the default template-->
-<div id="edit_guidance_alert_dialog" class="modal" style="display:none">
+<div id="edit_guidance_alert_dialog" class="modal" role="dialog" aria-label="<%=_("Missing Fields Alert")%>" style="display:none">
   <ul id="missing_fields_edit_guidance"></ul>
```
briri: I think this is ok for now. This ties into the larger issue of the site not having a consistent method for relaying form input errors.
Please make sure the focus gets set on the close button when the dialog opens.

app/views/layouts/_signin_signout.html.erb
```diff
@@ -29,16 +29,16 @@
...
-    <a href="#header-login-form" data-toggle="modal" dialog-width="450px"><%= _('Sign in')%></a> /
-    <a href="#header-register-form" data-toggle="modal" dialog-width="450px"><%= _('Create account')%></a>
+    <a id="sign_in_modal" href="#header-login-form" data-toggle="modal" dialog-width="450px"><%= _('Sign in')%></a> /
+    <a id="create_account_modal" href="#header-register-form" data-toggle="modal" dialog-width="450px"><%= _('Create account')%></a>
   <% end %>
 <% end %>
 
 <% if !user_signed_in? && !isActivePage(root_path) then %>
-  <div id="header-login-form" class="modal" style="display:none">
+  <div id="header-login-form" class="modal" role="dialog" aria-labelledby="#sign_in_modal" style="display:none">
```
briri: These two should be `aria-labelledby="some_id`. I had added an `<h2>` to both of them that is only visible to screen readers: `<h2 aria-hidden="false" class="aria-only"><%= _('Sign in') %></h2>`
So just add an id to those I think and use the aria-labelledby attribute. Then make sure the focus is set on the first element in the forms (I think it should be by default but best to verify)

briri: The doc says that the `aria-labelledby` attributes here are supposed to point to the `<h2>` on the partials.

Issue #483, opened 2017-06-30
Use Rails ActionView helpers in views


briri:
Switch `<label>` and `<input>` tags over to Rails view helpers. For example switch `<label for="email"><%= _('Email') %></label>` to `<%= f.label :email %>`

briri:
@xsrust noted on the latest PR that using the Rails helpers for labels prevents those labels from being translated. We should ensure that all labels are using the standard html tags `<label for="guidance_published"><%= _('Published') %></label>`. Input fields can use the helpers still.

briri:
FYI. the `forms.scss` file will now automatically apply a colon to the end of any `<label>` that does not have the `no-colon` or `checkbox-label` class (e.g. `<label><%= _('Published') %></label>` results in: `Published:` and `<label class="no-colon"><%= _('Published') %></label>` results in: `Published`). 

It will also add an asterisk at the end of your labels if you give it the `required` class (e.g. `<label class="no-colon"><%= _('Published') %></label>` results in: `Published * :`).

stephaniesimms:
Closing this issue pending next round of accessibility testing.

Issue #501, opened 2017-07-07
reCAPTCHA needs integration with the contact_us form


jollopre:
We use reCAPTCHA and contact_us gems for the contact_us form however integrating them seems rather cumbersome and currently does NOT work.

reCAPTCHA gem integrates with Rails controllers by invoking verify_recaptcha (see https://github.com/ambethia/recaptcha#rails-installation) however we can't integrate that with contact_us (another gem) unless we modify that gem.

An investigation about how google's reCAPTCHA works (https://developers.google.com/recaptcha/intro) should help us to determine whether or not we should remove the Ruby gem and use directly that library instead.

briri:
I agree with you we should remove the contact_us gem entirely. It would be a pretty simple form to implement and we would then not have to worry about the interaction between the two.

vyruss:
@jollopre what do you mean by "does not work" on master or development? It appears to work on the live services.

jollopre:
@vyruss I mean that although reCAPTCHA works in isolation, it is not needed to confirm that you are not a robot in order to send a form with data to the server.
That's mainly because the verification of the reCAPTCHA is not done in the controller...

briri:
removing this from the board since its included now in v1.0.4

PR #502, opened 2017-07-07
DMPRoadmap/roadmap#483

jollopre:
-Replaced wrong parameter names since it is expected contact_us_contact[...]
-Fixed logic for when an email is valid and re-introduced the previous regex, which is compatible with contact_us gem
-Replaced disabled fields (e.g. email, name) with readonly attribute, otherwise these parameters are not send to the server

### DIFF
--- a/app/views/contact_us/contacts/new.html.erb
+++ b/app/views/contact_us/contacts/new.html.erb
@@ -16,36 +16,31 @@
       <fieldset class="side-by-side">
         <% if ContactUs.require_name %>
           <div class="form-input">
-            <label for="contact_us_contact_name"><%= _('Name') %>*</label>
-            <input type="text" id="contact_us_contact_name" name="contact_us[contact_name]" 
-                   class="left-indent required input-large" value="<%= current_user.nil? ? '' : current_user.name(false) %>"<%= current_user.nil? ? '' : 'disabled="true"' %> />
+            <%= f.label(:name, _('Name')+'*') %>
+            <%= f.text_field(:name, class: "left-indent required input-large", value: current_user.nil? ? '' : current_user.name(false), readonly: !current_user.nil?) %>
           </div>
         <% end %>
         
         <div class="form-input">
-          <label for="contact_us_contact_email"><%= _('Email') %>*</label>
-          <input type="email" id="contact_us_contact_email" name="contact_us[contact_email]" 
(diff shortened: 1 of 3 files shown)


jollopre:
The issue is not finished yet but it's worth merge back otherwise the contact_us does not work

app/views/contact_us/contacts/new.html.erb
```diff
@@ -16,36 +16,31 @@
       <fieldset class="side-by-side">
         <% if ContactUs.require_name %>
           <div class="form-input">
-            <label for="contact_us_contact_name"><%= _('Name') %>*</label>
-            <input type="text" id="contact_us_contact_name" name="contact_us[contact_name]" 
-                   class="left-indent required input-large" value="<%= current_user.nil? ? '' : current_user.name(false) %>"<%= current_user.nil? ? '' : 'disabled="true"' %> />
+            <%= f.label(:name, _('Name')+'*') %>
```
briri: You can get the '*' character by adding `class: 'required'` for these labels. Its fine to wait though for the final push with the recaptcha fixes.

lib/assets/javascripts/dmproadmap/forms.js
```diff
@@ -49,7 +49,7 @@ function validatePassword(sPassword) {
 
 // ---------------------------------------------------------------------------
 function validateEmail(sEmail) {
-  var filter = /^[a-zA-Z0-9]+[a-zA-Z0-9_.-]+[a-zA-Z0-9_-]+@[a-zA-Z0-9]+[a-zA-Z0-9.-]+.[a-z]{2,4}$/;
+  var filter = /[^@\s]+@(?:[-a-z0-9]+\.)+[a-z]{2,}$/;
```
briri: should we update the regex in the validators as well?

briri:
Looks good.
@jollopre is recaptcha not working in the development and master branches or is this just an issue with this branch?

jollopre:
does not work on master either! See issue #501

lib/assets/javascripts/dmproadmap/forms.js
jollopre: This JS regex maps the same regex specified at contact_us model https://github.com/JDutil/contact_us/blob/master/app/models/contact_us/contact.rb#L7 

I'm not aware of having in are User models a validator for email...

briri: Just checked. Looks like `app/validators/email_validator.rb` is using the same regex that contact_us uses:
`/\A([^@\s]+)@((?:[-a-z0-9]+\.)+[a-z]{2,})\z/i`

Issue #161, opened 2017-03-03
Revise global navigation


stephaniesimms:
Non-logged-in version: https://wireframe.cc/BatUuz
Logged-in version: https://wireframe.cc/FgAwAU

Menu items Google doc: https://docs.google.com/document/d/1iMOP6wU6Qv7RObjM8El9EeZzDyeyke62qewS8qZx4-0/edit

Pivotal ticket: https://www.pivotaltracker.com/story/show/131906263
related ticket https://www.pivotaltracker.com/story/show/120873081

briri:
implemented changes found in the wireframes and deployed to dev and stage

stephaniesimms:
per UX call 12 Jul - transfer screenshot and logo variance comments from #424

vyruss:
It's not clear to me how to reinstate current-page highlighting in the navbar (for DMPonline CSS)

briri:
@vyruss The `views/layouts/_navigation.html.erb` still uses the `isActivePage()` method to determine what page the user is on. This method sets the `active` class on the menu item.

Right now the CSS does the following:
```css
/* Standard menu item */
color: $header-text-color;
background-color: $primary-color;

/* Active menu item */
color: $primary-color;
background-color: $white;
```
The two variables, `$primary-color` and `$white` are set in your dmproadmap.scss file. 

If you find that you need to be more granular than that feel free to create 2 new variables in dmproadmap.scss and reference them in the dmproadmap/base.scss file.

sjDCC:
I really like the icons in this wireframe:

![icons](https://user-images.githubusercontent.com/6169940/28792337-24c96b70-7627-11e7-9bb2-16200f4d3762.PNG)

If we have space to include those that would be great. It brings it back to one click for actions not two, which is preferable

stephaniesimms:
we all like the icons @sjDCC . i think the main reason for not using them is real estate - this table is extremely crowded as is and the dropdown Actions menu seemed like the best solution. @JEK-III also provided some rationale for the use of icons - if we introduce them here we should do so throughout the tool for consistency and that would require more time and effort (correct me if i'm wrong here). open for discussion though.

sjDCC:
Yeah, I figured space may preclude it. Nevermind.

sjDCC:
Does this ticket have any work left to do in it? Doesn't seem so to me from reading though. Close out?

stephaniesimms:
this predates Bootstrap work and seems moot now. i'm going to close.

PR #503, opened 2017-07-07
New navigation menus, accessibility for images and fontawesome icons

briri:
- Implemented new navigation menus as found in current wireframes #161 
- Updated `label.required` styling to add an '*' to the front of the label
- Updated contact us page to use css to apply '*' to labels
- Fixed layout issues on org details page
- Updated header to use a white background and to replace the roadmap logo with the institution's logo.
- Updated dragonfly logo resize to set logo height at 165
- Added `role="button"` to all popover question mark buttons
- Added aria tags to fontawesome icons
- Added alt text to all images
- fixed a few tests affected by the above changes
- removed old commented out lines from Org model


briri:
- Standardized create/update/delete success messages
- Updated tests accordingly
- Updated positioning of messages

app/controllers/annotations_controller.rb
```diff
@@ -18,7 +18,8 @@ def admin_create
     guid_save = guidance.present? ? guidance.save : true
 
     if ex_save && guid_save
-      redirect_to admin_show_phase_path(id: @question.section.phase_id, section_id: @question.section_id, question_id: @question.id, edit: 'true'), notice: _('Information was successfully created.')
+      typ = (!ex_save && !guid_save ? 'example answer and guidance' : (!guid_save ? 'guidance' : 'example answer'))
```
xsrust: because of the above if statement will requires both ex_save and guid_save to be true, this code will always return 'example answer'.  This should be revised with `example_answer.present?` and `guidance.present?`

app/controllers/annotations_controller.rb
```diff
@@ -74,15 +75,16 @@ def admin_update
     @section = @question.section
     @phase = @section.phase
     if ex_save && guid_save
-      redirect_to admin_show_phase_path(id: @phase.id, section_id: @section.id, question_id: @question.id, edit: 'true'), notice: _('Information was successfully updated.')
+      typ = (!ex_save && !guid_save ? 'example answer and guidance' : (!guid_save ? 'guidance' : 'example answer'))
```
xsrust: because of the above if statement will requires both ex_save and guid_save to be true, this code will always return 'example answer'.  This should be revised with `example_answer.present?` and `guidance.present?`

app/controllers/application_controller.rb
```diff
@@ -80,6 +80,10 @@ def failed_update_error(obj, obj_name)
   def failed_destroy_error(obj, obj_name)
     "#{_('Could not delete the %{o}.') % {o: obj_name}} #{errors_to_s(obj)}"
   end
+  
+  def success_message(obj_name, action)
```
xsrust: Since this is just a helper function to create text, perhaps this could live in a helper?  I noticed that we have a few other error/message creators in the application controller as well, but I think this type of function should be re-factored to be part of a helper.

app/models/org.rb
```diff
@@ -124,98 +124,6 @@ def short_name
   end
 
   ##
-  # finds all organisations who have a parent of the passed organisation type
```
xsrust: Thanks for removing block comments

app/views/layouts/_navigation.html.erb
```diff
@@ -1,74 +1,81 @@
...
-        <% end %>
+    <% end %>
+
+
+    <li class="highlight-color-button left-indent<%= ' active' if isActivePage(root_path) %>">
+      <%= link_to _('My Dashboard'), plans_path %>
+    </li>
+
+    <% 
+    if current_user.can_org_admin? && !current_user.org_id.nil? && !action_name.include?("admin_")
```
xsrust: This is much cleaner/more readable than the existing code for redirecting to the org-admin pages

xsrust:
A `rake gettext:find` may be needed as well, there seem to be quite a few new strings added.

app/views/layouts/_navigation.html.erb
```diff
@@ -1,74 +1,81 @@
...
-    <!-- roadmap page -->
-      <li <%= 'class=active' if isActivePage(roadmap_path) %>>
-        <%= link_to _('Future plans'), roadmap_path %>
+    </ul>
+  </li>
+  
+  <% if user_signed_in? %>
+  
+    <!-- Navigation for organisation admin -->
+    <% if (user_signed_in? && current_user.can_org_admin? && ( action_name.include? "admin_" ) ) %>
```
jollopre: user_signed_in? is checked in line 18 already, there is no need to check again since you are within 'then' block already.

app/views/layouts/_navigation.html.erb
```diff
@@ -1,74 +1,81 @@
...
       </li>
-    
-      <% if !user_signed_in? %>
-        <% if MANY_LANGUAGES %>
-            <!-- language dropdown -->
-            <li class="language-switcher">
-              <a class="dropdown" id="change-language" href="#">
-                <%= _('Change language') %> <span class="caret"><span>
-              </a>
-
```
jollopre: I did not know that we are removing the dropdown of languages. As far as I know, we have the option on the user's profile, but what about anonymous users?

app/controllers/application_controller.rb
briri: Yes I agree. In standard Rails weirdness though, helpers are available to views but not easily available to controllers so the old standard was to place anything shared by controllers into the base application_controller. 
It looks like they have switched to something called 'concerns' instead as of Rails 4 which seems to play the role of a helper module. I'll read up on it and switch some of the functions in the application_controller over. We should be moving in that Rails 4+ direction.

app/controllers/annotations_controller.rb
briri: Yes indeed and now that I'm looking at it 'example answer and guidance' should probably be in a `_('...')` fastgettext call

app/views/layouts/_navigation.html.erb
briri: Just wanted to be extra sure that they were signed in ;)
Good catch. I'll update it.

app/views/layouts/_navigation.html.erb
briri: They have the change language dropdown in the upper right corner next to the Signin/Logout links.

app/controllers/application_controller.rb
briri: Did some more reading on this and I can't see a use case where we would want to generate these standard error/success messages outside of a controller. Most of the controllers call these methods so I think it makes sense to leave them in the base application_controller like we've done with the auth, previous path and localization stuff.

Issue #148, opened 2017-03-03
Copy template option to create a new one


stephaniesimms:
Copy template option to create a new template 
* do not allow cloning/copying for funder templates (but do allow for Super Admin)
* do allow cloning for “Own template” and Default template (= DCC template, this option added to Finnish instance)

Pivotal ticket: https://www.pivotaltracker.com/story/show/119417325

Screenshot of DMPonline Admin area where users can currently: create a template, edit Own templates, customize Funder templates
![dmponline_dashboard_admin2](https://cloud.githubusercontent.com/assets/12703638/23565361/2c2e719e-0002-11e7-8fbd-a0106daa7568.png)

Add option to copy Own template on Admin templates page
![dmponline_dashboard_owntemplates](https://cloud.githubusercontent.com/assets/12703638/23565366/2f6fe2c0-0002-11e7-96c6-d19579a9c336.png)

stephaniesimms:
For the UI, here's a screenshot of the copy/clone plan feature. There are many plan-related actions that resulted in UX recommending a dropdown menu, which now includes a "Duplicate" action. For the template actions I don't think we need the dropdown. It should work to just add "Duplicate" next to "Edit" in the Actions column for the Own templates table.

![screen shot 2017-06-27 at 4 16 27 pm](https://user-images.githubusercontent.com/12703638/27614080-31faef6e-5b54-11e7-9307-8df47f5a29fc.png)

stephaniesimms:
The functionality to copy a template works as expected. We haven't addressed the Admin UI/UX yet, but @JEK-III can you advise about the table layout here? 

- for consistency with the plans area, the action should be "Make a copy" (instead of "Copy")
- the table maybe needs a dropdown list as in the plans table..?
- alternatively, all tables for templates and guidance could be reformatted w/more consistent column widths, etc.

Templates
![screen shot 2017-07-17 at 8 42 02 am](https://user-images.githubusercontent.com/12703638/28276987-5aadc46c-6acd-11e7-813b-020e9b175f11.png)

Guidance
![screen shot 2017-07-17 at 8 43 32 am](https://user-images.githubusercontent.com/12703638/28277002-64e23b0c-6acd-11e7-8cb5-7da740542f65.png)

stephaniesimms:
@sjDCC please confirm that copy template functionality works as desired before moving this along. I think it makes sense to address table layout/UI things in a separate issue. Closing this one.

sjDCC:
[...]

PR #504, opened 2017-07-10
Adds copy template option #148

kmcgre:


xsrust:
The logic all looks good to me.  It would be useful to confirm with @briri that the error messaging follows the new formatting he has implimented in his work last week.

app/controllers/templates_controller.rb
```diff
@@ -323,4 +323,29 @@ def admin_template_history
...
+    new_copy.dmptemplate_id = loop do
+      random = rand 2147483647
+      break random unless Template.exists?(dmptemplate_id: random)
+    end
+
+    if new_copy.save
+      flash[:notice] = 'Template was successfully copied.'
+      redirect_to admin_template_template_path(id: new_copy.id, edit: true), notice: _('Information was successfully created.')
+    else
+      flash[:notice] = failed_create_error(new_copy, _('template'))
```
briri: Change this to `flash[:alert]`. We have a div on the main `layouts/application.html.erb` that will display this and has all of the aria markup to announce it correctly to screen readers.

briri:
Just update that one line. Otherwise everything looks good!

app/controllers/templates_controller.rb
```diff
@@ -323,4 +323,29 @@ def admin_template_history
     @current = Template.current(@template.dmptemplate_id)
   end
 
+  # PUT /org/admin/templates/:id/admin_copy
+  # -----------------------------------------------------
+  def admin_copy
+    @template = Template.find(params[:id])
+    authorize @template
+
+    # Unpublish the live version
```
vyruss: Perhaps remove this comment line as it may be confusing

Issue #141, opened 2017-03-02
Switch existing help/tooltips from click to hover events


briri:
Tooltips: user has to click to turn these on/off and screen can get crowded if they are all are on. It is a standard and would be preferable if the tooltips appeared when the user hovers over the item with their mouse.

From pivotal: https://www.pivotaltracker.com/story/show/131576749

briri:
Just focus on non org-admin/super-admin pages for this first pass

briri:
@JEK-III should we display tooltips like the example below when the user hovers over the item or when the item gains focus? My suspicion is that on focus is preferable for accessibility since some users do not use a mouse.

JEK-III:
@briri On focus– for exactly that reason.

vyruss:
Agree with on focus - as some people will use this on a touchscreen enabled device which does not support hover.

stephaniesimms:
This appears to be working as expected. Closing issue.

PR #511, opened 2017-07-13
popover and tooltip triggered by focus and hover. DMPRoadmap/roadmap#141

jollopre:
The code implemented works for both:
- Users without mouse, in that case focus/blur is triggered
- Users with mouse, in that case mouseenter/mouseleave is triggered. Note, I have not used vanilla 'hover' to avoid duplicating the anonymous functions or creating named functions handlers for both.

We are using links to trigger popover and tooltips which requires to preventDefault behaviour. Moving to buttons would remove the need of attaching click handlers, although require changes at many files.

### DIFF
--- a/app/views/templates/_edit_template.html.erb
+++ b/app/views/templates/_edit_template.html.erb
@@ -12,7 +12,7 @@
       <%= f.label _('Description'), for: template.description %>
       <%= text_area_tag("template-desc", template.description, class: "input-large tinymce") %>
       <div class="inline">
-        <a href="#" data-toggle="popover" rel: "popover" data-html: "true" role="button" 
+        <a href="#" data-toggle="popover" rel="popover" data-html="true" role="button" 
            data-content="<%= _('Enter a description that helps you to differentiate between templates e.g. if you have ones for different audiences') %>" aria-label="<% _('More information: Template descriptions') %>">
         <span class="fa fa-question-circle" aria-hidden="true"></span></a>
       </div>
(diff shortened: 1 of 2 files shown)


lib/assets/javascripts/dmproadmap/utils.js
```diff
@@ -20,28 +20,29 @@ $(document).ready(function(){
...
-    if($(this).attr('data-content') != undefined){
+  $("[data-toggle='tooltip']").on('click', function(e){
+    e.preventDefault();
+  });
+  $("[data-toggle='tooltip']").on('focus mouseenter', function(e){
+    e.preventDefault();
+    if($(this).attr('data-content') !== undefined){
       var y = $(this).width() + 35;
-      $(this).parent().append('<div class="tooltip-message" style="left: ' + y + 'px;">' + $(this).attr('data-content') + '</div>');
+      $(this).after('<div class="tooltip-message" style="left: ' + y + 'px;">' + $(this).attr('data-content') + '</div>');
```
briri: Yes, that's cleaner and simpler

app/views/templates/_edit_template.html.erb
```diff
@@ -12,7 +12,7 @@
       <%= f.label _('Description'), for: template.description %>
       <%= text_area_tag("template-desc", template.description, class: "input-large tinymce") %>
       <div class="inline">
-        <a href="#" data-toggle="popover" rel: "popover" data-html: "true" role="button" 
+        <a href="#" data-toggle="popover" rel="popover" data-html="true" role="button"
```
briri: good catch.

lib/assets/javascripts/dmproadmap/utils.js
jollopre: it was needed too because focus and mouseenter might be both triggered which would create the div twice.

Issue #497, opened 2017-07-05
Revise Share plan page


stephaniesimms:
UX wireframe @JEK-III - http://h90c3m.axshare.com/#g=1&p=write_plan
Related issues for other write plan tabs #494 #495 

- Add plan visibility settings to top of page; blue notepad mentions outstanding need to determine threshold for adding "public" and "org" DMPs to list - TBD
- Revise Manage collaborators table
- Revise Invite collaborators section

briri:
The plan visibility section should be copied over from the edit plan details page. This section should be disabled if `plan.is_test?`.

briri:
Just need to wire up the JS logic for the visibility section. 

![screen shot 2017-07-11 at 3 59 57 pm](https://user-images.githubusercontent.com/1204467/28094533-13d2042e-6652-11e7-9270-90cf8ee5f4d5.png)

stephaniesimms:
Functionality works as expected w/exception of related issue #516 for test plans

@JEK-III please confirm or add to the list of style changes:
- Remove boldface for field names
- Remove colons for field names
- Align everything on this page w/the "Add collaborator" button at the bottom (so all headings, text, table should shift to the right
- Font size for headings should be smaller (refer to wireframes)

sjDCC:
There are still colons when sharing a plan but I guess these may be addressed as part of global styling so safe to close out here?

I also wonder if we should change 'Read Only' to lower case 'o' since that seems to be our norm elsewhere

Page looking good though

![capture](https://user-images.githubusercontent.com/6169940/28475524-8493d8e6-6e44-11e7-8557-d7a8116227ab.PNG)

stephaniesimms:
I concur w/ @sjDCC suggestions above re 
- removing boldface and colons 
- and changing to "Read only"

These are documented in the general styling ticket so maybe only the second item needs to be addressed here.

stephaniesimms:
@JEK-III and @briri - please update wireframe and application to say "Read only"

@briri - align all elements on this page per wireframe (to the inner rule established by the radio buttons and Email box below) 

![screen shot 2017-07-27 at 10 44 13 am](https://user-images.githubusercontent.com/12703638/28684263-9cfa1fd0-72b8-11e7-9529-8ebd53433e86.png)

briri:
[...]

Issue #491, opened 2017-07-05
Revise Project Details page 


stephaniesimms:
UX wireframes @JEK-III : http://h90c3m.axshare.com/#g=1&p=write_plan
Functional Specification (global spec for all write plan pages): Implement with a non-table layout; attach labels to input fields
Related issues for write plan pages #494 #495 #497 #498 

Revise the main plan writing pages per wireframes. blue notepad functional spec for Page Title at top of page, use "Project Title" field on the Create Plan page

First tab is renamed "Project Details." Blue notepad icons in wireframe contain functional specs:
- prepopulate Project Title and Funder from Create plan page
- prepopulate Principal Investigator Name, ORCID, email w/info in user profile
- check Data Contact person same as PI (default); if user unchecks display Name and Email fields for different Data Contact

Guidance selections move to right of Project Details, also described in related issue #495 for Write Plan page. 
Note blue notepad functional specs:
- guidance choices to display 
- add selector box at bottom to search for additional guidance from Org list; options to type ahead or "See full list"

briri:
I started work on this one a while ago but then refocused to finish the CSS rewrite. Here's a list of what has been done so far:
- Added a data_contact_email and data_contact_phone field to the plans table and page.
- Added rake task to extract emails and phone numbers from the existing data_contact field
- Switched over from tables to fieldset in the form
- Added a visibility section (has since been moved to the 'share' tab)

@sjDCC and @stephaniesimms the latest wireframe has removed the data contact phone number. There are valid phone numbers though in the dmponline data. They are currently stored in the one data_contact field (e.g. `John Doe, t: +12345678, e: john.doe@institution.org`). I have a script that was extracting these values and placing them in the appropriate field. Should I update that script to leave the phone number in the name field or does it make sense to add a phone field back to the page. 

Things left to do:
- Move the visibility section to the 'share' tab
- Remove the data_contact_phone field from the DB and add any new ones (wait for response before removing the column) and update rake task 
- Copy over the 'mock plan' checkbox and JS from the create plan page
- Implement the guidance portion

briri:
![screen shot 2017-07-11 at 3 56 29 pm](https://user-images.githubusercontent.com/1204467/28094473-a756d6da-6651-11e7-9cf4-42609b7778ca.png)

[...]

Issue #175, opened 2017-03-14
Provide users with consistent success/failure messaging


briri:
Each controller method handles success/failure notifications in different ways. Some do not provide any messaging!

We are now using ActiveRecord validations which place errors into the model's error array. Rails is designed to bubble validation errors up to the UI via this array. We should build a helper method that accepts an ActiveRecord model and will provide a standard success/failure message along with any specific model errors (e.g. 'invalid email address', 'title cannot be blank', etc.)

We should also follow a pattern like this: https://mattbrictson.com/dynamic-rails-error-pages for displaying friendly HTTP 401, 403, 404 and 500 pages.

Consult with UX for styling and placement of these messages.

briri:
related to #173

briri:
@JEK-III and @rhu001 please review the system's current success/failure messages that appear when you try to save changes on different pages. We should provide users with something consistent.

briri:
The error messaging has been consolidated into the following message formats:

If the field was empty or blank and it should not have been the message will be: 
```
Could not [create/update] your [object]
[field] - must not be [blank/empty]

Example:
  Could not update your Plan
  Title - must not be blank.
```

For other types of errors we use a similar format but show the value the user entered instead of the field name:
```
Could not [create/update] your [object]
'[value]' - [message]

Example:
  Could not update your Organization
  'erg334b' - is not a valid email address
```

briri:
Consolidate the success messaging like we did for the error messaging.

Go back through and make sure that errors are being assigned to the `flash[:alert]` and success messages are assigned to `flash[:notice]`. We now display different icons and potentially colors for each one.

briri:
related to #457

briri:
@rhu001 and @JEK-III can you provide us with recommendations for the location and styling of error messages and notifications?

rhu001:
Hi @briri is there a place where you have a listing of some or all of the various messages and notifications currently available so we get a sense of the different lengths of messages and types of messages?

briri:
Hey @rhu001 this isn't an exhaustive list, but it should cover the majority of messages:
```
Information was successfully updated.
Could not create your %{o}.      <--- where 'o' is the object type (e.g. Plan, template, etc.)
Could not update your %{o}.
Could not delete your %{o}.

Organisation was successfully updated.
There seems to be a problem with your logo. Please upload it again.
No institutions are currently registered.
Your institution does not seem to be properly configured.
Please choose an institution

[...]

PR #512, opened 2017-07-13
Refactored plan edit details and share pages.

briri:
- Refactored the Project details page per #491 
- Refactored the Plan sharing page per #497 
- Consolidated alert/notice/tooltip styling and site's background color to the dmproadmap.scss file so that these settings are more easily changed when branding the site
- Updated alert/notice message colors per #175


app/policies/plan_policy.rb
```diff
@@ -56,6 +56,10 @@ def visibility?
     @plan.administerable_by?(@user.id)
   end
 
+  def set_test?
+    @plan.administerable_by?(@user.id)
```
xsrust: Currently update is set as @plan.editable_by?(@user.id)  

Which one is the correct behavior?
I  can see a case for only owners/co-owners to be able to set visibility, test status, and other plan details

app/views/plans/_plan_title.html.erb
```diff
@@ -1,8 +1 @@
-<h1><%= plan.title %></h1> 
-
-<% if plan.visibility == 'is_test' %>
-  <div class="roadmap-info-box">
-    <i class="fa fa-exclamation-circle" aria-hidden="true"></i>
-    <span><%= _('This is a') %> <strong><%= _('test plan') %></strong>.</span>
-  </div>
-<% end %>
+<h1><%= plan.title %></h1>
```
xsrust: Does this still need to be a partial? 
is the intention down the line to move back towards a conditionally different title?

app/controllers/plans_controller.rb
```diff
@@ -38,9 +38,13 @@ def new
   def create
     @plan = Plan.new
     authorize @plan
-
+    
     @plan.principal_investigator = current_user.surname.blank? ? nil : "#{current_user.firstname} #{current_user.surname}"
-    @plan.data_contact = current_user.email
+    @plan.principal_investigator_email = current_user.email
+    
+    orcid = current_user.identifier_for(IdentifierScheme.find_by(name: 'orcid'))
```
vyruss: @xsrust I think the identifier_schemes table is convenient because we can tack on descriptions, urls and other values

xsrust: I agree.  I was jusg suggesting that we save the `IdentifierScheme.find_by(name: 'orcid')` within the model for identifier_schemes instead of running a query each time we want to select it.  just like Perms, the content of the record for the orcid identifier scheme will remain constant while the tool is used

app/views/plans/_edit_details.html.erb
```diff
@@ -1,95 +1,162 @@
...
-    </fieldset>
-    
     <div class="bordered left-indent">
-      <div class="bordered"><strong><%= _('Plan Guidance Configuration') %></strong></div>
+      <h2><%= _('Plan Guidance Configuration') %></h2>
       <p><%= _('To help you write your plan, %{application_name} can show you guidance from a variety of organisations.') % {application_name: Rails.configuration.branding[:application][:name]} %></p>
-      <br />
-      <p><% _('Select up to 3 organisations to see their guidance.') %></p>
-      <% #@important_ggs.inspect %>
+      <p><%= _('Select up to 3 organisations to see their guidance.') %></p>
```
vyruss: I'm not sure about the 3 guidance limit... perhaps we'll need to check that with users.

app/controllers/plans_controller.rb
briri: agreed @xsrust. I'll open a new ticket for that. A constant would work for this. You need to update the users table, add routes, etc. anyway to add a new identifier scheme, so it requires the site to be bounced.

app/controllers/plans_controller.rb
```diff
@@ -152,43 +164,42 @@ def show
...
-    @selected_guidance_groups = @plan.guidance_groups.pluck(:id)
-    @based_on = (@plan.template.customization_of.nil? ? @plan.template : Template.where(dmptemplate: @plan.template.customization_of).first)
-
-    flash[:notice] = "#{_('This is a')} <strong>#{_('test plan')}</strong>" if params[:test]
-    @is_test = params[:test] ||= false
-    
-    respond_to :html
-  end
-
+#  def edit
```
briri: yep. will do.

app/policies/plan_policy.rb
briri: not sure. I'll pose the question to @sjDCC and @stephaniesimms

app/views/plans/_plan_title.html.erb
briri: nope. i'll get rid of the world's smallest partial ;)

app/views/plans/_edit_details.html.erb
briri: I didn't actually set the limitation in the code yet. need to ask @sjDCC and @stephaniesimms about it.

xsrust: Durring the call with UX on Wednesday, it was mentioned that 3 was an arbitrary limit due to spacing of the new guidance tabs within plan edit.  If more than 3 are needed, @JEK-III said he would be able to re-design the guidance display to display more, either through a second row of tabs or otherwise.

briri: thanks for the clarification. I'll add some JS logic to prevent the selection of more than 3?

app/controllers/plans_controller.rb
briri: see ticket #513

lib/assets/javascripts/dmproadmap/forms.js
```diff
@@ -1,39 +1,56 @@
...
+    type: method,
+    data: data,
+    contentType: 'application/json',
+    accepts: 'application/json'
+  }).done(function(data){
+    $("#notification-area span:not(.aria-only)").html(data['msg']).css('width', 'auto')
+          .attr('role', (data['code'] === 1 ? 'status' : 'alert'))
+          .attr('class', (data['code'] === 1 ? 'roadmap-info-box' : 'roadmap-alert-box'))
+          .parent().css('visibility', 'visible').fadeIn();
+  });
```
jollopre: I encourage you to use .then(resolved,rejected) rather than just done, otherwise for any error the user will not be informed at all.

lib/assets/javascripts/dmproadmap/forms.js
```diff
@@ -1,39 +1,56 @@
...
+    }else{
+      err.html(errorMessage).attr('role', 'tooltip');
+      $(autocomplete).addClass('red-border');
+    }
+  }
 }
 
 // ---------------------------------------------------------------------------
-function toggleInputError(input, errorMessage){
+function toggleInputError(input, errorMessage, allowBlank = true){
```
jollopre: I'm not aware of default params working without using a transpiler

lib/assets/javascripts/views/plans/index.js
```diff
@@ -1,23 +1,12 @@
...
-            .css('width', 'auto').parent().css('visibility', 'visible');
-        $(self).parent().siblings("#visibility-" + id).html(__('Private'));
-      }else{
-        // Display an error message
-        $("div.roadmap-alert-box span:not(.aria-only)").show().html(data['msg'])
-            .attr('role', 'alert').css('width', 'auto').css('visibility', 'visible');
-        e.preventDefault();
-      }
-    });
+    remoteSave("/plans/" + id + "/set_test", 'POST', JSON.stringify(params));
```
jollopre: remoteSave does not seem like a proper name for this function since is not only "trying to save something on the server" but also changes visibility of elements.

lib/assets/javascripts/dmproadmap/forms.js
briri: will do

lib/assets/javascripts/dmproadmap/forms.js
briri: thanks. didn't realize it was not reliable

lib/assets/javascripts/views/plans/index.js
briri: agreed. I went back and forth with 3-4 names before finally giving up. Any suggestions? I didn't want to use something too generic that could cause confusion with other libraries.

jollopre: I would get the body of the done function and would refactor in a new function. Perhaps animateNotification...

app/policies/plan_policy.rb
briri: they both agreed that owners and co-owners should be able to change visibility and set the test flag

briri:
- removed unecessary endpoints from plans_controller
- removed _plan_title.html.erb
- added JS logic to prevent user from selecting more than 6 guidance groups (the limit discussed in the UX meeting)
- switched the `remoteSave` ajax wrapper method to `asyncRequest(params, callbacks)`. The `params` value should be a hash of jquery ajax parameters (e.g. `{url: '/my/path', type: 'POST', ...}`) and the `callbacks` should be a hash containing `{success: functionA, failure: functionB}` The method calls the `animateNotification()` function which sets the accessible alert/notice message. See `lib/assets/javascripts/views/plans/index.js` and the `set_test` endpoint on the plans_controller for an example of usage. If all you want to do is set the notification message you can send an empty hash to the function.
- Created a new animateNotification function
- Fixed issue with viewing the read-only version of the plan details
- Updated the 'my dashboard', 'project details' and 'plan share' pages to work with these changes

xsrust:
All requested changes made or tickets created so this is looking good to me

PR #515, opened 2017-07-17
Minor adjustment to get autocomplete box's errors to appear properly

briri:
Also includes a text change on a link in the shib DS page


briri:
Rebased with the latest changes from @kmcgre and fixed typo causing error in accessible_combobox partial.

Issue #498, opened 2017-07-05
Revise plan Download (formerly Export) page


stephaniesimms:
UX wireframes @JEK-III : http://h90c3m.axshare.com/#g=1&p=write_plan
Functional Specification (global spec for all write plan pages): Implement with a non-table layout; attach labels to input fields
Related issues for write plan pages #491 #494 #495 #497

- Tab renamed "Download"
- Button renamed "Download Plan"
- Replace included elements options w/new, abbreviated menu of options; blue notepad w/functional specs - Fieldset legend for checkboxes
- by default export all answers (no tickbox for this)
- Project Details encompasses all info captured on Project Details page (still to-do - coversheet design for this info, related to #436)
- note functional spec for when/how to display unanswered questions tickbox
- add tooltip beside supplemental sections tickbox
- set default to most common file format (assuming PDF @sjDCC ? if not PDF most likely DOCX)

briri:
@JEK-III there is no message displayed when I click on the blue notepad icon next to the supplemental sections checkbox. Are 'supplemental sections' customized sections of a funder template? If so should this checkbox be hidden when the plan does not have customized sections?

Use the following format for the tooltip button next to the supplemental sections checkbox:
```ruby
<a href="#" data-toggle="popover" rel: "popover" data-html: "true" role="button" 
             data-content="<%= _('Here's our helpful message') %>" 
             aria-label="<%= _('More information: Supplemental sections') %>">
          <span class="fa fa-question-circle" aria-hidden="true"></span></a>
```

xsrust:
@JEK-III
As brian mentioned above, the notepad icon next to supplemental sections is blank.  Is the intended behavior to hide customized sections, and only display on customized templates?

Additionally, how are plans with multiple phases supposed to be handled? I dont know of any with more than 3 phases in the current dataset.  Perhaps this could be a drop-down box which appears when the plan has multiple phases?

JEK-III:
@xsrust 
The intended behavior is to only show that checkbox on customized templates. I adjusted the way elements are layered in the wireframe so the annotation should work now.

As far as multi-phase templates go, I see nothing wrong with the current approach of having separate tabs for each phase, labelled with the name of the phase. (In which case there will be no tab labeled "write plan")

xsrust:
@JEK-III Thanks for that.  The wireframe looks good to me on the guidance/annotation front.

[...]

Issue #346, opened 2017-05-22
Show number of users


dsisu:
Hi all

Please provide some on-screen statistics for org_admin users. For example, when they click in the 'Users' tab, they should be able to see the total number of users. Could work the same with templates and guidance but those are easier to count at a glance. I know org_admins have asked to see total number of users and plans so we could start with that.

Thank you!

briri:
@sjDCC and @stephaniesimms can this be closed out now that we have the Admin -> Users page and the statistics page?

stephaniesimms:
yes i think we can close this now

Issue #154, opened 2017-03-03
Generate usage stats


stephaniesimms:
Adapt the prototype DMPTool dashboard
* carry this over to new platform as a replacement for DMPTool API queries to get usage stats
* evaluate DMPonline/Roadmap API queries for reuse in usage dashboard
* provide ability to export CSVs
* add visualizations that show: 1) time, 2) how org stats compare to global stats

Org admins want to know the following (per @sjDCC comments on Edinburgh reports):
* new users per month
* total users for their org
* total DMPs per month
* DMPs per template per month

Example of what this could look like in Roadmap - for discussion. Note that 'Date Range' field should be constrained to monthly, annual, and/or total reports. Under 'Institution' (change to Org) a user can select to display their own Org usage or "All Orgs."
![screen shot 2018-01-22 at 4 47 34 pm](https://user-images.githubusercontent.com/12703638/35252260-778a7c40-ff94-11e7-9181-1a03e8d53f87.png)

Screenshot of DMPTool Usage dashboard
![screen shot 2017-03-03 at 12 45 44 pm](https://cloud.githubusercontent.com/assets/12703638/23568441/5b1e51c4-000f-11e7-9be9-7d28d3278d44.png)

sjDCC:
Details on the API can be seen at: https://github.com/DMPRoadmap/roadmap/wiki/API-Documentation

You can:
-  Query the number of registered users within an optional date range
-  Query the number of plans using a template owned by your organisation
-  Retrieve number of plans organised by template
-  Retrieve metadata on all plans from an organisation’s users

For reference the stats we pull for Edinburgh each month cover:
- new users per month
- total users for the org
- total number of DMPs per month**
- DMPs per template per month**

** these last two are in a table with funders along one side and months across the other and running total for DMPs per funder and DMPs per month at ends

It would be good to set these stats into wider context as the DMPTool dashboard above does

briri:
@stephaniesimms and @sjDCC is allowing a custom date range instead of predefined combobox options (e.g. last 7 days, annual, etc.) ok? After discussing in the dev meeting today we have determined that it would be easier to just let the user select their own range.

stephaniesimms:
that sounds great to me. admins have asked for this and will appreciate being able to tailor the reports using a specific date range. if it's possible to default to something like "last 30 days" that would be good and then allow them to customize from there.

sjDCC:
[...]

PR #519, opened 2017-07-19
Xsrust/498

xsrust:
Updates to the plan downloads page according to ticket #498
Also adds internationalization which I had earlier forgotten to the notification prefs page


app/controllers/plans_controller.rb
```diff
@@ -235,9 +237,8 @@ def export
     @plan = Plan.find(params[:id])
     authorize @plan
 
-    # If no format is specified, default to PDF
-    params[:format] = 'pdf' if params[:format].nil?
-
+    # We should re-work this into something more useful than creating a new one
+    # every time a plan gets exported
```
vyruss: Yes I think this all ties in with the statistics discussion. Tagging issues #154 and #346

app/models/exported_plan.rb
```diff
@@ -99,24 +99,40 @@ def title
...
-        if questions.present?
-          questions.each do |question|
-            answer = self.plan.answer(question.id)
-            q_format = question.question_format
-            if q_format.option_based?
-              options_string = answer.question_options.collect {|o| o.text}.join('; ')
-            else
-              options_string = ''
-            end
+      if question_headings
```
vyruss: Not convinced at all about CSV format's usefulness at this stage - we've had fewer than 240 downloads in total and I'm pretty sure more than half of those are us testing it...

app/views/users/_notification_preferences.html.erb
```diff
@@ -12,49 +12,49 @@
     <div class="form-input">
       <%= hidden_field_tag 'prefs[users][new_comment]', false %>
       <%= check_box_tag 'prefs[users][new_comment]', true, @prefs[:users][:new_comment] %>
-      <%= label_tag 'prefs[users][new_comment]', 'A new comment has been added to my DMP', :class => 'checkbox-label' %>
+      <%= label_tag 'prefs[users][new_comment]', _('A new comment has been added to my DMP'), :class => 'checkbox-label' %>
```
briri: Thanks for catching these.

Issue #242, opened 2017-04-12
Co-owners should not be able to delete the plan


briri:
This is an issue that apparently exists in the DMPonline_V4 codebase but we should correct it in Roadmap.

If you share your plan with someone else and give them administrator access, they have the ability to delete the plan. The plan should not get deleted in this situation, we should just remove their access to the plan so that it falls off of their 'View Plans' page.

sjDCC:
I'm happy for this to be addressed in a later release. Doesn't strictly need to be in MVP v.0.3. There's a warning at present which isn't ideal but works as a stopgap.

briri:
Ah ok. Changing it to MVP instead of MVP-0.3 and moving it off the board for now

stephaniesimms:
I can't test this @briri because I can't access any new user accounts. The invitation email contains a link that gives me this error (screenshot 1). And when I try to login directly, I get the second error.

![screen shot 2017-06-01 at 2 10 01 pm](https://cloud.githubusercontent.com/assets/12703638/26700924/07f3dd46-46d4-11e7-82ca-f66ce9e6ea39.png)

![account-creation-user-invited-error](https://cloud.githubusercontent.com/assets/12703638/26700911/fd2501e2-46d3-11e7-8bd9-67a7d6def7a6.png)

stephaniesimms:
This is working as expected now. The only change is cosmetic: the table formatting needs to be fixed to display only 2 columns (screenshot below of current format)

![screen shot 2017-06-05 at 1 31 26 pm](https://cloud.githubusercontent.com/assets/12703638/26801995/b772b8f6-49f3-11e7-9627-5b8773e43a94.png)

stephaniesimms:
Table formatting looks great now - standardized 3 column widths. All works as expected.

sjDCC:
Share page isn't giving me options at the moment so can't test. Have two accounts ready though so just drop me a line when good to go.
![share](https://user-images.githubusercontent.com/6169940/27614099-6bb89bf0-5b97-11e7-8fb2-9baca121d60f.PNG)

FYI, the 'share' action from the dropdown list on 'my plans' page is throwing one of the red error screens. Assume this is because there isn't content in page.

stephaniesimms:
Retested and all works as expected for plan sharing functions. Will investigate plan delete warning separately. Closing issue.

stephaniesimms:
[...]

Issue #492, opened 2017-07-05
Error generating Public DMP export for plans with multiple phases


briri:
The current `public_export` method in the plans_controller generates an error if you try to export a plan whose template has multiple phases (e.g. Horizon2020 - see "Jose's Plan" on roadmap-stg). The method attempts to loop through the phases and generate a pdf for each one which results in the following error from the controller:
` AbstractController::DoubleRenderError in PlansController#public_export `

@sjDCC @stephaniesimms How do we want to handle the export of a public DMP with multiple phases (e.g. Horizon2020)? The plan export for logged in users allows them to select which phase they want to export and a PDF is generated for a specific phase. The public DMPs page though doesn't currently offer this. Do we want to just generate one PDF that contains all phases?

related to #84

stephaniesimms:
For multi-phase templates the visibility setting should apply to all phases (we can add messaging to make this clear). So it makes sense to me that we would provide a single PDF that contains all completed phases in the public DMPs list. @sjDCC what do you think?

sjDCC:
Yes, I think if there is more than 1 completed phase we should include them all in public DMP. 

When funders ask the same questions at each phase (i.e. when it's more a case of different versions of the same thing / an evolving, living plan) we may  want a different approach, but we can review this when we add proper versioning post MVP. For now these examples (e.g. H2020) are represented as phases too.

xsrust:
@sjDCC @stephaniesimms You've both mentioned that all 'completed' phases should be exported.  As far as I know, we dont have any way to track what the user has and hasn't completed other than checking what they have answered.  Should the default behavior be to export all phases, or only those with answered questions?

Also, given the rework of Downloads in #498, What settings apply to public DMP export.  i.e. should we export questions as headings, do we export unanswered questions with the text "question unanswered" for the answer, Do we export the project details, and should we export customized sections?

sjDCC:
Like you say, there's not currently a marker or flag of completeness, and I think the progress bar is an accumulation of all questions over all phases so that's not a useful indication either... 

With that in mind, I would only export phases with answered questions (not automatically all phases) and would apply the following settings:
- do not include plan details as some of these will be added on the cover page #436 
- export section headings and questions (when answered)
- don't export unanswered questions if they can be easily filtered out
- do include cutomised sections since this is a public DMP not response to funder

What do you think @stephaniesimms?

stephaniesimms:
[...]

PR #523, opened 2017-07-20
Xsrust/mvp

xsrust:
Addresses #242, with implementation as detailed in the ticket

Note: this code will require a db migration to add the 'active' flag to the roles table


xsrust:
Forgot to check out a new branch before committing changes for #492.  

So now this PR includes the work for public DMP export with multiple phases

app/controllers/roles_controller.rb
```diff
@@ -66,6 +66,21 @@ def destroy
     redirect_to controller: 'plans', action: 'share', id: @role.plan.id
   end
 
+  def archive
```
vyruss: I'm not 100% on the `archive` action name for this. Perhaps `deactivate` is a bit clearer on what it does - "remove" is certainly not accurate (even though it eventually refers to "remove plan" action)

app/policies/role_policy.rb
```diff
@@ -19,4 +19,8 @@ def update?
   def destroy?
     @role.plan.owned_by?(@user.id)
   end
+
+  def archive?
```
vyruss: Again not 100% sold on the name

app/views/plans/index.html.erb
```diff
@@ -7,24 +7,24 @@
 
 <!-- if the user has projects -->
 <p>
-<% if @plans.count > 0 %>
+<% if @plans.length > 0 %>
```
vyruss: good change.

config/routes.rb
```diff
@@ -224,7 +224,11 @@
       end
     end
 
-    resources :roles, only: [:create, :update, :destroy]
+    resources :roles, only: [:create, :update, :destroy] do
+      member do
+        put :archive
```
vyruss: Same comment about action name.

app/policies/plan_policy.rb
```diff
@@ -9,47 +9,47 @@ def initialize(user, plan)
   end
 
   def show?
-    @plan.readable_by?(@user.id)
+    @plan.readable_by?(@user.id) && Role.find_by(user_id: @user.id, plan_id: @plan.id).active
```
briri: It makes sense to keep this active check out of the `readable_by?` method. Is the check necessary though? These plans are filtered out of the user's dashboard, so the likelihood of them hitting these policy checks is slim. Even if they did via a bookmark or something, they should technically be able to access them.

This is maybe a discussion for a separate ticket though. I could see use cases where we would send them to these endpoints to view the plan and restore/unhide them if we decide to offer that option.

app/models/role.rb
```diff
@@ -37,6 +38,10 @@ def access_level
     end
   end
 
+  def set_defaults
+    self.active = true if self.new_record?
```
xsrust: Thanks.  stole it from some guy's blog on best practices for setting defaults in rails

app/policies/plan_policy.rb
xsrust: The check is probably not needed, but If a user removes themself from a plan, and other users cant see that they can access the plan, I like ensuring that behavior in code.

xsrust:
Just realized while reading over your comments that I didn't filter the 'share plan' page to remove the roles which are not active.  I'll add that, make the recommended changes, and update the PR.

briri:
looks good @xsrust

Issue #518, opened 2017-07-18
Create Account: Fix styling/code issues for Organisation dropdown


briri:
- Use the auto-complete box (see `views/plans/new.html.erb` for an example) for the org dropdown list
- Update styling (add `left-indent` class, etc.)
- Make sure JS is still working to hide/show the other org box
- Make sure that org selection is saved properly

kmcgre:
@sjDCC @stephaniesimms 
org_id and other_organisation now being saved to database. There's no text box for other_organisation in the edit personal details page for the user so I'll add that.

What should happen if a user selects an organisation from the dropdown, clicks the link to show the other org box and enters an "other organisation"?

sjDCC:
Thanks @kmcgre Definitely good to list the other org value in a field under edit profile

I guess your second question is about the account creation process? If so, one should overrule / clear the value for the other as people can only either a) select an org from the list or b) enter an 'other org' value. They can't have an entry for each.

Just checked on DMPonline live site and this is how it works. If I select an org from the list (screengrab 1) and then click the link for "My organisation isn't listed" the dropdown value clears and is changed to 'Other' and a box appears for entering the value (screengrab 2). The same works vice versa too.

![step1](https://user-images.githubusercontent.com/6169940/28415340-2740d11a-6d47-11e7-9736-e802099efdd9.PNG)
![step2](https://user-images.githubusercontent.com/6169940/28415347-2ae2e1a0-6d47-11e7-98a0-5631732ea6e3.PNG)

vyruss:
@sjDCC so we would want that behaviour on the Edit profile page too, correct?

sjDCC:
Yes, definitely. We only ever want them to have one value - an org from the dropdown list, or a text entry for 'other org'

stephaniesimms:
I'm not clear about the expected behavior - after testing here are some notes:
- I was just able to create an account without selecting an Org. I left the field blank and was not forced to select "Other" nor did I receive an error msg. When I navigate to the Edit Profile page the box for Org is blank (screenshot below)
- When I click "My organisation isn't listed" on the Create Account page and under the Edit Profile page nothing happens. Should it autopopulate the box w/"Other" when a user clicks this link @sjDCC ? 
- the only way to successfully save my profile is to type "Other" into the box after following the error prompt 
- We should also be careful if we implement autosave throughout the site (#450) that we make exceptions for things like this that require validation  

[...]

PR #529, opened 2017-07-21
Create Account: Fix styling/code issues for Organisation dropdown #518

kmcgre:


briri:
Just update the broken test and this will be good. It looks like maybe the accept_terms param needs to be updated to match what the form is sending back. The test is sending 'false' but the controller is doing a `!params[:accept_terms]` check.

```
  1) Failure:

RegistrationsControllerTest#test_user_receives_proper_error_messaging_if_they_have_not_accepted_terms [/home/travis/build/DMPRoadmap/roadmap/test/functional/registrations_controller_test.rb:26]:

--- expected

+++ actual

@@ -1 +1 @@

-"You must accept the terms and conditions to register."

+"Error processing registration. Please check that you have entered a valid email address and that your chosen password is at least 8 characters long."
```

kmcgre:
I was looking at that but the form can't be submitted if the accept checkbox isn't checked so is this test needed?

briri:
I think the test is beneficial. I'll fix on my end. Pulling this through and deploying so UAT can begin.

vyruss:
@briri @kmcgre I'm in the process of turning it into a capybara test so we can test before submitting

PR #538, opened 2017-07-24
Bugfix for Public Export

xsrust:
Adds authentication to public export to ensure exported plan is publicly_visible

### DIFF
--- a/app/controllers/plans_controller.rb
+++ b/app/controllers/plans_controller.rb
@@ -2,7 +2,7 @@ class PlansController < ApplicationController
   require 'pp'
   helper SettingsTemplateHelper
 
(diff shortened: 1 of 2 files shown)


xsrust:
Failing test is because of the implementation of this Patch:

I added an exception to allow for non-signed-in users that the plan passed to auth must be publicly_visible

The plans#index option does not pass a plan into auth, and so throws an error on the edge case of a non-signed-in user attempting to access the plans#index page.  

As discussed on slack, I will update tomorrow with a more-full-fledged approach to the auth on these public pages.

briri:
yes, this is a good approach to protecting plans via that public_export path.

Issue #436, opened 2017-06-16
Create cover page for Public DMPs PDFs


stephaniesimms:
Related issue #300 
DMPs that appear in the Public DMPs list should include a copyright statement on the cover sheet (statement below copied from DMPTool public DMPs cover sheet):

Copyright information: The above plan creator(s) have agreed that others may use
as much of the text of this plan as they would like in their own plans, and customize
it as necessary. You do not need to credit the creators as the source of the
language used, but using any of their plan's text does not imply that the creator(s)
endorse, or have any relationship to, your project or proposal.

xsrust:
The above text has been previously added to the public dmp export.

The default admin-details section has been removed according to comments in #492 .

@sjDCC @stephaniesimms, Which admin details, and what information(other than the above) should be included on the coverpage for the public PDF export?

sjDCC:
@stephaniesimms is out at a wedding til next so week so I suggest we just go with the existing DMPTool model for now. I've included a screengrab below, but you can see other examples at: https://dmptool.org/public_dmps

![public-dmp-cover-page](https://user-images.githubusercontent.com/6169940/28420220-58d23e32-6d59-11e7-85c4-2c7e0bfb68a2.PNG)

So I figure we need to include:
- Title
- Plan creator + co-owners (if any)
- Institution (from controlled dropdown list OR value from 'other org' field, OR 'Non Partner Institution' for US examples where they don't have a name recorded)
- Last modified (@briri does this mean the public DMP is automatically updated / regenerated any time changes are made?)
- Copyright notice (text already added)

I would be inclined to make a couple of additions / changes:
- I'd capitalise DMP in the provenance label to reinforce the acronym i.e. 'A Data Management Plan created using {TOOL_NAME}'
- I'd add the funder if there is one OR record the template name so it's clear what set of requirements it's based on
- I would add grant ID if this field has been completed to help link together outputs
- Possibly include the 'project abstract / plan description' if this has been filled in to give people an overview before reading plan

Build it as per DMPTool for now and @stephaniesimms can confirm preferences on other suggestions later. Thanks @xsrust

xsrust:
For the Affiliation: section, Which institution should be listed? Just that of the creator, or those of all co-owners?

sjDCC:
I would say just that of the creator. What do you think @stephaniesimms

stephaniesimms:
@sjDCC and @xsrust - I agree that just the creator's affiliation should be listed.

[...]

Issue #158, opened 2017-03-03
Export funder templates


stephaniesimms:
Build functionality to export funder templates only. Funder templates are available as public resources as in DMPTool.
- Add "DMP Template" under "Explore" menu tab on homepage as in UX wireframe @rhu001 http://h90c3m.axshare.com/#g=1&p=global_nav_variations   
- Create table for funder template info modeled on DMPTool ["DMP Requirements"](https://dmptool.org/guidance) (screenshot below)
- Table includes columns for Template name, Funder name, Links to funder requirements, Links to funder-provided sample plans (optional, if provided by funder)
- Include a tooltip for Sample plans with the following text: "Sample plans are provided by a funder, an institution, or a trusted party."
- Provide ability to export templates as DOCX and PDF files - preferred to have both, but minimum requirement is DOCX to provide an editable file in a user-preferred format
- Provide ability to search the list
- Do not include the A-Z links as shown in screenshot below

![screen shot 2017-07-18 at 2 17 10 pm](https://user-images.githubusercontent.com/12703638/28340556-0d00160e-6bc5-11e7-947e-557c08ec6d3e.png)

Open question about including guidance and/or adding separate guidance export functionality. This is more complicated to bundle w/template export since this should not require a user to be logged in. Therefore guidance needs to be selected separately for export (post-MVP).

Pivotal ticket: https://www.pivotaltracker.com/story/show/119417249

Screenshot of DMPTool homepage w/DMP Requirements and Public DMPs tabs - should be available without logging in
https://dmptool.org/guidance
https://dmptool.org/public_dmps
![dmptool_homepage](https://cloud.githubusercontent.com/assets/12703638/23570669/2878dfa4-001a-11e7-98d2-16b74ad19897.png)

xsrust:
Implemented:
 - index page for template export with template_name, org_name, and downloads fields
 - Docx/PDF export for templates
 - Searching the list via real-time filtering table

Discussion:
 - Currently no guidance is being exported with the PDF/DOCX
 - We could add in annotations/example answers, but the formatting of the documents will need to be considered, same with themed guidance
 - For option based questions, I am currently listing all options.  Should the question type additionally be listed, e.g. datepicker, multiselect, radio?
 - Other than adding centering to the title, I have not done much additional styling on the documents, so formatting/layout may need work.(currently mimics our other export's layout/look/feel)

Don't currently exist in Roadmap DB:
 - Funder Links
 - Sample Plans

[...]

PR #543, opened 2017-07-27
Xsrust/mvp #436, #158

xsrust:
Covers #436 and #158.  

Also re-factors public actions to a new controller and applies auth to that controller


app/controllers/plans_controller.rb
```diff
@@ -309,11 +313,18 @@ def public_export
...
+    # name of owner and any co-owners
     @creator_text = @plan.owner.name(false)
     @plan.roles.administrator.not_creator.each do |co_owner|
       @creator_text += ", " + co_owner.name(false)
     end
-    @affiliation = @plan.owner.org.abbreviation
+    # Org name of plan owner
+    @affiliation = @plan.owner.org.name
+    # set the funder name
+    @funder = @plan.template.org.funder? ? @plan.template.org.name : nil
```
briri: should we use `@plan.funder` here instead? The create plan process auto-populates that field now but is likely not applicable to existing data. I'll make a note in the ticket to have @sjDCC and @stephaniesimms review the different scenarios in UAT

briri:
Looks good @xsrust. I like the move to a dedicated controller/policy object for these public pages.

Issue #552, opened 2017-08-03
Setup Bootstrap grid system


briri:
Update the common partials in `views/layouts/` (`_header.html.erb`, `footer.html.erb`, `application.html.erb`) to use Bootstrap's grid system.

http://getbootstrap.com/css/#grid

Remove or replace the styling in `lib/assets/stylesheets/dmproadmap/base.css`

briri:
This ticket does not address any of the conversation taking place to determine what content the header should contain or how it should be styled. That will be covered in a future ticket. This ticket just refers to setting up the grid system. We should aim to support the original basic structure of the header:
```
                      |                                                                 |         PROFILE/LANGUAGE NAV
                      |          BANNER TEXT                               |
    LOGO        |                                                                 |        
                      |---------------------------------------|
                      |           MAIN NAVIGATION                       |
```

briri:
Pulled in Boostrap and updated the core layout files to use Bootstrap's grid system. Also went through and setup every page with a default of `<div class="row"><div class="col-md-12">...</div></div>` We can/should adjust as each page is refactored to work with Bootstrap's CSS and JS

The basic layout is setup as: 
```html
<html>
  <head> ... </head>
  <body>
    <header>
      <div class="row">
        <div class="col-md-4">... logo markup ... </div>
        <div class="col-md-4">... branding text / nav markup ... </div>
        <div class="col-md-4">... sign in/out nav markup ... </div>
      </div>
    </header> 

    <main>
      <div class="row">
        <div class="col-md-12">... page content ...</div>
      </div>
    </main>

    <footer>
      <div class="row">
        <div class="col-md-8">... footer menu markup ... </div>
        <div class="col-md-4">... logo markup ... </div>
      </div>
    </footer>
  </body>
</html>
```

The change in page structure has affected the existing styling as expected.
![screen shot 2017-08-07 at 1 09 01 pm](https://user-images.githubusercontent.com/1204467/29044443-520730de-7b74-11e7-8809-a9d679bdcef1.png)

stephaniesimms:
visual inspection on stage. closing issue.

PR #574, opened 2017-08-07
Setup Boostrap's grid system

briri:
Switched the main layout (`application.html.erb`, `_header.html.erb` and `_footer.html.erb`) files to use Bootstrap's grid system.
Then updated every page to do the same. See #552 for details

Also moved some JS that was attached to the edit profile page within a `<script />` tag to the page's JS file.


briri:
yes switching to container-fluid is an easy adjustment. you lose the centered page content look though.

Issue #554, opened 2017-08-03
Refactor CSS/JS on 'Contact Us' page (Core page)


briri:
Refactor the Contact Us page to use Bootstrap.

This page will serve as an example/reference for how to implement forms on our other pages.
http://getbootstrap.com/css/#forms
http://getbootstrap.com/css/#buttons

Remove or replace the functionality in `lib/assets/stylesheets/dmproadmap/forms.css` and `lib/assets/dmproadmap/javascripts/forms.js`

stephaniesimms:
visual inspection on stage. closing issue.

PR #582, opened 2017-08-11
DMPRoadmap/roadmap#554

jollopre:
Refactors contact_us view with Bootstrap
Uses ariatiseForm for augmenting control forms, validate value and add aria tags


lib/assets/javascripts/utils/ariatiseForm.js
```diff
@@ -0,0 +1,122 @@
...
+    });
+    var ariaInvalid=(function(value){
+        return { 'aria-invalid': value };
+    });
+    var validationStates={
+            hasWarning: 'has-warning',
+            hasError: 'has-error',
+            hasSuccess: 'has-success'
+    };
+    var getTypeForSubmittableElement=(function(el){
```
briri: This is good. I like that it is generic. We may want to refactor later to allow for custom error messages, maybe setting a specific attr on the input like 'dmproadmap-error'. This would allow us to provide messaging like 'Passwords must be t least 8 characters and contain ...'

Issue #553, opened 2017-08-03
Refactor CSS/JS on 'My Dashboard' (Core page)


briri:
Refactor the My Dashboard page to use Bootstrap.

This page will serve as an example/reference for how to implement tables on our other pages.
http://getbootstrap.com/css/#tables

Remove or replace the functionality in `lib/assets/stylesheets/dmproadmap/tables.css` and `lib/assets/javascripts/dmproadmap/tables.js`

briri:
Bootstrapped the My Dashboard page.

![screen shot 2017-08-11 at 12 27 48 pm](https://user-images.githubusercontent.com/1204467/29228924-a76cf9cc-7e90-11e7-9654-968a17fead4a.png)

stephaniesimms:
visual inspection on stage. closing issue.

PR #583, opened 2017-08-11
Bootstrapped the My Dashboard page

briri:
- Removed old `tabs.js` and `tables.js` files
- Moved JS logic behind table sorting and filtering to `collateTable.js` and `filteriseTable.js` as part of `dmproadmap.utils`
- Bootstrapped the My Dashboard page and shared table filter code per #553 
- Removed css overrides for basic bootstrap look/feel for tabs and footer. We will do this at the end after all of the pages have been bootstrapped
- add `<div class="row"><div class="col-md-12"> ... </div></div>` around each page's header and initial paragraph 

@jollopre most of the file updates on this one just involve wrapping the `<h1>` in the 'row + col-md-span' structures for Bootstrap


app/views/plans/index.html.erb
```diff
@@ -1,104 +1,105 @@
...
 <div class="row">
   <div class="col-md-12">
     <% if @plans.length > 0 %>
-      <table class="dmp_table tablesorter">
-        <thead>
-          <% if @plans.length > 10 %>
+      <div class="table-responsive">
+        <table class="table table-hover tablesorter">
+          <thead>
+            <% if @plans.length > 10 %>
```
jollopre: We should have 10 as constant. Where has been agreed that 10 is the minimum to display the filter box?

lib/assets/javascripts/utils/collateTable.js
```diff
@@ -0,0 +1,31 @@
...
+  });
+  
+  ctx.init = ctx.init || (function(options){
+    if($ && options && options.selector){
+      /* Bind the table to the external tablesorter JS (see vendor/tablesorter) */
+      $(options.selector).tablesorter(); 
+    
+      /* Tablesorter doesn't provide icons so we add our own here */
+      /* we also bind their click event to display the up/down arrow */
+      $(options.selector + " th.tablesorter-headerUnSorted:not(.sorter-false) div.tablesorter-header-inner")
```
jollopre: I would simplify this selector, so that it does not depend on tablesorter to insert the span. (e.g. $(options.selector + 'th') and at the end invoke $(options.selector).tablesorter(); to augmentate the HTML according to that library. Note we would need to excludethe actions columns but there is already a class 'sorter-false'.

lib/assets/javascripts/utils/filteriseTable.js
```diff
@@ -0,0 +1,47 @@
...
+  filteriseTable adds filter capabilities to an HTML table 
+  table rows are shown/hidden as the user enters text into the filter input field 
+  all rows are made visible when the user clicks the 'clear' icon
+*/
+(function(ctx){
+
+  var filter = (function(el){
+    var query = $(el).val(),
+        regex = new RegExp(query, 'i'),
+        matched = false;
```
jollopre: matched is never used, please remove it.

lib/assets/javascripts/utils/filteriseTable.js
```diff
@@ -0,0 +1,47 @@
...
+
+  var clear = (function(el){
+    $(el).val('');
+    $(el).closest("table").find("tbody tr").show();
+  });
+
+  ctx.init = ctx.init || (function(options){
+    if($ && options && options.selector){
+      /* Bind the keyup  function to the input field's keyup event */
+      $(options.selector).keyup(function(e){
```
jollopre: debounced keyup will improve performance since execution of regex could be expensive operation.

app/views/plans/index.html.erb
briri: It was discussed in a quick conversation. I agree that it should be a constant to make it easily changeable in one spot, will move to `config/initializers/constants`.

lib/assets/javascripts/utils/collateTable.js
briri: yes, tablesorter has some support for its own sort icons. I will update to use its built-in functionality

lib/assets/javascripts/utils/filteriseTable.js
briri: good catch

briri:
updated to use debounce instead of firing table filter method on keyup. Also updated the tablesorter to use the builtin icon functions.

lib/assets/javascripts/utils/filteriseTable.js
```diff
@@ -0,0 +1,73 @@
+/*
```
jollopre: There is no need of a map of debounced functions, regex.match is computationally more expensive than test, there is not need for checking query.length. See my proposal below

/* 
  filteriseTable adds filter capabilities to an HTML table 
  table rows are shown/hidden as the user enters text into the filter input field 
  all rows are made visible when the user clicks the 'clear' icon
*/
(function(ctx){

  var filter = (function(el){
    var query = $(el).val();
    var regex = new RegExp(query, 'i');
    $.each($(el).closest("table").find("tbody tr"), function(idx, tr){
        if(regex.test($(tr).text())){
          $(tr).show();
        }else{
          $(tr).hide();
        }
      });
  });

  var clear = (function(el){
    $(el).val('');
    $(el).closest("table").find("tbody tr").show();
  });

  ctx.init = ctx.init || (function(options){
    if($ && options && options.selector){
      var debounced = dmproadmap.utils.debounce(filter);
      $(options.selector).keyup(function(e){
        debounced(options.selector);
      });

      $(options.selector).siblings("#clear_filter").click(function(e){
        e.preventDefault();
        clear(options.selector);
        debounced.cancel();
      });
    }
  });

})(define('dmproadmap.utils.filteriseTable'));

app/views/shared/_table_filter.html.erb
```diff
@@ -1,7 +1,10 @@
-<div class="toolbar">
+<div id="tablefilter-<%= SecureRandom.hex %>" class="tablefilter form-group pull-right">
```
jollopre: Remove the tag id, I guess you introduced it for the map of debounced functions, which is not needed.

lib/assets/javascripts/utils/filteriseTable.js
jollopre: Happy to explain my changes if you want

lib/assets/javascripts/utils/filteriseTable.js
```diff
@@ -0,0 +1,46 @@
...
+  ctx.init = ctx.init || (function(options){
+    if($ && options && options.selector){
+      var id = $(this).attr("id");
+
+      /* initialize a debounced listener for the filter box */
+      var debounced = dmproadmap.utils.debounce(filter);
+
+      /* Bind the clear function to the clear icon's click event */
+      $(options.selector).keyup(function(){
+        debounced(this, 250);
```
jollopre: the call to debounced function internally calls to the function that you have passed in L31, i.e. the filter function and passes the arguments this (e.g. the current HTML element) together with the integer 250. The second argument is never used in filter function, which only has el as argument, so for is misleading here. I guess you wanted to decrease the time in which the filter function would be triggered. I resolved some conflicts from this PR together with cleaning the call to debounced.

PR #584, opened 2017-08-15
fixes bug with duplicate guidance and adds rake task to cleanup database

xsrust:
### original issue:
code which applies customizer's customizations to an updated base template copying over all guidance, not just that of the customizer, this caused duplicate funder guidance.

### bugfix: 
only copy over the current user's org's annotations

### data correction:
rake task added to remove all these duplicate annotations from the codebase
rake migrate:remove_duplicate_annotations

Also looks like the editor i've switched to, atom.io, has ripped trailing whitespace out of saved files

### DIFF
--- a/app/controllers/templates_controller.rb
+++ b/app/controllers/templates_controller.rb
@@ -11,7 +11,7 @@ class TemplatesController < ApplicationController
   def admin_index
     authorize Template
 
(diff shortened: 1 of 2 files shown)


xsrust:
hadnt saved most current version of lib/tasks/migrate.rake file

Issue #566, opened 2017-08-03
Refactor CSS/JS for Modal dialogs


briri:
Replace the Jquery-UI CSS and JS on the `lib/assets/javascripts/dmproadmap/modal.js` and `lib/assets/stylsheets/modal.scss` with Bootstrap's version of modal.
http://getbootstrap.com/javascript/#modals

Then verify that all of the site's modals are properly implemented (e.g. the sign in/create account forms triggered from the `app/views/layouts/_signin_signout.html.erb`)

stephaniesimms:
visual inspection on stage. closing issue.

PR #585, opened 2017-08-15
DMPRoadmap/roadmap#566

jollopre:
Removed non-relevant modal at phases/edit.html.erb

Doing some housekeeping for non-relevant modals. DMPRoadmap/roadmap#566

modals according to bootstrap. DMPRoadmap/roadmap#554

modal partial as layout to be re-used across the site. DMPRoadmap/roadmap#554


jollopre:
NOTE: the JS file (lib/assets/javascripts/views/guidances/admin_edit.js) has a call to the previous modal library (jquery-ui) that has been left there intentionally since the views that use this function has to ariatised with ariatiseForm and therefore, that popup for showing non-valid fields will no longer apply.

PR #598, opened 2017-08-21
Issue558

briri:
Pushing this before the work on #558 is complete so that @jollopre can pull the JS changes in prior to the upgrade to ES6.

Still need to bootstrap the sign in and create account forms. Will send over as a separate PR.


lib/assets/javascripts/constants.js
```diff
@@ -0,0 +1,15 @@
...
+    PASSWORD_MIN_LENGTH: 8,
+    PASSWORD_MAX_LENGTH: 128,
+
+    VALIDATION_MESSAGE_PASSWORD: "The password must be between 8 and 128 characters.",
+    VALIDATION_MESSAGE_EMAIL: "You must enter a valid email address.",
+    VALIDATION_MESSAGE_REQUIRED: "This field is required.",
+    VALIDATION_MESSAGE_PASSWORDS_MATCH: "The passwords must match."
+  };
+
+  for(var key in vals){
```
jollopre: Not need to iterate over vals object keys. You could just type ctx = vals; and from outside would be accessed similarly, i.e. dmproadmap.constants.PASSWORD_MIN_LENGTH

lib/assets/javascripts/utils/ariatiseForm.js
```diff
@@ -44,48 +44,52 @@
...
-        else
-            return 'unknown';
+    var getValidationTypeForElement=(function(el){
+      var validation = $(el).attr('data-validation');
+      // if the specified validation type is defined
+      if(validation && dmproadmap.utils.validate[validation]){
+        return $(el).attr('data-validation');
+
+      }else if($(el).attr('aria-required') === 'true'){
+        return 'required';
```
jollopre: Are you assuming that absence of data-validation attribute should fallback to type 'required'. This is too strong assumption that I can think cases that would fail such as expecting a number, or an email... Absence of data-validation, should fallback to the default validation defined for the type, like it was before adding data-validation.

lib/assets/javascripts/utils/ariatiseForm.js
```diff
@@ -44,48 +44,52 @@
     3. Prevent form to be submitted
 */
 (function(ctx){
-    var requiredFields=(function(selector){
-        return $(selector).find('.form-control').filter('[aria-required="true"]');
+    // Collect all elements that are either required or have data-validation defined
+    var validatableFields=(function(selector){
+        return $(selector).find('.form-control').filter('[data-validation],[aria-required="true"]');
     });
     var blockHelp=(function (id,type){
```
jollopre: this function has turned into impure since now depends on defaultValidationError to return a value. The logic to determine what message to display should be defined outside... And returning empty string for defaultValidationError is not representative for the user. She/he could thing there is a bug.

Issue #587, opened 2017-08-16
Introducing Webpack to handle assets pipeline


jollopre:
Since JS ecosystem is mature enough nowadays to bundle all front-end assets (e.g. javascripts, stylesheets, images, etc). It is best to replace the sprockets assets pipeline from Rails with task bundler like Webpack. 

There are many advantages by moving towards a JS run bundler:
- Easy integration of new JS features from ES6 (modules, const, let, arrow functions, etc)
- Easy integration of unit and integration JS test
- More active community to solve problems or issues
- Easy splitting of file bundlers, for instance per page if the current JS bundled (application.js) grows considerably
- and many more...

briri:
This is complete. Closing since there is no UAT component

PR #600, opened 2017-08-22
DMPRoadmap/roadmap#587

jollopre:
First time will be needed:

1. cd lib/assets
2. npm install
3. npm run bundle [-- -p] 

and later just:
1. cd/lib/assets
2 npm run bundle [-- -p]

Note the argument -- -p it is just for production.


briri:
It looks like we might need to throttle Travis to provide enough time for the webpack stuff to process

Issue #603, opened 2017-08-23
seeds.rb missing key data values


briri:
The seeds.rb file appears to be missing key data values like `formattype` in the `question_formats` table.

Look through the tables and make sure all of the columns are represented in the seeds.rb then build a rake task to allow installations to correct seed issues.

PR #604, opened 2017-08-23
Fix for Question Format issue #603

briri:
This patch updates the `seeds.rb` file that new installations use to seed their database tables. The file was missing the `question_formats.formattype` field and the `token_permissions_type` table was missing entries for 'templates' and 'statistics'


jollopre:
References to my thoughts:
- answer's plan: https://github.com/DMPRoadmap/roadmap/blob/master/app/views/answers/_new_edit.html.erb#L40
- templates's edit/creation:
https://github.com/DMPRoadmap/roadmap/blob/master/app/views/questions/_preview_question.html.erb#L14

PR #610, opened 2017-08-28
Bootstrapped Devise forms

briri:
- bootstrapped forgotten password, invitation, sign in and create account forms per #558.
- converted ariatiseForm for ES6
- consolidated all old views/devise JS files into a single devise.js 
- created stub ariatiseFormSpec.js 
- created passwordHelper.js to handle the addition of the 'Show password' checkboxes and the password-password_confirmation match validation


.gitignore
```diff
@@ -8,6 +8,7 @@
 # Ignore all logfiles, tempfiles, public assets,
 /log/*.log
 /tmp
+/vendor
```
jollopre: vendor folder is no longer used to hold third party libraries for front, therefore there is no need to add an ignore line.

lib/assets/javascripts/constants.js
```diff
@@ -1,2 +1,13 @@
 export const PASSWORD_MIN_LENGTH = 8;
 export const PASSWORD_MAX_LENGTH = 128;
+
```
jollopre: I would create another file for just messages, although this will go away when the translations library is in place...

lib/assets/javascripts/dmproadmap/forms.js
```diff
@@ -59,7 +59,7 @@ function validatePassword(sPassword) {
   if(sPassword.trim().length >= 8 && sPassword.trim().length <= 128){
     return '';
   }else{
-    return __('Passwords must have at least 8 characters');
+    return 'Passwords must have at least 8 characters';
```
jollopre: I guess is hardcoded since this file is conforming to ES5 still.

lib/assets/javascripts/views/devise/devise.js
```diff
@@ -0,0 +1,40 @@
+import ariatiseForm from '../../utils/ariatiseForm';
+import { addMatchingPasswordValidator, addShowHidePasswordControl } from '../../utils/passwordHelper';
+import { SHOW_SELECT_ORG_MESSAGE, SHOW_OTHER_ORG_MESSAGE } as constants from '../constants';
```
jollopre: the syntax for this import is not fine. Remove as constants and the add the URL as ''../../constants please

lib/assets/javascripts/views/devise/devise.js
```diff
@@ -0,0 +1,40 @@
...
+
+/*
+  Ariatise all of the devise forms and attach a
+  @param value to check
+  @return true or false
+*/
+if (document) {
+  document.addEventListener('DOMContentLoaded', () => {
```
jollopre: this is vanilla JS, we are using jQuery library for adding listeners, we have to be consistent, either vanilla JS or Jquery but not both, please. I would be more inclined to use vanilla JS since browsers are more consistent nowadays, we can discuss this in our weekly dev's meeting.

app/views/shared/_sign_in_form.html.erb
```diff
@@ -0,0 +1,35 @@
...
+      </span>
+    </div>
+
+    <h3 class="center-text">- <%= _('or') %> -</h3>
+  <% end %>
+  
+  <div class="form-group">
+    <%= f.label(:email, _('Email'), class: 'control-label') %>
+    <%= f.email_field(:email, class: 'form-control', "aria-required": true) %>
+    <%= f.check_box :remember_me, class: 'form-control' %>
```
jollopre: This checkbox does not appear inlined with the label. I think bootstrap got a default class for aligning input checkbox together with label, see the example here (https://getbootstrap.com/docs/3.3/css/#forms-example)

lib/assets/javascripts/utils/passwordHelper.js
```diff
@@ -0,0 +1,71 @@
...
+          } else {
+            valid(cnf, `#${id}`);
+          }
+        });
+      }
+    }
+  }
+};
+
+export const addShowHidePasswordControl = (options) => {
```
jollopre: This is a nice utility, although the HTML itself for adding the label and show password can be added directly in the view and I personally think it will improve readability of that HTML view (e.g. app/views/shared/_sign_in_form.html.erb) .

lib/assets/javascripts/dmproadmap/forms.js
briri: Yes, it can't find `__` anymore and I knew we were moving to a different solution for translations so figured I would just hard code the value for now

lib/assets/javascripts/views/devise/devise.js
briri: ok. I was having trouble with this. It seems as though jQuery isn't ready at the point this code was executing.

lib/assets/javascripts/utils/passwordHelper.js
briri: ok. I can move the actual HTML part back onto the form and just keep the JS in this file

lib/assets/javascripts/constants.js
briri: Yeah, lets wait to move these until Jed is in place. If we at least consolidate them all here it will be easier to move I think later on

.gitignore
briri: ok. will remove this and just delete the old directory from my machine

briri:
switched passwordHelper.js to only handle the text to password switch for the show password checkboxes. remove vendor/ from .gitignore

briri:
Can't seem to fix the CRLF error coming up in the linter for application.js. not sure what the issue is. None of the other files are affected.
I also updated the views/contacts/new.js to use the new ES6 based ariatiseForm.js

jollopre:
I have fixed the linter problems and improved some JS lines. See commit https://github.com/DMPRoadmap/roadmap/commit/199f335ba6d94250f57cd2bb9ae7477ba8292435

Happy to explain why...

PR #613, opened 2017-08-31
Issue559

jollopre:
Guidance toggle has been commented out. Do you want me to address the re-factoring in this issue?


app/views/plans/_edit_details.html.erb
```diff
@@ -1,132 +1,208 @@
...
-	      <ul id="priority-guidance-orgs">
-	        <%= render partial: "guidance_choices", 
-	                   locals: {choices: @important_ggs, form: f,
-	                            current_selections: @selected_guidance_groups} %>
-	      </ul>
+<div class="row">
+    <div class="col-md-8">
+    <%= form_for plan, html: {method: :put, class: 'form-horizontal edit_plan' } do |f| %>
+      <div class="form-group"><!-- title -->
+        <div class="col-md-12">
```
briri: You shouldn't nest columns like this. Columns should appear as children of `.row`. You should structure it like this:
```html
<div class="row">
  <div class="col-md-12>
     [title content]
  </div>
</div>
<div class="row">
  <div class="col-md-8">
    <div class="row">
      <div class="col-md-12">
        [form content]
      </div>
    </div>
  </div>
  <div class="col-md-4">
    [guidance section]
  </div>
</div>
```
or this if you want to use the grid to align the tooltips you can nest rows within a column (https://getbootstrap.com/docs/3.3/css/#grid-nesting):
```html
<div class="row">
  <div class="col-md-8>

    <div class="row">
      <div class="col-md-12">
        [title content]
      </div>
    </div>

    <div class="row">
      <div class="col-md-8">
        [1st form input]
      </div>
      <div class="col-md-4">
        [1st tooltip/popover]
      </div>
    </div>

    <div class="row">
      <div class="col-md-12">
        [2nd form input - has no tooltip/popover]
      </div>
     </div>

     ... repeat for each form input ... 
  </div>

  <div class="col-md-4">
    [guidance section]
  </div>
</div>
```

jollopre: I'm using horizontal form, i.e. form-horizontal from bootstrap. This means that .form-group class behaves as a grid row, so there is no need for adding .row. See documentation: https://getbootstrap.com/docs/3.3/css/#forms-horizontal

Note, that I respect row and column grid structure to separate the left container (plan details) from the right (guidance choices).

briri: ugh. One of Bootstrap's shortcut anti-patterns. `.form-group` == `.row` when inside a `.form-horizontal` and you use the same col classes.

Issue #560, opened 2017-08-03
Refactor CSS/JS for Share plan page


briri:
Refer to the work done in #555 for tabs and #554 for forms and #553 for tables

stephaniesimms:
visual inspection on stage. closing issue.

PR #615, opened 2017-08-31
JS for share plan page and moved JS for views back into separate files

briri:
Finished up JS for #560 and moved JS for views back into individual files
- Moved the consolidated JS for views back into separate files.
- Removed old plan share js from es5
- Added aria-required to share plan forms
- Removed redundant data-validation from passwords
- Added radio buttons to the ariatiseForm.js and validations


lib/assets/javascripts/views/plans/share.js
```diff
@@ -1,8 +1,13 @@
+import ariatiseForm from '../../utils/ariatiseForm';
+
+if ($) {
```
jollopre: Please, remove the if since webpack ensures that jquery is imported and consider calling passing to $ an anonymous arrow function, i.e. no need for ready.

jollopre:
The linter is complaining at share.js file since you have not added new line at the end. I suggest you to run npm bundle before sending a PR in order to see if webpack produces a valid JS bundle according to our linter definition.

briri:
closing because this is combined in new PR with ariatiseForm.js tests

Issue #550, opened 2017-08-01
Sharing plan w/ capitalised email of an existing user fails


vyruss:
When you share a plan with an existing user but using different capitalisation of their email, you get an error saying `user.role can't be blank`.

sjDCC:
Just adding a screengrab for context and clarification:

![capture](https://user-images.githubusercontent.com/6169940/28868771-5a02418a-7772-11e7-9e1b-042f1947eec3.PNG)

I couldn't share a plan to 'C.Lightfoot@uel.ac.uk' as she was already known in the system as 'c.lightfoot@uel.ac.uk'

briri:
this is complete. will be merged into master during next release

sjDCC:
Works in DMPonline now too so happy to remove from board. Thanks!

briri:
thanks for confirming @sjDCC

PR #618, opened 2017-09-01
updated plan sharing to be case-insensitive

xsrust:
Addresses #550 

Adds a where_case_insensitive method to User which takes as input the field to search for, and the value to match.

Updated the roles_controller to use this new method and respond with the correct case of the user's email in the message

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -6,8 +6,8 @@ class User < ActiveRecord::Base
   #   Include default devise modules. Others available are:
   #   :token_authenticatable, :confirmable,
   #   :lockable, :timeoutable and :omniauthable
-  devise :invitable, :database_authenticatable, :registerable, :recoverable, 
-         :rememberable, :trackable, :validatable, :omniauthable, 
+  devise :invitable, :database_authenticatable, :registerable, :recoverable,
+         :rememberable, :trackable, :validatable, :omniauthable,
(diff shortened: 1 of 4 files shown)


xsrust:
Didnt realise this was still sitting in the PR's

This Now also includes an update to the default pdf settings. @sjDCC 
All margins are set to 10, font-size is set to 10, and the font is changed to times new roman.
PDF demo below.
[sam's plan.pdf](https://github.com/DMPRoadmap/roadmap/files/1281063/sam.s.plan.pdf)

app/models/user.rb
```diff
@@ -269,7 +269,14 @@ def self.from_omniauth(auth)
   def deliver_invitation(options = {})
     super(options.merge(subject: _('A Data Management Plan in %{application_name} has been shared with you') % {application_name: Rails.configuration.branding[:application][:name]}))
   end
-
+  ##
+  # Case insensitive search over User model
+  # @param field [string] The name of the field being queried
+  # @param val [string] The string to search for, case insensitive
+  # @return [ActiveRecord::Relation] The result of the search
+  def self.where_case_insensitive(field, val)
```
briri: In Rails 5 we can move this method to the new ApplicationRecord.rb to make it available to all of the models

briri:
Looks good @xsrust

PR #619, opened 2017-09-04
Issue612

jollopre:
Orgs with organisation org_type can now be customised.

### DIFF
--- a/app/controllers/templates_controller.rb
+++ b/app/controllers/templates_controller.rb
@@ -11,11 +11,13 @@ class TemplatesController < ApplicationController
   def admin_index
     authorize Template
 
-    funder_templates, org_templates = [], []
+    funder_templates, org_templates, organisation_templates = [], [], []
 
     # Get all of the unique template family ids (dmptemplate_id) for each funder and the current org
     funder_ids = Org.funders.includes(:templates).collect{|f| f.templates.where(published: true).valid.collect{|ft| ft.dmptemplate_id } }.flatten.uniq
     org_ids = current_user.org.templates.where(customization_of: nil).valid.collect{|t| t.dmptemplate_id }.flatten.uniq
+    # Get all of the unique template family ids (dmptemplate_id) for each organisation type
+    organisation_ids = Org.organisation.includes(:templates).where( templates: { published: true, migrated: false }).collect{ |org| org.templates.collect{| ot| ot.dmptemplate_id } }.flatten.uniq
 
     org_ids.each do |id|
       current = Template.current(id)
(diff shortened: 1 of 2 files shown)


xsrust:
Behavior looks sensible to me.  Perhaps check with @stephaniesimms and @sjDCC  on the section title being 'Default templates'

briri:
Do we want to allow any template from a Org defined as an organization to be customized? The ticket references the default template specifically. If @sjDCC and @stephaniesimms are ok with allowing every one of those templates to be customized I would caution against using the 'Default templates' terminology. The system only allows for one default template and it is important during the create plan stage. Reusing the term in this manor could be confusing.

sjDCC:
My preference for now is just to let the default template in each tool be customised so that title works for me. It is not appropriate though if more than the default template displays to customise.

@jollopre what does the current functionality do - only select default templates or any template from an organisation? (i.e. not funder or institution)

We've been discouraging third parties such as DataOne and EUDAT from creating templates as it's not obvious when they'd be used and by whom. If some exist and we also display them in a list to customise then we shouldn't use the heading 'default templates'. I don't think that approach is desirable though as it'll open up questions about when to use these templates etc. I'm not sure any templates for orgs exist (except the default ones), but if so they're templates that our current system doesn't expose / make it feasible to select via the create plan wizard afaik....

@stephaniesimms what do you think?

jollopre:
Currently, the code I implemented for default template(s) table view is displaying any template from any orgs type 'organisation' that is published and not migrated. Recalling from Jimmy's email, to display just one default template, i.e. the one from the primary organisation (e.g. DCC, Tuuli, etc) we will need either:

1. Creating an statement in code that only display the default template for a primary organisation or
2. Add a new org type (e.g. customisable generic org or primary org) or
3. Add customisable flag that belongs to org type 'organisation'

I would be more inclined to take the third option. My reasons are:

- No need for additional statements in code (e.g. if ...)
- More flexibility to permit other orgs marked as organisation to be displayed as part of the default templates
- No need for an additional org type that is just for one tuple in the db

@briri, @vyruss, @xsrust, @raycarrick-ed what are your thoughts?

xsrust:
I agree with @jollopre on the customizable flag leading to clean code/flexibility.
It also makes it easy to do the pundit validation.

Speaking of the pundit validation, it seems we arent validating that the org of the template passed to admin_customize is a funder (or rather, after this addition, is customizable), either way, we should add this.

sjDCC:
I think probably super-admins - what do you think @stephaniesimms 

Would be good to add the validation you mention too @xsrust

stephaniesimms:
I'm not sure I understand this issue 100% so please clarify my questions/comments below if you see the need @sjDCC et al - 
- Can there ever be more than 1 default template? my understanding was that the DCC template is *the* one and only default template unless a given service, e.g., Tuuli or DMPAssistant chooses to implement their own.
- If this is the case, I don't see the need to create a separate table for the 'Default template' in the templates area. We might consider just including it in the Funder templates table and providing the same permissions to customize it.
- Super admins would have the authority to make the funder and default templates customizable

briri:
So is the decision to only allow funder templates and the default template to be customized. 

If so, the simplest solution would be to just add the one default template (designated by the `templates.is_default` field) to the table currently containing the funder templates. If we go that route we should probably make it the first one in the table and change the title of the section to "Customizable templates" instead of "Funder templates".

sjDCC:
Hi @stephaniesimms I think there can only ever be one default template and that's the one we want to allow org-admins to customise. As such I'm happy for it to be included in the same table as funder but to rename that as 'customisable templates' as @briri suggests.

@jollopre's question had been whether we wanted templates from any organisation to be customised e.g. DCC, DataOne, EUDAT, SSI... I'm not sure there are many templates from other org types (i.e. not funders or unis) but I suspect we only want the default one being customised. The initial request was specifically for the default one.

briri:
So we want to only allow the default template + funder templates in the list. Also change the heading above the table to read "Customisable templates". Make sure that the default template is in the first row of the table.

Issue #567, opened 2017-08-03
Refactor CSS/JS for accordions


briri:
Replace the Jquery-UI CSS and JS on the `lib/assets/javascripts/dmproadmap/accordion.js` and `lib/assets/stylsheets/accordion.scss` with Bootstrap's version of collapse.
http://getbootstrap.com/javascript/#collapse

Then verify that all of the site's accordions are properly implemented (e.g. the plan write page)

stephaniesimms:
visual inspection on stage. closing issue.

PR #629, opened 2017-09-07
Bootstrapped Accordions on Write Plan page

briri:
Bootstrapped the accordions on Write Plan page #567 

Note that the expand/collapse all links on the section accordion expand/collapse all of the guidances contained within the sections as well (the guidance is a nested accordion within the section accordion panel). Don't see a way around this without hacking Bootstrap a bit.

Will include tests for expandCollapseAll.js on Monday (forgot to do a `git add` from my work machine yesterday)


lib/assets/javascripts/utils/expandCollapseAll.js
```diff
@@ -0,0 +1,47 @@
+import 'bootstrap-sass/assets/javascripts/bootstrap.min';
```
jollopre: only import the bootstrap plugin that you are going to use (e.g. ... javascripts/bootstrap/collapse.js

lib/assets/javascripts/utils/expandCollapseAll.js
```diff
@@ -0,0 +1,47 @@
...
+ *      <div id="collapseA" class="panel-collapse collapse" role="tabpanel" 
+ *                  aria-labelledby="headingA">
+ *        <div class="panel-body">
+ *          This is test section A.
+ *        </div>
+ *      </div>
+ *    </div>
+ *  </div>
+ */
+export default () => {
```
jollopre: Nice collapsible abstraction!

lib/assets/javascripts/views/phases/edit.js
```diff
@@ -1,7 +1,14 @@
+import expandCollapseAll from '../../utils/expandCollapseAll';
+
+$(() => {
+  // Attach handlers for the expand/collapse all accordions
+  expandCollapseAll();
```
jollopre: As the module expandCollapseAll always walks through any accordion-controls class, i.e. the selector is hardcoded in the module, perhaps is better if we rename this filename to something more generic since it will be used in multiple views, not only at phases/edit.

PR #631, opened 2017-09-08
Bootstrapped Plan Download 

briri:
- Renamed show_export.html.erb to download.html.erb more closely match page name (updated corresponding controller, policy, link in tab and route)
- Renamed export_settings.js to download.js and converted to ES6. Dropped all of the old auto-save JS. Can reimplement if needed with shared ES6 JS once that's been built out
- Bootstrapped the form #561


briri:
@jollopre it looks like the merge conflict is in the application.js file. You should be able to just resolve it directly in Github. I wonder if it was caused by the new rebasing workflow

briri:
Resolved merge conflict on write plan page (pulled in @jollopre changes to the edit plan page).

briri:
Fixed tests and linter issues

PR #635, opened 2017-09-12
Issue634

jollopre:
Optimisation made at two levels:
- Retrieving only sections, questions... related to the phase selected
- Iterating over the questions related to that phase instead of all the plan questions to create the hash of question_guidance.


briri:
looks much more efficient

Issue #632, opened 2017-09-08
Refactor JS/CSS for Create Plan page


briri:
Forgot this one. Need to bootstrap the create plan page and convert its JS over to ES6.

stephaniesimms:
visual inspection on stage. closing issue.

PR #636, opened 2017-09-12
Boostrapped the Create Plan page

briri:
- Converted JS for page over to ES6
- Added stub test and fixture
- Converted jquery autocomplete combobox to Bootstrap and ES6
    
TODO:
- Add in AJAX calls to retrieve template options based on org/funder selections
- Fix issue with the hidden id field being updated when combobox item is selected
- Finish tests for the autoComplete.js


briri:
@jollopre I'm merging what I have for the demo tomorrow. Ticket #632 still has some work to be done.

Issue #562, opened 2017-08-03
Refactor CSS/JS for Write Plan page


briri:
Refer to the work done in #555 for tabs and #554 for forms

stephaniesimms:
visual inspection on stage. closing issue.

PR #651, opened 2017-09-21
DMPRoadmap/roadmap#562

jollopre:
- guidances_notes layout according to bootstrap. new notes view according to bootstrap. Display logic for each view at JS. DMPRoadmap/roadmap#562
- new notes working without js.erb templates. Fixed functional test broken. TODO update number of comments view. DMPRoadmap/roadmap#562
- $.ajax attributes url and method extracted from closest form action and method attributes respectively. DMPRoadmap/roadmap#562
- number of comments template and includes method at answer for abstracting logic. DMPRoadmap/roadmap#562


briri:
Looks good @jollopre. I like the use of the html form tag to extract the ajax target and method. We should see if its possible to move the ajax off into a reusable JS function. The Create plan, Share plan and My dashboard pages are making ajax calls.

Issue #647, opened 2017-09-18
Bootstrap refactor Org Details page


briri:
ES6 and Bootstrap refactoring of the Org Admin, Org Details page.

Include #639 in this work

stephaniesimms:
visual inspection on stage. closing issue.

PR #663, opened 2017-10-04
Org admin details page(#647)

bhavi:


app/views/orgs/admin_edit.html.erb
```diff
@@ -6,71 +6,80 @@
...
+      </div>
     <% end %>
+
+    <div class="form-group" id="banner-text">
+      <%= f.label _('Top banner text'), for: @org.banner_text, class: "col-sm-2 control-label" %>
+      <div class="col-sm-8">
+        <%= text_area_tag('org_banner_text', @org.banner_text, id: "org_banner_text", class: "form-control") %>
+        <script>
+          /* Initialises an editor for textarea defined above with id org_banner_text */
+          dmproadmap.utils.tinymce.init({ selector: '#org_banner_text' });
```
briri: We don't want JS embedded in the HTML anymore. Move this to the page's corresponding JS file. See @jollopre example of wiring up a tinymce editor on the views/answers/ JS files

lib/assets/javascripts/views/orgs/admin_edit.js
```diff
@@ -1,20 +1,25 @@
...
-          alert(__('Please only enter up to 165 characters, you have used') + " " + getStats('org_banner_text').chars + ". " + __('If you are entering an URL try to use something like http://tinyurl.com/ to make it smaller.'));
-          return false;
-      }
+// Import TinyMCE
+import tinymce from 'tinymce/tinymce';
+import ariatiseForm from '../../utils/ariatiseForm';
+import { SHOW_ORG_BANNER_MESSAGE } from '../../constants';
+
+// Returns text statistics for the specified editor by id
+function getStats(id) {
```
briri: This is valid, but we've been trying to follow this pattern for function declarations:
`const getStats = (id) => {`
The `function getStats(id)` approach hoists the function to the top of the scope while creating it as an anonymous function attached to a variable keeps it within the current scope

PR #670, opened 2017-10-06
Bootstrapping of popovers and tooltips

briri:
Converted all popovers to tooltips.


briri:
Tests are failing. Will fix and resubmit

PR #672, opened 2017-10-09
if no defaults specified, obtuse error. now improved.

raycarrick-ed:
While setting up my devenv, I hit an error which was due to not having an default preferences set. But it was down in user.rb and obscure. So I've added a catch for it. I am just really using this to test that I have set up right and to go through a small change cycle before tackling anything else.


briri:
Looks good @raycarrick-ed. Welcome back!

Issue #667, opened 2017-10-05
Remove any unused ERB views


briri:
There are quite a few erb files out there that are no longer relevant. Go through and remove them.
(Use grep to search for references to them in app/controllers and app/views)

stephaniesimms:
can't test in UI. closing issue.

PR #673, opened 2017-10-09
Removed unused views phase 1

briri:
removed unused views. Renamed those in annotations/templates until those pages have been Bootstrapped DMPRoadmap/roadmap#667

moved plan navigation partial from shared to plans dir

removed old es5 version of org details JS. was causing JS compilation issues

fixed typos in location of plan nav partial

removed unused views and renamed a few in annotations and templates until those pages have been bootstrapped and we can confirm they are not needed.


briri:
tests failing. closing and will resubmit once fixed

PR #674, opened 2017-10-09
Removed unused views

briri:
Removed unused views and renamed ones in annotations/templates until those pages have been Bootstrapped DMPRoadmap/roadmap#667

moved plan navigation partial from shared to plans dir

removed old es5 version of org details JS. was causing JS compilation issues

fixed typos in location of plan nav partial

removed unused views and renamed a few in annotations and templates until those pages have been bootstrapped and we can confirm they are not needed.

switched API docs link to point to wiki docs

added template history view back in. remove token_permission_type controller and test. it was being used to list the API permission types and was linked from the 'How to use API' link. Updated that link to point to Wiki docs

added answer locking partial and phase preview back

added question preview back


briri:
Closing to rebase with latest changes

PR #700, opened 2017-10-12
Updated dragonfly resize to 75px height (retain aspect ratio)

briri:
- decreased dragonfly resize of logos to 75px height
- added rake task `bundle exec rake logos:resize_all` to auto resize all logos to the new 75px
- also removed old ES5 copy statements from webpack config


briri:
closing to fix broken test. will resubmit

PR #717, opened 2017-10-13
Integrated base layout CSS changes for new asset pipeline

vyruss:


vyruss:
rechecking and resubmitting

Issue #716, opened 2017-10-13
Make a copy of a plan throws error


stephaniesimms:
Clicking "Make a copy" of a plan on My Dashboard throws the following error

![screen shot 2017-10-12 at 5 08 04 pm](https://user-images.githubusercontent.com/12703638/31524658-fbe06e2a-af6f-11e7-906a-23b9ed44c0af.png)

stephaniesimms:
working as expected. closing issue.

Issue #713, opened 2017-10-12
Clarify Plan Visibility options on Share page


stephaniesimms:
Revise and add the following to the description below "Set plan visibility" to clarify for users 1) threshold for public or org visibility and 2) why they cannot set visibility for test plans:

"Public and organisational visibility are intended for finished plans. You must answer at least 50% of questions to enable these options. Test plans are private by default and not included in usage information."

![screen shot 2017-10-12 at 4 36 45 pm](https://user-images.githubusercontent.com/12703638/31524009-c906c818-af6b-11e7-9ccf-917847946e64.png)

stephaniesimms:
looks good for now. we can clarify further later on if necessary. closing issue.

Issue #711, opened 2017-10-12
Download plan issues after Bootstrap


stephaniesimms:
- html and pdf should open in a separate tab; current behavior opens them in the same tab
- "PDF Formatting" section should only display if user selects to download pdf (see functional blue notepad specs on UX wireframe http://h90c3m.axshare.com/#g=1&p=write_plan
- for any file format, if I check to download "Project details cover sheet" only, I still get section headings
- capitalize first word only for section headings: "Download settings" "PDF formatting"

stephaniesimms:
all issues working as expected now. closing issue.

PR #721, opened 2017-10-13
Fixed issues with plan duplicattion, plan visibility and plan download

briri:
- updated share visibility message and updated threshold to 50% done instead of 1 question answered #713
- added debounce to new plan to prevent multiple ajax calls for available templates 
- fixes to Plan download and share pages #711
- fixed issue with plan duplicate/copy #716
- fixed issue with plan duplication test

### DIFF
--- a/app/controllers/plans_controller.rb
+++ b/app/controllers/plans_controller.rb
@@ -174,7 +174,12 @@ class PlansController < ApplicationController
     @plan = Plan.find(params[:id])
     authorize @plan
     @visibility = @plan.visibility.present? ? @plan.visibility.to_s : Rails.application.config.default_plan_visibility
-    @allow_visibility = (@plan.num_answered_questions >= 1 && !@plan.is_test?)
+
+    min_percentage = Rails.application.config.default_plan_percentage_answered
+    nanswers = @plan.num_answered_questions()
+    nquestions = @plan.num_questions()
+    value=(nanswers.to_f/nquestions*100).round(2)
+    @allow_visibility = (value >= min_percentage && !@plan.is_test?)
   end
 
 
@@ -290,13 +295,9 @@ class PlansController < ApplicationController
     respond_to do |format|
       if @plan.save
         @plan.assign_creator(current_user)
-        flash[:notice] = success_message(_('plan'), _('copied'))
-        format.js { render js: "window.location='#{plan_url(@plan)}?editing=true'" }
-        # format.html { redirect_to @plan, notice: _('Plan was successfully duplicated.') }
-        # format.json { head :no_content }
+        format.html { redirect_to @plan, notice: success_message(_('plan'), _('copied')) }
       else
-        flash[:alert] = failed_create_error(@plan, 'Plan')
-        format.js {}
+        format.html { redirect_to plans_path, alert: failed_create_error(@plan, 'Plan') }
       end
     end
   end
@@ -454,7 +455,6 @@ class PlansController < ApplicationController
     if templates.empty?
       templates << Template.where(is_default: true, published: true).first
     end
-
     templates = (templates.count > 0 ? templates.sort{|x,y| x.title <=> y.title} : [])
   end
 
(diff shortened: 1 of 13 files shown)


app/controllers/plans_controller.rb
```diff
@@ -174,7 +174,11 @@ def share
     @plan = Plan.find(params[:id])
     authorize @plan
     @visibility = @plan.visibility.present? ? @plan.visibility.to_s : Rails.application.config.default_plan_visibility
-    @allow_visibility = (@plan.num_answered_questions >= 1 && !@plan.is_test?)
+
+    nanswers = @plan.num_answered_questions()
+    nquestions = @plan.num_questions()
+    value=(nanswers.to_f/nquestions*100).round(2)
+    @allow_visibility = (value >= 50.00 && !@plan.is_test?)
```
jollopre: Extracting 50 to a constant will make the code more reusable. The visibility restriction of 50% answered might not apply to any running instance.

app/views/plans/_share_form.html.erb
```diff
@@ -1,5 +1,5 @@
 <h2><%= _('Set plan visibility') %></h2>
-<p><%= _('Public or organisational visibility is intended for finished plans. You must answer at least one question to enable these options.') %></p>
+<p><%= _('Public and organisational visibility are intended for finished plans. You must answer at least 50% of questions to enable these options. Test plans are private by default and not included in usage information.') %></p>
```
jollopre: You can re-use the constant here for the message

lib/assets/javascripts/views/plans/new.js
```diff
@@ -38,7 +39,7 @@ $(() => {
   const getMethod = jQueryForm => jQueryForm.attr('method');
 
   // When one of the autocomplete fields changes, fetch the available templates
-  const handleComboboxChange = () => {
+  const handleComboboxChange = debounce(() => {
     if ($('#plan_org_id').length && $('#plan_funder_id').length) {
       const validOrg = ($('#plan_org_id').val().trim().length > 0 || $('#plan_no_org').prop('checked'));
```
jollopre: please re-use functions, e.g isValidNumber from utils/isValidInputType for both const

jollopre: nice usage of debounce for improving communication with server.

briri:
Changes made

lib/assets/javascripts/views/plans/new.js
```diff
@@ -38,29 +40,27 @@ $(() => {
   const getMethod = jQueryForm => jQueryForm.attr('method');
 
   // When one of the autocomplete fields changes, fetch the available templates
-  const handleComboboxChange = () => {
-    if ($('#plan_org_id').length && $('#plan_funder_id').length) {
-      const validOrg = ($('#plan_org_id').val().trim().length > 0 || $('#plan_no_org').prop('checked'));
-      const validFunder = ($('#plan_funder_id').val().trim().length > 0 || $('#plan_no_funder').prop('checked'));
+  const handleComboboxChange = debounce(() => {
+    const validOrg = (isValidText($('#plan_org_id').val()) || $('#plan_no_org').prop('checked'));
```
jollopre: isValidNumber should equally work since tries to convert the string to a number. Moreover, as the id are integers you would ensure that a valid id is sent back to the server.

config/application.rb
```diff
@@ -90,5 +90,8 @@ class Application < Rails::Application
     #   is_test                   - (NOT advisable because test plans are excluded from statistics)
     #   privately_visible         - Only the owner and people they invite can access the plan
     config.default_plan_visibility = 'privately_visible'
+   
+    # The percentage of answered questions needed to enable the plan visibility section of the Share plan page
+    config.default_plan_percentage_answered = 50.00
```
jollopre: I would put that constant under config/initializers/constant.rb and I would leave application.rb for rails specific.

briri: ah. agreed. I had forgotten about that file. I'll make that adjustment and send along in my next PR

Issue #645, opened 2017-09-18
Bootstrap refactor Guidance show/edit/create pages


briri:
ES6 and Bootstrap refactoring of the Org Admin, Guidance show/edit/create pages

stephaniesimms:
visual inspection of pages. closing issue.

PR #741, opened 2017-10-17
Bootstrapped the Guidances and Guidance Groups pages

briri:
- added select and tinymce to ariatiseForm. Bootstrapped guidances and new guidance pages
- guidance pages #645
- bootstrapped the guidance_groups pages (converted form to labels over fields and ariatised the form)
- added table sorting and filtering to guidances page tables


app/views/guidance_groups/_guidance_group_form.html.erb
```diff
@@ -0,0 +1,14 @@
+<div class="form-group">
```
jollopre: nice extraction of the form for being used at new/edit. We should have more of these

lib/assets/javascripts/utils/ariatiseForm.js
```diff
@@ -77,6 +77,8 @@ const getValidationTypeForElement = (el) => {
       return $(el).attr('type'); // available types at https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#Form_<input>_types
     } else if ($(el).is('select')) {
       return 'select';
+    } else if ($(el).is('.tinymce')) {
```
jollopre: we have data-validation attribute for handling cases that are not wise enough with just type's element. Moreover, using the class tinymce to validate that element assumes that we have to define that class any time a textarea is tinymce augmented. There are textareas in other views that do not define tinymce class to load tinymce and therefore ariatising the form will not work for those textareas.

PR #763, opened 2017-10-20
added helper task to detect active tab

### LINKED ISSUE #715: Reload Notification Preferences page after saving
After updating my notification preferences and clicking "Save" the page reloads and returns me to "Personal Details" tab (screenshot below). Desired behavior is to reload the "Notification preferences" tab.

screen shot 2017-10-12 at 5 00 35 pm

briri:
fixes #715
Added isActiveTab method to application_helper

### DIFF
--- a/app/controllers/registrations_controller.rb
+++ b/app/controllers/registrations_controller.rb
@@ -82,7 +82,6 @@ class RegistrationsController < Devise::RegistrationsController
       @other_organisations = Org.where(parent_id: nil, is_other: true).pluck(:id)
       @identifier_schemes = IdentifierScheme.where(active: true).order(:name)
       @languages = Language.sorted_by_abbreviation
-      @tab = params[:tab]
       if params[:skip_personal_details] == "true"
         do_update_password(current_user, params)
       else
@@ -156,7 +155,7 @@ class RegistrationsController < Devise::RegistrationsController
       set_gettext_locale  #Method defined at controllers/application_controller.rb
       set_flash_message :notice, success_message(_('profile'), _('saved'))
       sign_in current_user, bypass: true  # Sign in the user bypassing validation in case his password changed
-      redirect_to edit_user_registration_path(tab: @tab), notice: success_message(_('profile'), _('saved'))
+      redirect_to "#{edit_user_registration_path}\#personal-details", notice: success_message(_('profile'), _('saved'))
 
     else
       flash[:alert] = message.blank? ? failed_update_error(current_user, _('profile')) : message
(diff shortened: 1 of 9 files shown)


app/helpers/application_helper.rb
```diff
@@ -26,6 +26,15 @@ def isActivePage(path)
     return request.fullpath() == path
   end
 
+  # This is not the most elegant solution. Bootstrap though seems to add a '-tab' suffix
+  # to the end of the query string param, so just strip it out when comparing
+  # ---------------------------------------------------------------------------
+  def isActiveTab(tab)
```
jollopre: The logic to determine what tab is active should be best handled on the JS side. Bootstrap is not adding the tab query param, someone of us has added tab into the url form path for each of the three forms (e.g. https://github.com/DMPRoadmap/roadmap/blob/CDL-MVP/app/views/devise/registrations/_personal_details.html.erb#L1)

FYI, boostrap tabs click handler preventDefault which means it does not add the active tab to the URL. Why don't you read this thread (https://stackoverflow.com/questions/12131273/twitter-bootstrap-tabs-url-doesnt-change), it might help us to implement it on the JS side

briri: Yeah, I was originally going to do a JS solution for this, but the linter creates an error of 'window is not defined'. 
Since this was the only situation (that I noticed) where we had tabs that needed this functionality it was easy to just add it in the erb file. The plan tabs are unique pages so don't have the issue

briri:
ok. found a way to tell ESlinter  that window/document are ok right in the JS file. Converted all of it over to tabHelper.js

app/controllers/registrations_controller.rb
```diff
@@ -156,7 +155,7 @@ def do_update(require_password = true, confirm = false)
       set_gettext_locale  #Method defined at controllers/application_controller.rb
       set_flash_message :notice, success_message(_('profile'), _('saved'))
       sign_in current_user, bypass: true  # Sign in the user bypassing validation in case his password changed
-      redirect_to edit_user_registration_path(tab: @tab), notice: success_message(_('profile'), _('saved'))
+      redirect_to edit_user_registration_path, notice: success_message(_('profile'), _('saved'))
```
jollopre: Could you add the hash #personal-details here too?. I'm aware that is the default active tab when the view get rendered but in case we change the default active in future, we will not need to look for this line for when changes at personal details are made.

lib/assets/javascripts/utils/tabHelper.js
```diff
@@ -0,0 +1,14 @@
+/* eslint-env browser */ // This allows us to reference 'window' below
```
jollopre: Perfect, if we start using window across multiple files we can transfer that to lib/assets/.eslintrc.json under env key like jasmine and jquery.

briri:
added the hash for the default personal-details tab

briri:
fixed test broken by adding hash to redirect URL

Issue #740, opened 2017-10-17
Standardize functionality to Select an Org


stephaniesimms:
The same functionality to select an org should be used on Create Account and Edit Profile pages

- Initial step should prompt user to either Begin typing to see a filtered list (grayed out text in box) OR present a link to "Select an organisation from the list" 
- Clicking "Select an org..." should open the full list - currently clicking the link does nothing
- "My organisation isn't listed" should appear at the top of the list
- If user selects "My org isn't listed" the field should auto-populate with "Other" 

On Create Account page:

![screen shot 2017-10-11 at 9 47 13 am](https://user-images.githubusercontent.com/12703638/31454152-5046895c-ae69-11e7-8a92-c5d2ae721203.png)

On Edit Profile page:

![screen shot 2017-10-11 at 9 50 12 am](https://user-images.githubusercontent.com/12703638/31454253-b270ce8a-ae69-11e7-94c7-88ac8718b503.png)

briri:
**Dev notes:**
There seems to be something wrong with the linkage between the autocomplete dropdown box and the 'other org' textbox on the create plan page. 

Use the setup on the edit profile page as an example. The link should be 'My organisation isn't listed' when the autocomplete box is available. When the user clicks this, the autocomplete should be hidden and the other org textbox should become visible and the link should change to read 'Select an organisation from the list'

stephaniesimms:
@briri When I click "My org is not in the list" I get a different link that says "Select an org from the list" - and when I click this link it switches back to "My org is not in the list" - I never get a box prepopulated with "Other"

This behavior occurs on both the Create account and Edit profile pages. At least it's consistent.

briri:
Hi @stephaniesimms, perhaps I was unclear on how this org/other org should work, or maybe its just confusing or unclear with the current implementation. The autocomplete box and textbox do look very similar.
Here is how I set it up:
If the user opens the page and they already have an 'Organization' defined they will see:
![screen shot 2017-10-26 at 8 25 45 am](https://user-images.githubusercontent.com/1204467/32061877-bcad0200-ba27-11e7-8b9b-897db8c4c13c.png)
If the user opens the page and they already have an 'Other Organization' they will see:
![screen shot 2017-10-26 at 8 26 17 am](https://user-images.githubusercontent.com/1204467/32061934-e8a69466-ba27-11e7-9ef5-1a9842a72255.png)

[...]

PR #767, opened 2017-10-23
Fixes for Org / Other Org 

briri:
See issue #740
- Org autocomplete and other org textbox now display correctly
- added js and css for autocomplete box
- added logic to disable org section of edit profile while account is linked. User must unlink now before changing org
- added warning for org admin about losing privileges if they change org
- added logic to hide the link institutional credentials portion of profile page if user is using other org
- Updated text of Shibboleth section to 'Institutional credentials' per UX wireframe


app/views/shared/_my_org.html.erb
```diff
@@ -0,0 +1,20 @@
...
+
+<%= f.label :org_name, _('Organisation'), class: 'control-label' %>
+<%= render partial: "shared/accessible_combobox",
+             locals: {name: "user[org_name]",
+                      id: "user_org_name",
+                      default_selection: default_org,
+                      models: orgs,
+                      attribute: 'name'} %>
+
+<% if others %>
```
xsrust: We can join this if statement with the one below as they both check the same condition

app/views/devise/registrations/_personal_details.html.erb
```diff
@@ -41,21 +44,23 @@
...
-      <% else %>
-          <%= label_tag(:scheme_name, scheme.name.capitalize, class: 'control-label') %>
-      <% end %>
-      <div class='identifier-scheme'>
-        <%= render partial: "external_identifier_#{scheme.name}",
-                   locals: { scheme: scheme,
-                             id: current_user.identifier_for(scheme)} %>
+    <% if scheme.name == 'shibboleth' && current_user.org.present? %>
+      <div class="form-group col-xs-8">
+        <% if scheme.name == 'shibboleth' %>
```
xsrust: This check is not needed because of the above if statement

app/views/devise/registrations/_personal_details.html.erb
```diff
@@ -41,21 +44,23 @@
...
+            <label class='control-label'>
+              <span class="aria-only" aria-hidden="false"><%= _('Institutional credentials') %></span>
+              <i class="fa fa-user" title="<%= _('Institutional credentials') %>" aria-hidden="true"></i>
+            </label>
+        <% else %>
```
xsrust: The above check to see if the `scheme.name` is shibboleth makes this unreachable.  
Should this be the else to the outer if on line 47?

app/views/shared/_my_org.html.erb
```diff
@@ -0,0 +1,20 @@
+<% others = allow_other_orgs ||= false %>
```
xsrust: nil evaluates to false in ruby so the `||=` is redundant

app/views/devise/registrations/_personal_details.html.erb
briri: yes. great catch.

briri:
revised the logic surrounding the identifier scheme links. We do not want the shib link to show if the person is not associated with an org.

app/views/devise/registrations/_personal_details.html.erb
```diff
@@ -43,18 +46,23 @@
...
+            <span class="aria-only" aria-hidden="false"><%= _('Institutional credentials') %></span>
+            <i class="fa fa-user" title="<%= _('Institutional credentials') %>" aria-hidden="true"></i>
           </label>
+        <% end %>
       <% else %>
-          <%= label_tag(:scheme_name, scheme.name.capitalize, class: 'control-label') %>
+        <%= label_tag(:scheme_name, scheme.name.capitalize, class: 'control-label') %>
+      <% end %>
+
+      <% if scheme.name != 'shibboleth' || (scheme.name == 'shibboleth' && current_user.org.present?) %>
```
jollopre: Why only is required org for an user if shibboleth, if we make it mandatory for any scheme we can avoid this if.

Issue #752, opened 2017-10-19
Add table of Org plans to My Dashboard


stephaniesimms:
UX wireframe @JEK-III http://h90c3m.axshare.com/#g=1&p=my_dashboard

Add a separate table for all the plans marked w/Org or Public visibility for a specific Org if the user has an Org affiliation. This table should appear below the My plans table on My Dashboard page.

Note blue notepad functional specs. UX recommendation to make the Org table a less salient color for production services. Okay to leave gray in Roadmap. 

Add Notifications area to this issue also?

stephaniesimms:
this looks good @jollopre - will create separate issue for formatting long tables w/page breaks. closing this issue.

PR #768, opened 2017-10-24
Issue 752

jollopre:
- Table of Org plans to My Dashboard. DMPRoadmap/roadmap#752
- Pundit policy for organisationally_exportable plans. DMPRoadmap/roadmap#752
- Eager loading of plans, templates and roles for publicly or organisationally visible plans. DMPRoadmap/roadmap#752

### DIFF
--- a/app/models/plan.rb
+++ b/app/models/plan.rb
@@ -55,6 +55,11 @@ class Plan < ActiveRecord::Base
   # Note that in ActiveRecord::Enum the mappings are exposed through a class method with the pluralized attribute name (e.g visibilities rather than visibility)
   scope :publicly_visible, -> { where(:visibility => visibilities[:publicly_visible]).order(:title => :asc) }
 
+  # Retrieves any plan organisationally or publicly visible for a given org id
+  scope :organisationally_or_publicly_visible_by_org, -> (org_id) {
+    joins(:template).where(
+      visibility: [visibilities[:organisationally_visible], visibilities[:publicly_visible]],
+      "templates.org_id": org_id).order(:title => :asc) } 
   ##
   # Settings for the template
   has_settings :export, class_name: 'Settings::Template' do |s|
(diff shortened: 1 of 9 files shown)


lib/assets/javascripts/utils/tableHelper.js
```diff
@@ -42,16 +43,10 @@ export const filteriseTable = (options) => {
       debounced(e.currentTarget);
     });
 
-    $(options.selector).siblings('#clear_filter').click((e) => {
+    $('.clear_filter').click((e) => {
       e.preventDefault();
-      clear(e.currentTarget);
+      clear(e.target);
```
briri: Just curious, but why switch to `target` here but not elsewhere (like debounced above)? I just looked up the Mozilla docs again and see that `currentTarget` returns the EventTarget instead of the DOM element so `target` does seem more appropriate.
https://developer.mozilla.org/en-US/docs/Web/API/Event/Comparison_of_Event_Targets

jollopre: In this case makes no different one or the other since both return the same element, i.e. the 'a' element in which the handler is attached. The use of currentTarget is more associated with event delegators, really useful for listening to events even on future elements added underneath. The keyup handler where the debounced function is invoked similarly could be changed to target and would work. I would be more worried about declaring an anonymous function anytime keyup is triggered but at this point, not worth to change it.

app/models/plan.rb
```diff
@@ -55,6 +55,11 @@ class Plan < ActiveRecord::Base
   # Note that in ActiveRecord::Enum the mappings are exposed through a class method with the pluralized attribute name (e.g visibilities rather than visibility)
   scope :publicly_visible, -> { where(:visibility => visibilities[:publicly_visible]).order(:title => :asc) }
 
+  # Retrieves any plan organisationally or publicly visible for a given org id
+  scope :organisationally_or_publicly_visible_by_org, -> (org_id) {
+    joins(:template).where(
+      :visibility => [visibilities[:organisationally_visible], visibilities[:publicly_visible]],
```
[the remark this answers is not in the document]
jollopre: nice catch, I can't wait to have a linter for ruby that will enforce us to use newer syntax

Issue #742, opened 2017-10-17
Remove "question not answered yet" message


stephaniesimms:
- Remove yellow message bar beneath each question "question not answered yet"
- Only display a blue timestamp message once a user enters text or makes a selection for another question type (radio button) and the system auto-saves it

![screen shot 2017-10-17 at 2 35 47 pm](https://user-images.githubusercontent.com/12703638/31690763-88f9020a-b348-11e7-9c61-56e22b822a4e.png)

stephaniesimms:
the issue w/"question not answered" message is resolved @jollopre 

but now I receive an autosave "question answered" message even if I don't enter anything into the text editor (screenshot below). desired behavior is to only autosave and generate a message if user enters something.

![screen shot 2017-10-22 at 4 33 43 pm](https://user-images.githubusercontent.com/12703638/31862955-234d00d4-b747-11e7-94d7-d49b0e8a6535.png)

stephaniesimms:
tested again and i'm still getting a "question answered" autosave message even if I don't enter anything into the text editor (screenshot below). desired behavior is to only autosave and generate a message if user enters something. 

a new issue is that the message begins counting the seconds since i "answered" the question. i'm not sure what interval it used before (minutes?) but the second counter is distracting and not necessary.

![screen shot 2017-10-24 at 11 21 01 am](https://user-images.githubusercontent.com/12703638/31934963-dcd7dc4c-b8ad-11e7-8cb9-d25ad18edb63.png)

jollopre:
@stephaniesimms I have replaced the seconds counter message by a more static one 'less than a minute ago'. In regards to autosaving an empty question, this could be related to the issue #530 which I'm going to tackle at some point.

Could you create a new plan in order to see if that behaviour still remains for you? I could not replicate it.

stephaniesimms:
thanks for replacing the seconds counter @jollopre 

i created new plans in different browsers (Chrome, Firefox) and w/different accounts. all of them exhibit the same issue. when i click inside the text editor and then move focus away it triggers autosave even if i haven't entered anything into the box

![screen shot 2017-10-24 at 5 26 55 pm](https://user-images.githubusercontent.com/12703638/31952432-de35193c-b8e0-11e7-8a8b-00d787e33253.png)
 (screenshot)

jollopre:
@stephaniesimms that makes more sense! The auto-save is triggered whenever the editor loses the focus but does not check whether or not the text is empty.

[...]

PR #769, opened 2017-10-24
Issue 742

jollopre:
Replaced seconds counter for timeago.js. DMPRoadmap/roadmap#742

### DIFF
--- a/lib/assets/javascripts/utils/timeagoFactory.js
+++ b/lib/assets/javascripts/utils/timeagoFactory.js
@@ -1,0 +1,17 @@
+import enGB from '../locale/en_GB/timeago';
+
+/* global timeago */
+const TimeagoFactory = (() => {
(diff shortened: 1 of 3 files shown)


lib/assets/javascripts/utils/timeagoFactory.js
```diff
@@ -0,0 +1,17 @@
+import enGB from '../locale/en_GB/timeago';
+
+/* global timeago */
+const TimeagoFactory = (() => {
+  timeago.register('en_GB', enGB);
```
briri: I'm going to assume we have a plan to tie this into the user's language selection later and that we'll have ones for es, de, fr, and en_US.

jollopre: yep, we will need to add more languages through register due to the fact that we wanted to change the seconds message...

PR #770, opened 2017-10-24
Default template selection bugfix

xsrust:
updated default template selection logic, added helper function to find default template


briri:
looks good Sam

PR #773, opened 2017-10-24
Bug fix for expand/collapse all links on Plan page

### LINKED ISSUE #730: Expand all plan questions also expands all guidance
On Write Plan page, when I click to expand all questions it also expands all Guidance. Desired behavior is to collapse all Guidance by default (screenshot below). Users can opt to expand individual pieces of guidance or expand all as they proceed with writing the plan.

screen shot 2017-10-16 at 1 08 41 pm

briri:
fixes #730 
- updated JS to only go one level deep when expanding/collapsing accordion sections

### DIFF
--- a/lib/assets/javascripts/utils/expandCollapseAll.js
+++ b/lib/assets/javascripts/utils/expandCollapseAll.js
@@ -40,7 +40,7 @@ export default () => {
     $.each($(el).children('a[data-toggle-direction]'), (i, a) => {
       $(a).click((event) => {
         event.preventDefault();
-        $(`#${accordion} div.panel-collapse`).collapse(`${$(a).attr('data-toggle-direction')}`);
+        $(`#${accordion}`).children('div.panel-collapse').collapse(`${$(a).attr('data-toggle-direction')}`);
       });
     });
   });


lib/assets/javascripts/utils/expandCollapseAll.js
```diff
@@ -40,7 +40,7 @@ export default () => {
     $.each($(el).children('a[data-toggle-direction]'), (i, a) => {
       $(a).click((event) => {
         event.preventDefault();
-        $(`#${accordion} div.panel-collapse`).collapse(`${$(a).attr('data-toggle-direction')}`);
+        $(`#${accordion}`).children('div.panel-collapse').collapse(`${$(a).attr('data-toggle-direction')}`);
```
jollopre: This change works but I just noticed that we have unnecessary loops and const here. This change should work too:

```
$('a[data-toggle-direction]').on('click', (e) => {
    e.preventDefault();
    const target = $(e.target);
    const accordion = target.parent().attr('data-parent');
    $(`#${accordion}`).children('div.panel-collapse').collapse(`${target.attr('data-toggle-direction')}`);
  });
```

Issue #727, opened 2017-10-16
Add "Request expert feedback" to Share plan page


stephaniesimms:
UX wireframe @rhu001 http://h90c3m.axshare.com/#g=1&p=write_plan (see Share tab)

- Add "Request expert feedback" section at bottom of page
- This is an optional feature that should only appear if an Org Admin has turned it on (from the Org details page - related issue #726 )
- Add auto email text below; trigger when user clicks button to "Request feedback"

Email title: [TOOL NAME]: [USER NAME] requested feedback on a plan

Hello [USER NAME]

[USER NAME] has requested feedback on a plan [“PLAN TITLE”]. To add comments, please visit the ‘Plans’ page under the Admin menu in [TOOL NAME] and open the plan. 

All the best
The [TOOL NAME] team

You may change your notification preferences on your profile page. Please do not reply to this email. If you have any questions or need help, please contact us at [HELPDESK EMAIL] or visit [CONTACT US URL].

briri:
**Dev Notes:**
- This will require a way for us to store whether or not this feature is on/off on the `orgs` table.  
- We will also need a place to store a feedback status enum on the `plans` model table.  Statuses needed: Off/nil, Feedback pending, Feedback in progress, Feedback complete (see below for workflow)
- The org details page will need to allow the org_admin the ability to set this value
- Add the button to the write plan page and trigger the email notification. 

**Workflow:**
_Assuming the org admin has enabled this feature in the org details screen_
1. User creates a plan and fills out their answers. 
2. User clicks the 'Request expert feedback' button on the Share plan page (should only be visible to owner/co-owners)
3. Email is sent to the Org Admins for that user's org. The plan's feedback status is set to 'pending'.
4. Org Admin receives email (will also see a notification on their dashboard ... in an future ticket)
5. Org Admin reviews the plan (commenting only ... not editing answers) and provides comments. 
6. When Org Admin attempts to leave the page a modal opens asking them to indicate whether they are still reviewing or have completed the feedback process (see #728) Which sets the `plans.feedback` status to either in progress or complete.
7. If the Org Admin has finished providing feedback an email (see #728) is sent to the owner/co-owners

briri:
We need to determine what email is being overridden on the Org Admin's Org details page. They we will need to make sure that the hooks are put into place to send out the appropriate emails. See PR #776 
Possible emails: 
- Email to plan owner(s) explaining feedback process (triggered when they click 'request feedback')
[...]

Issue #726, opened 2017-10-16
Revise Org Details page 


stephaniesimms:
UX wireframe @rhu001 http://h90c3m.axshare.com/#g=1&p=dmp_organizational_profile__admin_view___tab_versi
Related issue #729  to display up to three links in Org banner

Update the Org branding fields as displayed in wireframe:

- Org full name
- Org abbreviated name (not for display in banner)
- Org URL (e.g., http://cdlib.org/services/uc3)
- Org URL link text (e.g., UC3) - up to 3 links for e.g. data policy...
- Administrator contact email (e.g., uc3@ucop.edu)
- Admin contact email link text (e.g., UC3)
- Org logo file upload
- "Organisational Configuration Information" box should be displayed only for Orgs that have configured Shib (disregard blue notepad functional spec: this should say that the text cannot be edited by Org Admins; it can be edited by Super Admins in the Super Admin area) 

For plan commenting functionality add tab section "Request Feedback" with:

- Radio button that allows admins to turn on function for users to request comments/feedback on plans
- Email subject field and message field for admins to create an auto-email message sent to users who request feedback
- Include grayed-out example text below 

Subject: [TOOL NAME] Your plan has been submitted for feedback

Message: Hello [USER NAME]. Your plan “[PLAN NAME]” has been submitted for feedback from an administrator at your organisation. If you have questions pertaining to this action, please contact us at [ORG CONTACT EMAIL].

briri:
See #727 for Dev notes and the workflow for the Feedback process

briri:
Updated details tab. Moved the org logo section to the top so that it appears above the URLs and contact email. It looked lonely when placed at the bottom.

Note that the 'Add an additional URL' link is not visible in the screenshot because there are already 3 URLs defined.

Updated the org branding bar at the top to use these new links

![screen shot 2017-10-25 at 11 47 24 am](https://user-images.githubusercontent.com/1204467/32017426-81714acc-b97b-11e7-9c0e-2ecd605b97fd.png)

briri:
Added Request Feedback section. Question though regarding what the default email should be (OR do we want to require them to enter a subject and message). I am also unclear on who the intended audience is and at what point the email should get delivered.
[...]

PR #776, opened 2017-10-25
Added Feedback Request to Org Details

briri:
issue #726
- Added contact_name, links, feedback_enabled, feedback_email_subject and feedback_email_msg columns to the orgs table
- Added rake task to convert old orgs.target_url to the links JSON column
- Updated look of org details to match wireframe
  - Separated page into tabs
  - Moved Logo up above URLs and Admin contact
  - Added ability for admin to specify up to 3 URLs
- Updated branding header to display the URLs provided by Admins

You will need to run the following when deploying this one:
- rake db:migrate
- rake migrate:org_target_url_to_links

Open question still on the ticket for the exact text of the email and whether or not the email has a default. Added note to #727 to revisit the user_mailer content later.

### DIFF
--- a/app/controllers/orgs_controller.rb
+++ b/app/controllers/orgs_controller.rb
@@ -16,25 +16,38 @@ class OrgsController < ApplicationController
     attrs = org_params
     @org = Org.find(params[:id])
     authorize @org
-    @org.banner_text = params["org_banner_text"]
-    @org.logo = org_params[:logo] if org_params[:logo]
-
+    @org.logo = attrs[:logo] if attrs[:logo]
+    
+    tab = (attrs[:feedback_enabled].present? ? 'feedback' : 'profile')
+    
+    if attrs[:links].present?
+      if is_json_array_of_objects?(attrs[:links])
+        json = JSON.parse(attrs[:links])
+        # Make sure that the JSON hash is structured as: {"link":"string","text":"string"}
+        if json.all?{ |o| o['link'].present? && o['text'].present? }
+          @org.links = json
+        else
+          redirect_to "#{admin_edit_org_path(@org)}\##{tab}", alert: _('Unable to save your changes. Invalid URLs.')
+        end
+      else
+        redirect_to "#{admin_edit_org_path(@org)}\##{tab}", alert: _('Unable to save your changes. Invalid URLs.')
+      end
+      attrs.delete('links')
+    end
+    
     begin
-      if @org.update_attributes(org_params)
-        flash[:notice] = success_message(_('organisation'), _('saved'))
-        render action: "admin_edit" 
+      if @org.update_attributes(attrs)
+        redirect_to "#{admin_edit_org_path(@org)}\##{tab}", notice: success_message(_('organisation'), _('saved'))
       else
         # For some reason our custom validator returns as a string and not a hash like normal activerecord 
         # errors. We followed the example provided in the Rails guides when building the validator so
         # its unclear why its doing this. Placing a check here for the data type. We should reasses though
         # when doing a broader eval of the look/feel of the site and we come up with a standardized way of
         # displaying errors
-        flash[:alert] = failed_update_error(@org, _('organisation'))
-        render action: "admin_edit"
+        redirect_to "#{admin_edit_org_path(@org)}\##{tab}", alert: failed_update_error(@org, _('organisation'))
       end
     rescue Dragonfly::Job::Fetch::NotFound => dflye
-      flash[:alert] = _('There seems to be a problem with your logo. Please upload it again.')
-      render action: "admin_edit"
+      redirect_to "#{admin_edit_org_path(@org)}\##{tab}", alert: _('There seems to be a problem with your logo. Please upload it again.')
     end
   end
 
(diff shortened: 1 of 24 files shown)


app/mailers/user_mailer.rb
```diff
@@ -44,4 +44,28 @@ def api_token_granted_notification(user)
...
+    if user.org.present?
+      @org = org
+      @plan = plan
+      
+      # Use the generic feedback message unless the Org has specified one
+      subject = "#{_('A user has requested feedback for their DMP in')} #{Rails.configuration.branding[:application][:name]}"
+      subject = org.feedback_email_subject if org.feedback_email_subject.present?
+      
+      # Send an email to all of the org admins as well as the Org's administrator email
+      emails = user.org.users.select{ |usr| usr.can_org_admin? && usr != user }
```
jollopre: why don't you use a scoped query to select can_org_admin users distinct from current. If the org has a considerable amount of users this would not scale...

app/mailers/user_mailer.rb
```diff
@@ -44,4 +44,28 @@ def api_token_granted_notification(user)
...
+  def feedback_notification(user, plan)
+    @user = user
+
+    if user.org.present?
+      @org = org
+      @plan = plan
+      
+      # Use the generic feedback message unless the Org has specified one
+      subject = "#{_('A user has requested feedback for their DMP in')} #{Rails.configuration.branding[:application][:name]}"
+      subject = org.feedback_email_subject if org.feedback_email_subject.present?
```
jollopre: what about a ternary operation?

lib/assets/javascripts/utils/tabHelper.js
```diff
@@ -10,5 +10,6 @@ $(() => {
   $('.nav-tabs').click((e) => {
     $(e.target).tab('show');
     window.location.hash = $(e.target).attr('href');
+    scrollTo(0, 0);
```
jollopre: could you prefix with window.

lib/assets/javascripts/views/orgs/admin_edit.js
```diff
@@ -1,29 +1,67 @@
...
-  // Returns text statistics for the specified editor by id
-  const getStats = (id) => {
-    const body = tinymce.get(id).getBody();
-    const text = tinymce.trim(body.innerText || body.textContent);
-    return {
-      chars: text.length,
-    };
+  // We only allow up to 3 URLs
+  const toggleAddUrlLink = () => {
+    if ($('#org-link-section').find('div.org-link').length >= 3) {
```
jollopre: Today is 3 links, but tomorrow could be more. Please, extract it to the constants file

lib/assets/javascripts/views/orgs/admin_edit.js
```diff
@@ -1,29 +1,67 @@
...
+  });
+
+  $('#edit_org_feedback_form input[type="radio"]').click(() => {
+    toggleFeedback();
+  });
+
+  // Serialize URLs to JSON for form submission
+  $('#edit_org_profile_form').submit(() => {
+    const json = [];
+    $.each($('#org-link-section div.org-link'), (idx, e) => {
```
jollopre: If you use map for traversing is leaner since you need at the end that array of links JSON formatted

app/controllers/orgs_controller.rb
```diff
@@ -16,25 +16,28 @@ def admin_update
     attrs = org_params
     @org = Org.find(params[:id])
     authorize @org
-    @org.banner_text = params["org_banner_text"]
-    @org.logo = org_params[:logo] if org_params[:logo]
+    @org.logo = attrs[:logo] if attrs[:logo]
+    
+    if attrs[:links].present?
+      @org.links = JSON.parse(attrs[:links])
```
jollopre: I would validate that is an array of objects conforming to { link: string, text: string}

lib/assets/javascripts/views/orgs/admin_edit.js
briri: yes agreed. I also ended up pulling in the number-to-text package to update the '(Up to three)' message on the page. We will need to be able to swap in the appropriate locale though here. Maybe we can combine the handler for the one we will use for Timeago.

briri:
Made the requested changes. 

I placed the default email (that is overridable by Org Admins) in the constants.rb initializer but it should ideally be placed somewhere that gets loaded post-FastGetText so that we can take advantage of localization. A partial in the views/user_mailer dir is an option, but something I will address as we handle other tickets for this milestone.

Also added a JS package that converts numbers to text (see `views/orgs/admin_edit.js`) that will need to handle JS localization the same we we do for Timeago when we get to that ticket.

briri:
Note that you MUST run `db:migrate` and `migrate:org_target_url_to_links` rake tasks for this one!

app/mailers/user_mailer.rb
```diff
@@ -44,4 +44,27 @@ def api_token_granted_notification(user)
...
+  
+  def feedback_notification(user, plan)
+    @user = user
+
+    if user.org.present?
+      @org = org
+      @plan = plan
+      
+      # Use the generic feedback message unless the Org has specified one
+      subject = org.feedback_email_subject ||= EMAIL_FEEDBACK_REQUESTED_CONFIRMATION_SUBJECT
```
jollopre: do you really want to assign value to org.feedback_email_subject if is nil?

app/models/user.rb
```diff
@@ -48,7 +48,11 @@ def filter(query)
   # Scopes
   default_scope { includes(:org, :perms) }
 
-
+  # Retrieves all of the org_admins for the specified org
+  scope :org_admins, -> (org_id) {
```
jollopre: nice scoped query

Issue #530, opened 2017-07-21
Two answers saved for single question (Race?)


vyruss:
We found 18 instances of duplicate answers for the same question - this shouldn't happen in the code because we update existing answers. [Observed on DMPonline live service.]

briri:
@vyruss do the timestamps on the records indicate that this is something new since the prod release?

vyruss:
@briri yes, it seems to be present in Roadmap code as no such instances exist for migrated answers

stephaniesimms:
I'm not sure how to test this w/o insight into the database. It sounds like @sjDCC should also test on DMPonline to make sure it's resolved?

sjDCC:
I was only aware of 1 case which came through helpdesk but I don't have permissions to view that plan. @vyruss can you advise how this can be UAT-ed please?

vyruss:
Investigated and handled in #798 , closing

PR #778, opened 2017-10-26
Issue 530

jollopre:
- after_create plan callback for saving answers. answers#update. DMPRoadmap/roadmap#530
- error handling for when an answer does not exist. DMPRoadmap/roadmap#530
- error handling for answering questions at JS. DMPRoadmap/roadmap#530
- optimistic locking fixed. DMPRoadmap/roadmap#530
- fixed tests after create callback for answers. Refactored answers js file.


jollopre:
This PR addresses the issue #530. The bug was found under answers_controller and the reason was "when a plan did not have an answer associated for a question, checking its existence/creation was not atomic and therefore 2 or more threads could potentially create an answer".

In order to mitigate this race condition, I though first the usage of find_or_create from ActiveRecord, however this operation is not atomic. Finally, I decided to implement an after_create callback when a plan is created to automatically insert answers for any question. Using the callback abstract the answers creation and therefore plans can be created anywhere in code.

@vyruss, @briri if you like this approach taken, an additional step will be needed, i.e. we need a task to auto-populate plans before this commit with answer for any question, otherwise when answering a plan, it will prompt an error on the interface since now answers#update does only a question update like it is meant to do.

lib/assets/javascripts/views/answers/edit.js
```diff
@@ -0,0 +1,160 @@
...
+    @param { objecg } - jQueryForm to add events
+  */
+  const attachEventHandlers = (jQueryForm) => {
+    formHandlers({ jQuery: jQueryForm, attachment: 'on' });
+    const tinymceId = jQueryForm.find(`.${editorClass}`).attr('id');
+    Tinymce.init({ selector: `#${tinymceId}` });
+    editorHandlers(Tinymce.findEditorById(tinymceId));
+  };
+  TimeagoFactory.render($('time.timeago'));
+  Tinymce.init({ selector: `.${editorClass}` });
```
briri: I see what you mean about tinymce being a headache ;)

jollopre: :) Yep, since the form is refreshed whenever a stale answer is detected, all the events within that form have to re-created again :(

PR #780, opened 2017-10-27
updates notes_controller for pre-gened answers, added rake tasks to r…

xsrust:
…emove duplicate answers and init answers for all questions


briri:
looks good Sam. Just need to fix the broken test.

Issue #533, opened 2017-07-21
Error on edit profile page when saving changes


sjDCC:
When I try to change certain fields on edit profile e.g. name, organisation, I hit an error and it says save has been unsuccessful. Two fields are highlighted in yellow (recovery email and password) suggesting these need to be re-entered. That's not the desired functionality though so this is a bug.

It's possibly something that was introduced when we were adding new tabs for email notifications so have added in to that milestone to link up.

![capture](https://user-images.githubusercontent.com/6169940/28471611-9f3e2ffc-6e35-11e7-8981-0d086f6a468d.PNG)

vyruss:
@sjDCC the yellow highlighting may be the browser's autocomplete function

xsrust:
Some comments/questions about the intended functionality of this page:
- Should recovery email be mandatory? i.e. does it need to be filled in before the user saves their profile?  It seems odd to require a user enter 2 emails on their profile.
- Should the password field be mandatory on this page?  
  - When I last worked on this page, it was only needed to edit the email field
  - If it is only mandatory for some edits, we need to make this clear to the user, perhaps by show/hiding the field
- Should organisation be mandatory? I believe yes as a user needs to have an organisation for some of the assumptions the code makes.

xsrust:
Also is the intended behavior of the Org-not-listed button to select other from the list of organisations and to show a text field for a user to input their org name?  Currently it just clears out your selection.

sjDCC:
@briri - do you guys need recovery email to be mandatory? It's less important in our context afaik, but I think you may need it to enable SSO unlinking? Maybe it's just mandatory when they unlink or is that too late as they've often lost access to institutional email then?

@xsrust I think we now make first name, last name, email, org, and password mandatory. These are all details that we want every account to have, but I don't think we should be asking anyone to re-enter anything to save changes.

I think we need a mandatory org, even just to state 'other' and enter freetext name

Other views?

xsrust:
The password is mandatory to have with a user account, but the goal of the password field on that page is to verify the identity of the user by entering their current password when doing sensitive actions such as changing the account's email.

There is a separate tab for changing password.

sjDCC:
Ok, thanks for clarifying @xsrust 

I think we want to limit the number of changes that require the password to be entered. I'm happy with that for changing email, but originally it seemed to be needed for any change (e.g. first name or org) which isn't necessary

xsrust:
[...]

PR #782, opened 2017-10-27
Removed need to enter password on profile page except for email

briri:
Issue #533 
- User now only has to enter password when changing their email address
- When user changes their email a modal window opens asking them to enter their password to confirm the change (similar to @jollopre suggestion on the ticket)

### DIFF
--- a/lib/assets/javascripts/views/devise/registrations/edit.js
+++ b/lib/assets/javascripts/views/devise/registrations/edit.js
@@ -1,5 +1,7 @@
 import ariatiseForm from '../../../utils/ariatiseForm';
 import { DISABLE_ORG_COMBO_MESSAGE } from '../../../constants';
+import { isObject, isString } from '../../../utils/isType';
+import { isValidPassword } from '../../../utils/isValidInputType';
 import { addMatchingPasswordValidator, togglisePasswords } from '../../../utils/passwordHelper';
 
 $(() => {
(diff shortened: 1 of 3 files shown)


lib/assets/javascripts/views/devise/registrations/edit.js
```diff
@@ -17,4 +18,30 @@ $(() => {
...
+      const email = $(e.currentTarget).val();
+
+      if (isString(original) && isString(email)) {
+        if (original.toLowerCase() !== email.toLowerCase()) {
+          $('#password-confirmation').modal('show');
+        }
+      }
+    }
+  });
+
```
jollopre: what about pressing Escape or clicking outside the region enabled for the popup? You will need the body of this click handler function too...

lib/assets/javascripts/views/devise/registrations/edit.js
```diff
@@ -17,4 +18,30 @@ $(() => {
     $('#org-controls .combobox-clear-button').hide();
     $('#other_org_toggle a').hide();
   }
+
+  // If the user changes their email address display the password
+  // confirmation modal
+  $('#personal_details_registration_form #user_email').blur((e) => {
```
jollopre: using blur here may break usability principles. I would be more inclined to add another handler for submit, apart from the one we have for validate the fields... and trigger the popup there instead.

briri: yes good point. will refactor to fire on submit if the password hasn't been confirmed

briri:
updated to display the modal on form submission rather than on blur.

lib/assets/javascripts/views/devise/registrations/edit.js
```diff
@@ -17,4 +19,29 @@ $(() => {
     $('#org-controls .combobox-clear-button').hide();
     $('#other_org_toggle a').hide();
   }
+
+  // If the user has changed their email address display the password
+  // confirmation modal on form submission
+  $('#personal_details_registration_form [type="submit"]').click((e) => {
+    const newEmail = $('#personal_details_registration_form #user_email');
```
jollopre: although the selector works, is not needed to select the id user_mail given the parent personal_details_registration_form since IDs must be unique across the whole page.

lib/assets/javascripts/views/devise/registrations/edit.js
```diff
@@ -17,4 +19,29 @@ $(() => {
...
+    if (isObject($('#original_email')) && isObject($(newEmail))) {
+      const original = $('#original_email').val();
+      const email = $(newEmail).val();
+      const pwd = $('#password-confirmation #user_current_password').val();
+
+      // If the user changed the email and has not confirmed their password
+      if (isString(original) && isString(email)) {
```
jollopre: this if is not really needed since those element SHOULD be always string

Issue #731, opened 2017-10-17
Update existing email notifications text


stephaniesimms:
Existing email notifications that need to be updated with revised and standardized text in doc:
https://docs.google.com/document/d/1Zt3QfPZ2q6yMOCVFOevXviwRX-XbZeGSxAAvbRa9w-M/edit

- 2 from devise gem
- existing email text

briri:
Also make sure that we are checking the user's preferences before sending the message!

jollopre:
The email messages have been modified accordingly. @stephaniesimms, you can preview all, except devise related, at:

- `<YOURDOMAIN>/rails/mailers/user_mailer/welcome_notification`
- `<YOURDOMAIN>/rails/mailers/user_mailer/sharing_notification`
- `<YOURDOMAIN>/rails/mailers/user_mailer/permissions_change_notification`
- `<YOURDOMAIN>/rails/mailers/user_mailer/api_token_granted_notification`

stephaniesimms:
@jollopre - updated email text looks good, but [HELPDESK EMAIL] value is missing in final line of each email (see screenshot). "...please contact us at [HELPDESK EMAIL] or visit..."  For roadmap stage the helpdesk email address should be uc3@ucop.edu

![screen shot 2017-11-13 at 12 13 44 pm](https://user-images.githubusercontent.com/12703638/32747028-60f05996-c86c-11e7-9746-a212c6c34e07.png)

briri:
this is a configuration setting in the branding.yml file. I just added the helpdesk_email. It should be working now @stephaniesimms

stephaniesimms:
yes now it's working as expected. thanks @briri . closing issue.

PR #789, opened 2017-11-01
Issue 731

jollopre:
- welcome notification done. DMPRoadmap/roadmap#731
- sharing notification completed. DMPRoadmap/roadmap#731
- permissions change notification done. DMPRoadmap/roadmap#731
- api_token_granted_notification done. DMPRoadmap/roadmap#731
- change password done. shared plan for non-registered user. DMPRoadmap/roadmap#731

### DIFF
--- a/app/models/role.rb
+++ b/app/models/role.rb
@@ -37,6 +37,27 @@ class Role < ActiveRecord::Base
       return 1
     end
   end
+  # Returns a hash of hashes where each key represents an access level (e.g. see access_level method to understand the integers)
+  # This method becomes useful for generatic template messages (e.g. permissions change notification mailer)
+  def self.access_level_messages
+    {
+      3 => {
+        :type => _('co-owner'),
+        :placeholder1 => _('write and edit the plan in a collaborative manner.'),
+        :placeholder2 => _('You can also grant rights to other collaborators.')
+        },
+      2 => {
+        :type => _('editor'),
+        :placeholder1 => _('write and edit the plan in a collaborative manner.'),
+        :placeholder2 => nil,
+        },
+      1 => {
+        :type => _('read-only'),
+        :placeholder1 => _('read the plan and leave comments.'),
+        :placeholder2 => nil,
+      }
+    }
+  end
 
   def set_defaults
     self.active = true if self.new_record?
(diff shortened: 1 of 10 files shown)


app/models/role.rb
```diff
@@ -37,6 +37,27 @@ def access_level
       return 1
     end
   end
+  # Returns a hash of hashes where each key represents an access level (e.g. see access_level method to understand the integers)
+  # This method becomes useful for generatic template messages (e.g. permissions change notification mailer)
+  def self.access_level_messages
+    {
+      3 => {
+        :type => _('co-owner'),
```
briri: looks very useful. Just switch over to the newer `type: _('co-owner'),` syntax instead of `:type => _('co-owner'),`

app/views/devise/mailer/invitation_instructions.html.erb
```diff
@@ -1,8 +1,26 @@
+<%
+  tool_name = Rails.configuration.branding[:application][:name]
+  link = accept_invitation_url(@resource, :invitation_token => @token)
+  helpdesk_email = Rails.configuration.branding[:organisation][:helpdesk_email]
```
briri: same thing here. switch to the newer syntax please

test/mailers/previews/user_mailer_preview.rb
```diff
@@ -0,0 +1,17 @@
+class UserMailerPreview < ActionMailer::Preview
```
briri: Nice. This is a very useful tester

app/models/role.rb
jollopre: The reason I'm using arrow hash is because I don't want to mix old/new syntax for that hash definition. It is needed to preserve the first level Integer keys because:

- access_level method defined above returns Integer and only would match for Integer keys
- and if switched to ':' syntax it raises an error since Integer keys are not coerced to symbols in the new syntax

In case you are not familiar with the new colon syntax, it only coerces keys to symbols for TrueClass, FalseClass and StringClass, the rest of types raises a syntax error. I personally discourage this new syntax unless you are aware that every key gets coerced to symbol, if possible

app/views/devise/mailer/invitation_instructions.html.erb
jollopre: don't think is needed to explain again

PR #793, opened 2017-11-02
removed the default setting from plans.visibility

briri:
DB migration to remove the DB default value for plans.visibility. The default visibility is configurable per installation
- Added rails migration to to remove default from plans.visibility
- Added visibility to all Plan.create calls in tests
- Updated user_mailer views to use the default Rails contact_us_url route helper if an override is not defined in branding.yml

### DIFF
--- a/db/migrate/20171102164156_change_plan_visibility_default.rb
+++ b/db/migrate/20171102164156_change_plan_visibility_default.rb
@@ -1,0 +1,5 @@
+class ChangePlanVisibilityDefault < ActiveRecord::Migration
+  def change
+    change_column_default :plans, :visibility, nil  # default is application configurable
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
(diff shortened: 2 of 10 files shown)


app/views/user_mailer/sharing_notification.html.erb
```diff
@@ -3,7 +3,10 @@
   user_email = @user.email
   link = url_for(action: 'show', controller: 'plans', id: @role.plan.id)
   helpdesk_email = Rails.configuration.branding[:organisation][:helpdesk_email]
-  contact_us_url = Rails.configuration.branding[:organisation][:contact_us_url]
+
+  contact_us = Rails.configuration.branding[:organisation][:contact_us_url] ||= contact_us_url
```
jollopre: please, avoid the usage of memoized operator here. A ternary operator would fit nicer.

briri:
- switched to use ternary instead of memoized logic on contact_us
- moved email signature section that was used across most emails to a partial

briri:
will fix the memoized -> ternary code in a separate PR that addresses the issue of contact_us_url not being defined in the branding config. This one now just contains the change to the plan.visibility default

Issue #434, opened 2017-06-16
Allow Org Admins to view their user's plans & notifications about plans


briri:
UX wireframe @rhu001 - "DMP All Org Plans (Admin View)"
screenshot below

- Provide Org Admins with a list of all plans created by users at their Org that they can view in a read-only (read and comment) mode. Add this page as "Plans" in Admin dropdown menu. 
- Add search box above table
- When a user requests feedback on their DMP, admins need a notification/marker for which plans are awaiting feedback. If plans are awaiting feedback, a Notifications area should appear at the top of the Plans page. 
- Clicking on the link to a plan (e.g., "Plan Title A") takes the user to the write plan view where they can read and add comments.
- Plans should remain in the notifications area until the user/Admin signals that they are done providing feedback by clicking "done" in the modal dialog box #728 

![screen shot 2017-11-07 at 7 39 14 am](https://user-images.githubusercontent.com/12703638/32502145-d60f5796-c38e-11e7-9737-f707d557b436.png)

briri:
Refer to the 'DMP All Org Plans (Admin View)' page from the wireframe above.
**Dev Notes:**
- Add a new 'Plans' link to the Admin menu that brings you to the new page
- Create a plans page that
  - Displays notifications at the top. Notifications currently only consist of plans that have had a request for feedback (see #727)
  - Displays a list of all of the Org's plans. The user should not be able to click through to the plan from this table, just view the contents of the table

sjDCC:
This looks good, thanks @briri 

Two minor points based on John's wireframe:
1. We may want to change the 'filter' function to a search box above the table, though perhaps this is happening as a global change across all tables via another ticket?
2. On the dev notes it says "The user should not be able to click through to the plan from this table, just view the contents of the table" but the wireframe suggests each plan title will be a link to the plan in read-only mode. What was the rationale for changing this? I thought we were giving admins access to plan content too?

stephaniesimms:
@sjDCC #1 is being addressed in another ticket, yes.
#2 is incorrect in the wireframe. or maybe i'm forgetting if we agreed to provide org admins access to plan contents? currently, we only give them a list of plan names and metadata via the DMPTool API and this is sufficient for their needs. plan owners can opt to request feedback to provide admins w/access to the plan contents. the privacy policy probably makes it ok to provide all plan details but my sense is this implementation meets user needs (both owner and admin).

sjDCC:
[...]

PR #795, opened 2017-11-02
Added Request Feedback button to the share plan page

briri:
issue #727
- added new `plans.feedback_requested` flag to DB
- added 'Request Feedback' button to share plan page
- added confirmation email for owner and co-owners
- added notification email for org admins
- minor modifications to tooltips and email defaults (moved email constants to user_mailer)

### DIFF
--- a/app/controllers/plans_controller.rb
+++ b/app/controllers/plans_controller.rb
@@ -7,6 +7,8 @@ class PlansController < ApplicationController
   def index
     authorize Plan
     @plans = current_user.active_plans
+    # Exclude any plans where the user is a reviewer. They access those plans via Admin -> Plans menu
+    @plans.delete_if{ |p| p.reviewable_by?(current_user) }
     @paginable = params[:page]
     @organisationally_or_publicly_visible = @paginable.nil? ? Plan.organisationally_or_publicly_visible(current_user) :
     Plan.organisationally_or_publicly_visible(current_user).page(params[:page])
@@ -179,6 +181,8 @@ class PlansController < ApplicationController
     @plan = Plan.find(params[:id])
     if @plan.present?
       authorize @plan
+      # Get the roles where the user is not a reviewer
+      @plan_roles = @plan.roles.select{ |r| !r.reviewer? }
     else
       redirect_to(plans_path)
     end
@@ -336,6 +340,25 @@ class PlansController < ApplicationController
     else
       render status: :bad_request, json: {code: 0, msg: _("Unable to change the plan's test status")}
     end
+  end
+
+  def request_feedback
+    @plan = Plan.find(params[:id])
+    authorize @plan
+    alert = _('Unable to submit your request for feedback at this time.')
+    
+    begin
+     if @plan.request_feedback(current_user)
+       flash[:notice] = _('Your request for feedback has been submitted.')
+     else
+       flash[:alert] = alert
+     end
+    rescue Exception
+      flash[:alert] = alert
+    end
+    # Get the roles where the user is not a reviewer
+    @plan_roles = @plan.roles.select{ |r| !r.reviewer? }
+    render 'share'
   end
 
   private
--- a/app/mailers/user_mailer.rb
+++ b/app/mailers/user_mailer.rb
@@ -1,6 +1,6 @@
 class UserMailer < ActionMailer::Base
   default from: Rails.configuration.branding[:organisation][:email]
-
+  
   def welcome_notification(user)
     @user = user
     FastGettext.with_locale FastGettext.default_locale do
@@ -31,10 +31,10 @@ class UserMailer < ActionMailer::Base
   def project_access_removed_notification(user, plan, current_user)
     @user = user
     @plan = plan
-                @current_user = current_user
+    @current_user = current_user
     FastGettext.with_locale FastGettext.default_locale do
       mail(to: @user.email, 
-           subject: "#{_('Permissions removed on a DMP in')} #{Rails.configuration.branding[:application][:name]}")
+           subject: "#{_('Permissions removed on a DMP in %{tool_name}') %{ :tool_name => Rails.configuration.branding[:application][:name] }}")
     end
   end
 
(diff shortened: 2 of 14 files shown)


briri:
failing tests for this one are addressed in the other PR, #793, that removed the DB default from `plans.visibility`

app/models/plan.rb
```diff
@@ -573,6 +573,14 @@ def owner
   end
 
   ##
+  # the owner and co-owners of the project
+  #
+  # @return [Users]
+  def owner_and_coowners
+    return User.joins(:roles).where("roles.plan_id = ? AND roles.access >= ?", self.id, 14)
```
jollopre: Nice to see queries to db in the right place. Remember ruby returns the last value executed implicitly so not really needed the return in this case, but it is just a personal preference.

config/initializers/constants.rb
```diff
@@ -1,11 +1,3 @@
 LANGUAGES = (ActiveRecord::Base.connection.table_exists? 'languages') ? Language.sorted_by_abbreviation : []
 MANY_LANGUAGES = LANGUAGES.length > 1
 TABLE_FILTER_MIN_ROWS = 10
-
-# Default Feedback Request Email sent to the requesting user
-# This email can be overriden by local Org Admins on the Org Details page
-# TODO: Move this to a location that gets loaded AFTER FastGettext so that we can  use fastgettext to manage localisations
```
jollopre: Thanks for taking out the lines from here

app/controllers/plans_controller.rb
```diff
@@ -329,6 +329,33 @@ def set_test
     end
   end
 
+  def request_feedback
+    @plan = Plan.find(params[:id])
```
jollopre: did you consider the case for when a Plan id does not exist?

app/mailers/user_mailer.rb
```diff
@@ -1,11 +1,18 @@
 class UserMailer < ActionMailer::Base
   default from: Rails.configuration.branding[:organisation][:email]
 
+  @app_name = Rails.configuration.branding[:application][:name]
```
jollopre: nice refactoring here, however I'm not fan of initialising instance variables outside of its constructor. The two constants only apply for feedback notification method and are gettext functions indeed, I would not create constants for that

app/controllers/plans_controller.rb
briri: I don't think this is an issue ... or at least not something to address as part of this PR.

In the instance where the plan id does not match a valid record in the DB, Rails presents the user with a 404 (see ugly screenshot below from dmponline-test).

In the case where the plan id was missing from the URL, the code in the controller would never execute because Rails would not be able to route to the appropriate controller method.

screen shot 2017-11-08 at 1 46 15 pm

jollopre: It is a silent fail, i.e. not finding the record raises ActiveRecord::RecordNotFound exception with rails nicely maps to 404 http status however it is 5xx indeed since an error has been raised in the server. Catching that with begin end is not a big deal but if we want to be DRY, we should really have in place an exception module to globally handle any error.

briri:
Switched logic to user the new 'Reviewer' role. This role does not appear in the Plan's normal collaborator section and is excluded from the list of plans in My Dashboard. They will be made accessible in the new Admin -> Plans page that will be developed as a part of #434 

Requires `rake db:migrate`

briri:
tests failing because the updated schema.rb file contains a change to the default visibility which is in separate PR. will rebase this one once that PR has been merged to resolve the conflict in this file.

briri:
Schema.rb updated to add default back to plan.visibility. This could create conflict, so merge the plan visibility PR first

app/models/org.rb
```diff
@@ -142,6 +142,10 @@ def check_api_credentials
     end
   end
 
+  def org_admins
+    User.joins(:perms).where("users.org_id = ? AND perms.id IN (?)", self.id, [5, 6, 7, 9])
```
jollopre: using the name of the perm, although probably less fast would be more readable

app/models/plan.rb
```diff
@@ -189,8 +189,36 @@ def get_guidance_group_options
...
+    # Send an email to all of the org admins as well as the Org's administrator email
+    if user.org.contact_email.present?
+      admins << User.new(email: user.org.contact_email, firstname: user.org.contact_name)
+    end
+    admins.each do |admin|
+      UserMailer.feedback_notification(admin, self, user).deliver_now
+    end
 
-
+    self.save!
```
jollopre: this line would be more adequate right after L199, so that if fails saving roles would be no inserted and no one will receive email...

app/models/plan.rb
```diff
@@ -320,6 +348,17 @@ def owned_by?(user_id)
   end
 
   ##
+  # determines if the plan is reviewable by the specified user
+  #
+  # @param user_id [Integer] the id for the user
+  # @return [Boolean] true if the user can administer the plan
+  def reviewable_by?(user_id)
+    user_id = user_id.id if user_id.is_a?(User)
```
jollopre: user_id I believe is an integer so for never will be an object of class User

app/views/plans/_share_form.html.erb
```diff
@@ -30,7 +30,10 @@
       </tr>
     </thead>
     <tbody>
-      <% plan_roles = @plan.roles.where(active: true) %>
+      <% 
+        # Exclude reviewers from the list 
+        plan_roles = @plan.roles.where("active = ? AND access < ?", true, 16)
```
jollopre: the bit field gem we are using got nice build in helpers to avoid hardcoding the access integer. For example, @plan.roles.where({ active: true}).where(Role.not_reviewer_condition)

briri: ah ok. I will check out the gem then. I tried :reviewer but it was not working. will look into that

app/controllers/plans_controller.rb
briri: agreed. Something in exception module or application_controller.rb or perhaps even an ActiveRecord Concern (new concept in Rails 4.x ... we don't make use of them yet). Can you create a ticket for this? Something we can look into post MVP

app/models/org.rb
briri: yes that's true. will change to use names

app/models/plan.rb
briri: move the feedback_requested flag down to the end instead. The idea is that the flag would not be set in the event that any of the other more fragile steps fail (emails and role assignments)

app/models/plan.rb
briri: true, should be. I copied it from other ones Sam created like `administerable_by?` etc. Assumed that he did that because there was a situation where the User object was being supplied instead of the id. Going to leave it as is for now. We should change all of them if they aren't needed

app/views/plans/_share_form.html.erb
briri: refactored and move line to controller as well

briri:
made most of the requested changes. Lets create a ticket to handle our server errors better (see above comments). Ping me on Slack if you get a chance to review this one @jollopre. I'll deploy once its merged so things are ready for testing on Monday

app/models/plan.rb
```diff
@@ -189,8 +189,35 @@ def get_guidance_group_options
...
+    # Send an email to all of the org admins as well as the Org's administrator email
+    if user.org.contact_email.present?
+      admins << User.new(email: user.org.contact_email, firstname: user.org.contact_name)
+    end
+    admins.each do |admin|
+      UserMailer.feedback_notification(admin, self, user).deliver_now
+    end
+    
+    self.feedback_requested = true
+    self.save!
```
jollopre: this is safer right before sending any email. In case it failed, both creators and potential reviewers will receive an email for a non-existing request feedback for a plan.

app/models/plan.rb
```diff
@@ -573,6 +611,14 @@ def owner
   end
 
   ##
+  # the owner and co-owners of the project
+  #
+  # @return [Users]
+  def owner_and_coowners
+    User.joins(:roles).where("roles.plan_id = ? AND roles.access >= ?", self.id, 14)
```
jollopre: Could you explain to me why just 14? If the method is doing what I thinking... "give all the owners and co-owners of the current instance plan"... it should be:

- roles.plan_id = 'foo_id' && roles.access in (1,3,5,7,9,11,13,15) OR roles.access in (2,3,6,7,10,11,14,15)

briri: Well, we only have 5 available options actually stored in the DB (which is why I opted for 14 instead of all the possible values). We could probably just have used an Int enum for this column though. Since we only ever have 5 values in the DB:
```
Roles                                              Access           Nbr Recs
-----------------------------------------------------------------------------------
commenter                                          8                404
editor + commenter                                 12               1145
administrator + editor + commenter                 14               1532
creator + administrator + editor + commenter       15               18052
commenter + reviewer                               24               30
```

I found an even better solution though by digging through the gem's code (they're documentation isn't the best): `access_values_for(:flag)` It returns all of the values for the flag.
`Role.access_values_for(:reviewer, :commenter).min` returns 24
`Role.access_values_for(:creator).concat(Role.access_values_for(:administrator))` returns all of the values you referenced above

app/models/plan.rb
briri: Ok. This is a better solution. I refactored to add the Roles to `self.roles` and then send the emails if `self.save!` succeeds. The flag and roles are saved in a single transaction.

Issue #612, opened 2017-08-31
Allowing default templates to be customised


sjDCC:
Currently org-admins only have options to customise the funder templates, but not the default template (i.e. Tuuli or DCC generic DMP) which are presented when there is no relevant funder or institutional template. Ideally this would show up on the admin preview and allow people to add example answers, guidance and additional sections/questions as they can for the funder ones.

We suggest adding a third table in the middle for 'Default template(s)' and to provide the same 'customise' functionality as in the funder area
![add-default-templates-to-customise](https://user-images.githubusercontent.com/6169940/29916996-657de258-8e38-11e7-9c3f-bdc5fdf4e3cb.PNG)

@vyruss has suggested adding a flag to organisational templates to say they can be customised. Alternatively, we could display this by which template is set as default in the system, not by org type.

This is a request for Tuuli but would be useful in the main codebase

jollopre:
This issue has been fixed on CDL-MVP since we have new HTML markup. Specifically, the list of default templates visible to the end user are those:
- whose organisation they belong have org_type as 'organisation'
- are published
- are not migrated
- and is_default attribute is set to true

stephaniesimms:
new table for "Default templates" available on Template page. table styling will be addressed in separate issues. closing this issue. @sjDCC can you check to make sure this is the desired outcome here?

sjDCC:
Yes, looks good. Given that it's only 1 template as default, we may want to place it above the funder customisation ones (it can get lost on scroll below) but no biggie so happy to close out.

stephaniesimms:
I agree w/ @sjDCC that it would be good to place this at the top of the page. also, my understanding is that there should only ever be 1 default template. if this is true, please change the text to "Default template" (singular) and move it to the top @jollopre

sjDCC:
Yes, there will only ever be one default as far as i know

jollopre:
Changes made according to your comments:

- Default templates appears before Funder templates
- Default templates(s) message dynamically updated depending on the number of templates returned. This is more flexible in case in future we want to display more than one...

sjDCC:
Excellent thanks @jollopre

stephaniesimms:
looks good. closing issue.

PR #796, opened 2017-11-03
Issue 612

jollopre:
Default templates placed before funder templates with dynamic plural. DMPRoadmap/roadmap#612


app/views/templates/admin_index.html.erb
```diff
@@ -83,12 +83,12 @@
     <a href="<%= admin_new_template_path %>" class="btn btn-primary" role="button">
       <%= _('Create a template') %>
     </a>
+    <% if !current_user.org.organisation? %>
```
briri: All orgs should be able to see the default template in this list (unless they are the owner), so other organisations, funders and institutions should be able to customize the default.
@sjDCC can clarify if necessary

Issue #777, opened 2017-10-26
No templates available to create a new plan


stephaniesimms:
I can click through the wizard to create a new plan but no template is served up. When I tick the box that no funder is associated I should get the default template but there is no Write Plan tab (screenshot 1). If I select a funder template I get a json message w/the template ID (screenshot 2).

![screen shot 2017-10-26 at 9 36 20 am](https://user-images.githubusercontent.com/12703638/32040558-91776bfc-ba31-11e7-95a8-1b6c19cb9022.png)

![screen shot 2017-10-26 at 9 37 14 am](https://user-images.githubusercontent.com/12703638/32040569-98e101dc-ba31-11e7-84e3-c5f7e893b1a6.png)

briri:
This seems to be working for me now. I think one of the other PRs must have broken and then fixed it again. Please check again on stage

stephaniesimms:
@briri this is still not working for me. i get the same kinds of errors noted above for default template and just some json info for any other template.

stephaniesimms:
copying our slack comments over - default template is working now but the issue persists w/funder templates. based on the screenshot of the Template page below (note N/A status) it looks like a separate piece of template work may be causing the issue...

![screen shot 2017-11-01 at 2 27 34 pm](https://user-images.githubusercontent.com/12703638/32300116-49583782-bf15-11e7-8c19-adad5c3ee311.png)

stephaniesimms:
this is working as expected now. closing issue.

PR #797, opened 2017-11-03
fixed issues with plan create page

briri:
addresses plan creation issues on #777 
- separated out shared controller endpoint that was handling ajax template retrieval and form submission
- added error handler to display notice to user when no templates were found
- disable submit button until a template has been returned
- adjusted integration tests to work with new json endpoint

### DIFF
--- a/app/controllers/plans_controller.rb
+++ b/app/controllers/plans_controller.rb
@@ -43,9 +43,11 @@ class PlansController < ApplicationController
 
     # If the template_id is blank then we need to look up the available templates and return JSON
     if plan_params[:template_id].blank?
-      templates = template_options(org_id, funder_id)
-      render json: {"templates": templates.collect{|t| {id: t.id, title: t.title} }}.to_json
-      
+      # Something went wrong there should always be a template id
+      respond_to do |format|
+        flash[:alert] = _('Unable to identify a suitable template for your plan.')
+        format.html { redirect_to new_plan_path }
+      end
     else
       # Otherwise create the plan
       @plan.principal_investigator = current_user.surname.blank? ? nil : "#{current_user.firstname} #{current_user.surname}"
@@ -336,7 +338,6 @@ class PlansController < ApplicationController
     end
   end
 
-
   private
   def plan_params
     params.require(:plan).permit(:org_id, :org_name, :funder_id, :funder_name, :template_id, :title, :visibility,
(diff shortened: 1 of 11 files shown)


app/controllers/plans_controller.rb
```diff
@@ -330,6 +332,52 @@ def set_test
   end
 
 
+  # Collect all of the templates available for the org+funder combination
+  # --------------------------------------------------------------------------
+  def template_options()
```
jollopre: Have you considered moving this logic to the templates model?

lib/assets/javascripts/views/plans/new.js
```diff
@@ -49,14 +62,14 @@ $(() => {
...
       $('#plan_template_id option').remove();
 
-      // Fetch the available templates fbased on the funder and research org selected
+      // Fetch the available templates based on the funder and research org selected
       const jQueryForm = $('form#new_plan');
       const formElements = jQueryForm.serializeArray();
       $.ajax({
-        method: getMethod(jQueryForm),
-        url: getAction(jQueryForm),
+        method: 'POST',
```
jollopre: why are you hardcoding the method, could not you get it from the form method?

lib/assets/javascripts/views/plans/new.js
```diff
@@ -1,17 +1,32 @@
...
+    } else {
+      $('#new_plan button[type="submit"]').attr('disabled', true)
+        .attr('data-toggle', 'tooltip').attr('title', NEW_PLAN_DISABLED_TOOLTIP);
+    }
+  };
+
   // AJAX error function for available template search
   const error = () => {
-    // TODO: add error handling
+    renderAlert(NO_TEMPLATE_FOUND_ERROR);
```
jollopre: Nice to see here the usage of notification  messages through JS too

lib/assets/javascripts/views/plans/new.js
briri: Its not using the form's `action` in this case so I was reluctant to assume that the method would be the same.

app/controllers/plans_controller.rb
briri: yes, that would be more appropriate

briri:
- moved template_options endpoint from the plans controller to template controller
- switched to the more appropriate GET method instead of POST
- also rebased with latest from CDL-MVP

jollopre:
Please, fix the merge conflicts and merge back yourself

Issue #798, opened 2017-11-06
Update Answers Behavior on Live Service


xsrust:
Currently we have an issue with the creation of answers in the notes/answers controllers.
This has led to duplicate answers for some questions of some plans.

To fix this, the selection logic for answers on notes/answers controllers will be standardised to be by plan and question, not by an answer_id.  Secondly, the creation action will be re-factored to create the answer within a transaction which should mitigate the race condition

vyruss:
Should affect MVP too

jollopre:
@sjDCC could you try this scenario:

1. Create a new plan
2. Save a note BEFORE answering a question
3. Attempt to save answer for the question you noted something

sjDCC:
Thanks for the clarification @jollopre 

This seems to all be working. I can add a note and save this without affecting the question/answer. I only get the unsaved changes notice when I edit the main answer box. This saved and exported fine. I re-edited and exported and all was working as expected.

Closing out

jollopre:
@sjDCC thanks! This is good then

PR #799, opened 2017-11-06
updated answers and notes controller for #798

xsrust:
Wraps answer query/creation in transactions

Addresses bug on live service:
https://github.com/DMPRoadmap/roadmap/issues/798


xsrust:
Tests will fail, but these were already failing in development, see: 
https://github.com/DMPRoadmap/roadmap/pull/784

Issue #751, opened 2017-10-19
is being logged in really an error


mxsstrong:
I invited someone to a plan (actually myself, but just my gmail account) and clicked on the link to accept the invitation.  

I was directed to the application but it said Error you are already logged in.  
Is this really an error?

Here's the email (separate issue but could use some formatting)

Hello mxs4ever@gmail.com,

A colleague has invited you to contribute to their Data Management Plan at DMPRoadmap - STAGE.

Click here to accept the invitation (or copy https://roadmap-stg.cdlib.org/users/invitation/accept?invitation_token=6LfYmtPwZ-WUy3Sh2MQF into your browser).

If you don't want to accept the invitation, please ignore this email.
Your account won't be created until you access the link above and set your password.

All the best,
The DMPRoadmap - STAGE team.

![screen shot 2017-10-19 at 12 06 58 pm](https://user-images.githubusercontent.com/810543/31789287-86b8b5ce-b4c6-11e7-86c3-80907c5bb2a9.png)

mxsstrong:
Do we have other categories of messages other than error - Alert, Warning, etc

briri:
We're using the standard Rails flash messages (alert and notice). I agree that isn't really an error (or at least not a negative thing anyway). We will have to dig into the code though to see if this message is being thrown by one of our controllers or within the devise gem that we're using to handle authentication.

sjDCC:
I suspect this is because your browser was already logged in as marisa.strong@ucop when you followed the invite token issued to mxs4...

I'd noticed an error the other day which displayed in the standard yellow alert. Are we not colour coding them or is that to come?

briri:
noted the desire to set the notice/alert colors to mimic those in the current DMPonline site in #746

jollopre:
@sjDCC was right, this message comes up due to the user being signed in when a link for a new user creation according to the token is passed.

I have override the flash alert with a more sensible message "**Cannot validate invitation token unless you sign out**". We can certainly change it to a flash notice instead. 

Please, let me know if you come up with a better message to be displayed.

sjDCC:
That message sounds fine to me @jollopre. I think it's going to be a rare case anyhow, when the likes of us are testing sharing with multiple email addresses.

mxsstrong:
[...]

PR #806, opened 2017-11-08
Issue 751

jollopre:
Override flash alert message for when user is signed in and a URL invitation token is provided. DMPRoadmap/roadmap#751

### DIFF
--- a/app/controllers/users/invitations_controller.rb
+++ b/app/controllers/users/invitations_controller.rb
@@ -1,0 +1,14 @@
+class Users::InvitationsController < Devise::InvitationsController
+  protected
+    # Override require_no_authentication method defined at DeviseController (parent of Devise::InvitationsController)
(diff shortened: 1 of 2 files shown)


app/controllers/users/invitations_controller.rb
```diff
@@ -0,0 +1,10 @@
+class Users::InvitationsController < Devise::InvitationsController
+  protected
+    # Override require_no_authentication method defined at DeviseController (parent of Devise::InvitationsController)
+    def require_no_authentication
+      super
+      if flash[:alert].present?
+        flash[:alert] = _('Cannot validate invitation token unless you sign out')
```
briri: Looks good @jollopre just update the text of the alert here to match @mxsstrong recommendation in the ticket (switch :alert to :notice): "Notice: You are already logged in as another user. Please log out to activate your invitation"

jollopre:
@briri the message is now according to @mxsstrong suggestion.

Issue #638, opened 2017-09-15
Wrong error message when creating account for existing email


sjDCC:
When a user tries to create an account for an email already taken in the system, but using a different mix of upper and lowercase lettering, the wrong error is thrown:

![create-account-error](https://user-images.githubusercontent.com/6169940/30483600-5426c93c-9a1f-11e7-8b98-8148d14bfeae.PNG)

They only get the correct error that the email is taken when the case sensitivity matches.
![correct-error](https://user-images.githubusercontent.com/6169940/30483614-66e6e85e-9a1f-11e7-923d-2166550c63e3.PNG)

This is related to #550

briri:
seems to have been fixed along with #550.

briri:
@sjDCC can you verify that this is fixed in production? It seems to be working in CDL-MVP. I'd like to make sure that the fix made it into production though before closing

sjDCC:
No, this is still throwing the wrong error on production.

I tested this by trying to register with Sarah.Jones@glasgow.ac.uk. If I do the same with lowercase (i.e. sarah.jones@glasgow.ac.uk) I get the correct 'email address already registered' notice.

@vyruss can we deploy this fix before closing please?

jollopre:
@sjDCC now the tool should not allow you to create an account for an existing email anymore. You can test it whenever is deployed

stephaniesimms:
tested on roadmap stage and working as expected. closing.

Issue #322, opened 2017-05-18
Can't create an account if email already known by system


sjDCC:
Sometime plans are shared and the invitee doesn't follow the link. If so, there's a stem account that lies unvalidated in the system. If the holder of that invited email address subsequently tries to create an account they get ugly error messages. See #290 

At minimum we need an error message to say their email address is already taken, but if they don't have the email invite to follow the relevant link they're stuck, so a better workflow / solution is needed. Perhaps redirect them to a page to set a password for the unvalidated account as would be done when they follow to invite?

xsrust:
Perhaps we should redirect them via the change password workflow instead of set password?  The account has another person's plan shared with it so it's perhaps a little lax to just let them set a password without any confirmation that they own that email address.  (Just took a Computer security exam thismorning so perhaps paranoid)

sjDCC:
@vyruss was just talking about this on slack and said that they need the token so can't redirect to set / change password like I suggested. May need to delete stem account

vyruss:
Fixed for now by automatically deleting stem account when they try to sign up and letting them sign up normally. When a better solution comes along, we can apply that.

sjDCC:
This works now. Closing out

PR #807, opened 2017-11-09
Issue 638

jollopre:
- Sign up email case insensitive.
- Removed session key previous_url since does not work well. 
- Duck-typed User.where_case_insensitive method.

 DMPRoadmap/roadmap#638

### DIFF
--- a/app/controllers/registrations_controller.rb
+++ b/app/controllers/registrations_controller.rb
@@ -41,19 +41,20 @@ class RegistrationsController < Devise::RegistrationsController
 
   # POST /resource
   def create
-    #logger.debug "#{sign_up_params}"
     if !sign_up_params[:accept_terms] then
       redirect_to after_sign_up_error_path_for(resource), alert: _('You must accept the terms and conditions to register.')
     else
-      existing_user = User.find_by_email(sign_up_params[:email])
-      if !existing_user.nil? # If email exists
-        if (existing_user.password == "" || existing_user.password.nil?) && existing_user.confirmed_at.nil? # If user has not accepted invitation yet
-          existing_user.destroy # Only solution for now
-          super
-        else
+      existing_user = User.where_case_insensitive('email', sign_up_params[:email]).first
+      if existing_user.present?
+        if existing_user.accept_terms?
           redirect_to after_sign_up_error_path_for(resource), alert: _('That email address is already registered.')
+          return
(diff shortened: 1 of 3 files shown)


app/controllers/registrations_controller.rb
```diff
@@ -41,18 +41,12 @@ def new
 
   # POST /resource
   def create
-    #logger.debug "#{sign_up_params}"
     if !sign_up_params[:accept_terms] then
       redirect_to after_sign_up_error_path_for(resource), alert: _('You must accept the terms and conditions to register.')
     else
-      existing_user = User.find_by_email(sign_up_params[:email])
-      if !existing_user.nil? # If email exists
-        if (existing_user.password == "" || existing_user.password.nil?) && existing_user.confirmed_at.nil? # If user has not accepted invitation yet
```
briri: Why remove this? It was added to handle situations where a invitation email was sent but the user tries to create an account outside the invitation process. See #322. Devise creates the user record in the DB when it sends out the invitation so the user cannot create an account because the email is already known.

app/models/user.rb
```diff
@@ -285,10 +285,9 @@ def deliver_invitation(options = {})
   ##
   # Case insensitive search over User model
   # @param field [string] The name of the field being queried
-  # @param val [string] The string to search for, case insensitive
+  # @param val [string] The string to search for, case insensitive. val is duck typed to check whether or not downcase method exist
   # @return [ActiveRecord::Relation] The result of the search
   def self.where_case_insensitive(field, val)
-    User.where("lower(#{field}) = ?", val.downcase)
+    User.where("lower(#{field}) = ?", val.respond_to?(:downcase) ? val.downcase : val.to_s)
```
briri: a good safety check here. you could also have checked `is_a?(String)` but this is fine too.

jollopre: val is always an object with the downcase method check is more than enough here

app/controllers/registrations_controller.rb
jollopre: The logic introduced covers more sensibly the edge case for #322. Note before it was just a super, not executing our override lines for when a user does not exist.

briri: ok. will test it out once deployed to make sure #322 hasn't been reintroduced

PR #809, opened 2017-11-09
Added handling for contact-us override in branding.yml

briri:
- Allows the `contact_us_url` in branding.yml to be left blank so that the application's built-in contact-us form is used
- Added logic to footer.html.erb layout file to use the correct url
- Added a email_signature.html.erb partial that is used by most emails


app/views/user_mailer/_email_signature.html.erb
```diff
@@ -0,0 +1,16 @@
+<%
```
jollopre: Nice partial, it was copied/pasted across many mails...

config/branding_example.yml
```diff
@@ -9,6 +9,11 @@ defaults: &defaults
     copywrite_name: 'Curation Centre (CC)'
     email: 'tester@cc_curation_centre.org'
     helpdesk_email: 'someone@somewhere.com'
+
```
jollopre: Nice having this documentation here

config/initializers/contact_us.rb
```diff
@@ -27,4 +27,4 @@
   # Set the following variable to true if you are using localized paths
   # e.g. /en/contact-us OR /fr/contact-us
   config.localize_routes = true
```
jollopre: We are not using that anymore, and if so it would be query param for the language...

Issue #677, opened 2017-10-10
Bug on admin preview


sjDCC:
On the template preview on the admin area, the guidance text is being displayed in both guidance and example answer fields. This is incorrect. it should only show in the guidance panel.

If there is no guidance and an example answer is added it displays correctly, however if guidance exists, this overrides any entry in the example answer field. I'm sure this is only happening on the admin preview as I haven't seen any errors on the front end.

![admin-preview-error](https://user-images.githubusercontent.com/6169940/31391277-fb50c354-adcd-11e7-948a-a13d852bd5fe.PNG)

dsisu:
Any news on this issue? I received user enquiry.

jollopre:
After speaking with @sjDCC last Thursday, it has been agreed to be fixed in CDL-MVP branch. The preview functionality will be an attempt to view the template as if the admin user was writing a plan for that template.

sjDCC:
Just tested and looks good - closing out

PR #810, opened 2017-11-13
Issue 677

jollopre:
- phases edit view and its sub-templates using local variables instead of instance variables. DMPRoadmap/roadmap#677
- conditional displaying of sub-templates at edit phase. This permits being re-used when previewing a template phase. DMPRoadmap/roadmap#677

### DIFF
--- a/app/controllers/phases_controller.rb
+++ b/app/controllers/phases_controller.rb
@@ -7,13 +7,13 @@ class PhasesController < ApplicationController
   # GET /plans/:plan_id/phases/:id/edit
   def edit
 
-    @plan = Plan.eager_load2(params[:plan_id])
+    plan = Plan.eager_load2(params[:plan_id])
     # authorization done on plan so found in plan_policy
-    authorize @plan
+    authorize plan
 
     phase_id = params[:id].to_i
-    @phase = @plan.template.phases.select {|p| p.id == phase_id}.first
-    @readonly = !@plan.editable_by?(current_user.id)
+    phase = plan.template.phases.select {|p| p.id == phase_id}.first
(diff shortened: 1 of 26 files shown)


jollopre:
Note, the preview template functionality is now re-using the view phases/edit_plan_answers so that we don't have duplication of code anymore for displaying a phase

briri:
The code looks fine but I had concerns that the JS behind the answer autosaving would still fire when in preview mode. I deployed this branch locally and noticed the following issues.
- I don't see a 'Preview' button to access the preview pages
- I also tested the Plan edit page and am getting a 'Couldn't find Answer' message during the answer autosave (see screenshot below). I'm guessing this has to do with the conversion of instance variables to locals

screen shot 2017-11-14 at 10 54 08 am

jollopre:
@briri the template preview is accessed per phase, i.e. you have to select a phase and the preview button will be visible. It renders /phases/_edit_plan_answers.rb conditionally removing any HTML tag where a plan variable is needed.

Using instance variables in partials prevents re-using them unless the starting point, i.e. the controller action defines explicitly that variable. I would encourage all of us to avoid them if possible.

The JS that handles auto-saving is now aware of readonly mode and therefore does not add event handlers for forms (e.g. change, submit) neither does it for tinymce events (e.g. blur, focus). 

I have thoroughly tested both views (plan edit and template preview) and could not reproduce the screenshot you attached.

briri:
resolved a couple of merge conflicts introduced by PR #832 
I kept everything from the issue677 branch with the exception of the `plan = eager_load2`. I used the line from the PR but switched it from instance variable `@plan` to `plan`

Issue #733, opened 2017-10-17
Add email notice for new Comments (not admin feedback)


stephaniesimms:
Email title: [TOOL NAME]: A new comment was added to [PLAN TITLE]

Hello [USER NAME]

[COMMENTER NAME] has commented on the plan [“PLAN TITLE”]. To view the comments, please visit the My Dashboard page in [TOOL NAME] and open your plan. 

All the best
The [TOOL NAME] team

You may change your notification preferences on your profile page. Please do not reply to this email. If you have any questions or need help, please contact us at [HELPDESK EMAIL] or visit [CONTACT US URL]

jollopre:
I have assumed that the owner of the plan is the one who receives the mail regarding a new comment. The view can be tested at <HOSTNAME>/rails/mailers/user_mailer/new_comment

stephaniesimms:
you assumed correctly @jollopre!  i'll wait for the PR to test

stephaniesimms:
I can't add or save comments to test this now. When I create a new plan the Comments field will only display if there is no Guidance (screenshot - first question show Comments tab highlighted but no editor field). Where there is a Comments field, I can enter text but clicking Save does nothing (screenshot - second question can't save comment).

![screen shot 2017-11-13 at 3 59 16 pm](https://user-images.githubusercontent.com/12703638/32755762-a361b99a-c88b-11e7-9e85-ad6d707ef8cb.png)

jollopre:
@stephaniesimms, I have fixed that problem, not related to the issue though, so you should be able to add a comment and see whether or not ONLY the owner receives an email.

briri:
Same thing as #732 

contact_us_url is not in the branding.yml file nor should it have to be if the installation wants to use the application's built in contact us page. This line overrides Rails' built-in url helper: `contact_us_url = Rails.configuration.branding[:organisation][:contact_us_url]`

I updated this line in the other mailers to use Rails' built-in helper if an override is not defined in the yaml file. Will submit a PR to correct this one.

jollopre:
@briri, I think it would be easier to change the local variable contact_as_url to something that that not override any url helper from Rails. At least it's safer.

sjDCC:
Yes, seems to just be emailing out to the owner as expected. Closing out

stephaniesimms:
sorry i missed your note earlier @jollopre to clarify who should receive emails - it should be the plan owner and all co-owners. please update to include co-owners.

Issue #732, opened 2017-10-17
Add email notices for plan visibility changes


stephaniesimms:
Add the 3 new email notifications below (all 3 have same Email title), triggered automatically when a user changes plan visibility on Share page. User can also select to deactivate these auto email notices on Notification Preferences page under Edit Profile.

Email title: DMP Visibility Changed: [PLAN TITLE]

Hello [USER NAME]

The plan [“PLAN TITLE”] had its visibility changed to private.

Visibility definitions:
- Private: restricted to me and my collaborators
- Organisation: anyone at my organisation can view
- Public: anyone can view on the web

If you have questions pertaining to this action, please visit the My Dashboard page in [TOOL NAME]. 

All the best
The [TOOL NAME] team

You may change your notification preferences on your profile page. Please do not reply to this email. If you have any questions or need help, please contact us at [HELPDESK EMAIL] or visit [CONTACT US URL]

Hello [USER NAME]

The plan [“PLAN TITLE”] has had its visibility changed to institutional.

Visibility definitions:
- Private: restricted to me and my collaborators
- Organisation: anyone at my organisation can view
- Public: anyone can view on the web

If you have questions pertaining to this action, please visit the My Dashboard page in [TOOL NAME]. 

All the best
The [TOOL NAME] team

You may change your notification preferences on your profile page. Please do not reply to this email. If you have any questions or need help, please contact us at [HELPDESK EMAIL] or visit [CONTACT US URL]

Hello [USER NAME]

The plan [“PLAN TITLE”] has had its visibility changed to public.

Visibility definitions:
- Private: restricted to me and my collaborators
- Organisation: anyone at my organisation can view
- Public: anyone can view on the web

If you have questions pertaining to this action, please visit the My Dashboard page in [TOOL NAME]. 

All the best
The [TOOL NAME] team

You may change your notification preferences on your profile page. Please do not reply to this email. If you have any questions or need help, please contact us at [HELPDESK EMAIL] or visit [CONTACT US URL]

briri:
This will rely on #760

jollopre:
The mail for visibility changes in plan can be seen at <HOSTNAME>/rails/mailers/user_mailer/plan_visibility

stephaniesimms:
I created a new plan and tried to test visibility changes to trigger email notices on the Share page. Changing my plan visibility to "Org" or "Public" produced an error (screenshot 1) and did not send an email notice @jollopre . 

[...]

PR #827, opened 2017-11-14
Fixes for visibility, comment and feedback emails

briri:
- Visibility email: Allow contact_us_url to be undefined in branding.yml (use Rails' built-in route helper if not defined) #732 
- Comment email: Allow contact_us_url to be undefined in branding.yml (use Rails' built-in route helper if not defined) #733 
- Feedback emails: Add `raw` to feedback emails to properly handle inline HTML #727

### DIFF
--- a/app/views/user_mailer/new_comment.html.erb
+++ b/app/views/user_mailer/new_comment.html.erb
@@ -4,7 +4,7 @@
   plan_title = @plan.title
   user_name = @plan.owner.name
   helpdesk_email = Rails.configuration.branding[:organisation][:helpdesk_email]
-  contact_us_url = Rails.configuration.branding[:organisation][:contact_us_url]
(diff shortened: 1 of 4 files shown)


app/views/user_mailer/new_comment.html.erb
```diff
@@ -4,7 +4,7 @@
   plan_title = @plan.title
   user_name = @plan.owner.name
   helpdesk_email = Rails.configuration.branding[:organisation][:helpdesk_email]
-  contact_us_url = Rails.configuration.branding[:organisation][:contact_us_url]
+  contact_us = Rails.configuration.branding[:organisation][:contact_us_url] || contact_us_url
```
jollopre: that change should be sufficient for avoiding the override of contact_us_url helper. The one defined at the .yml it is within a hash so no problem should arise.

PR #828, opened 2017-11-15
update password confirmation modal to auto-submit form

briri:
submit form after password confirmation is entered on Profile page #533 
- added ids to buttons on modal
- added JS to submit form on 'Continue' click

### DIFF
--- a/lib/assets/javascripts/views/devise/registrations/edit.js
+++ b/lib/assets/javascripts/views/devise/registrations/edit.js
@@ -1,5 +1,4 @@
 import ariatiseForm from '../../../utils/ariatiseForm';
-import { DISABLE_ORG_COMBO_MESSAGE } from '../../../constants';
 import { isObject, isString } from '../../../utils/isType';
 import { isValidPassword } from '../../../utils/isValidInputType';
 import { addMatchingPasswordValidator, togglisePasswords } from '../../../utils/passwordHelper';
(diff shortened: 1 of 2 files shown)


lib/assets/javascripts/views/devise/registrations/edit.js
```diff
@@ -44,4 +44,9 @@ $(() => {
   $('#password-confirmation #user_current_password').change((e) => {
     $('#password-confirmation #user_password').val($(e.target).val());
   });
+
+  // Submit the form when the user clicks the confirmation button on the modal
+  $('#password-confirmation #pwd-confirmation').click((e) => {
```
jollopre: could you please remove the first part of the query selector (e.g. #password-confirmation)? you don't need to look for children of password-confirmation since id MUST be unique across the page. If we have duplicate ids across a page, this should be changed otherwise HTML validators would complain.

briri: sure. we had some issues in the past with some of the inputs auto-generated by the Rails helpers (e.g. #user_id) and JS for pages like edit profile affecting the login modal for example. They're not necessary here though so will remove.

briri: Can change that one, but the others need to be explicit like that otherwise they'll affect the login/create account modals (e.g. `$('#password-confirmation #user_current_password').change()`

briri:
@jollopre updated that one line. The others though need to be explicit and include the form identifier otherwise `#user_password` would effect both this confirmation modal and the login/create modals

briri:
removed unused import from JS file that was causing linter failure

Issue #811, opened 2017-11-13
Check preferences before sending emails


briri:
Update the logic for each email to first check the user's preferences to determine whether or not an email should be sent.

jollopre:
The PR #850, addresses a subset of conditional emails that we are sending. The following table shows any message we are sending and how they relate to the notification preferences in the user profile. Note, emails from devise regarding sign up, forgot password, etc are not displayed here since I'm assuming they have to be sent always. 

We have some emails that are sent regardless of any condition (e.g. status TODO?) as well as some conditional emails that need clarification (e.g. status NEEDS CLARIFICATION). @stephaniesimms and @sjDCC, please can you give me a sense for these TODO? and NEED CLARIFICATION

Notification preferences | key associated (branding.yml) | Email subject | Status
------------ | -------------|------------ | -------------|
A new comment has been added to my DMP | users.new_comment | %{tool_name}: A new comment was added to %{plan_title} | DONE
A plan has been shared with me | users.added_as_coowner | A Data Management Plan in %{tool_name} has been shared with you | DONE
Admin privileges granted to me | users.admin_privileges | Changed permissions on a Data Management Plan in %{tool_name} | NEEDS CLARIFICATION
My DMP's visibility has changed | owners_and_coowners.visibility_changed | DMP Visibility Changed: %{plan_title} | DONE
An organisation template is published | admins.template_published | NOT FOUND | TODO?
An organisation template is unpublished | admins.template_unpublished | NOT FOUND | TODO?
A user has requested feedback on a DMP | admins.feedback_requested | %{application_name}: %{user_name} requested feedback on a plan | DONE
NOT FOUND | NOT FOUND | Feedback confirmation message (can be customised by orgs) |  TODO?
NOT FOUND | NOT FOUND | API rights in %{tool_name} | TODO?
NOT FOUND | NOT FOUND | Permissions removed on a DMP in %{tool_name} | TODO?

sjDCC:
Here are my thoughts. @stephaniesimms can confirm too

Email subject / notification  | Who to send to and when |Comments 
------------ | -------------|------------ | 
 new comment has been added to my DMP| Alert owners and co-owners to change | Should be able to configure on email notification preferences|
A plan has been shared with me|Alert user that DMP has been shared and by whom. Email should confirm rights e.g. co-owner, editor, read-only | Shouldn't be able to opt out of this. No conditions - should send irrespective of preferences |
[...]

PR #829, opened 2017-11-15
Issue 811 (DO NOT MERGE)

jollopre:
- conditional_user_mailer concern for delivering email according to preferences. DMPRoadmap/roadmap#811
- unit test for the concern. DMPRoadmap/roadmap#811


jollopre:
I have implemented the logic through a rails concern. Let me know if you like this approach and I will update any UserMailer call accordingly.

PR #830, opened 2017-11-15
Updated Request Feedback default on Org details page

briri:
- Added placeholder and changed tooltip #726 
- Moved feedback confirmation message defaults to `helpers/mailer_helper.rb`
- Updated feedback logic to prevent adding the reviewer role if the user already has a role on the plan (current association allows multiple roles of the same type per user+plan). the user does not need the reviewer role if they are already a commenter, editor or co-owner/owner since they have the same or better permissions than reviewer.  
- Updated `mailers/user_mailer.rb` and `views/orgs/_feedback_form.html.erb` to use MailHelper

### DIFF
--- a/app/helpers/mailer_helper.rb
+++ b/app/helpers/mailer_helper.rb
@@ -1,0 +1,20 @@
+module MailerHelper
+
+  def feedback_confirmation_default_subject
+    _('%{application_name}: Your plan has been submitted for feedback')
+  end
+
+  def feedback_confirmation_default_message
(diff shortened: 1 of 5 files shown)


app/helpers/mailer_helper.rb
```diff
@@ -0,0 +1,20 @@
+module MailerHelper
```
jollopre: the code here makes more sense than were it was, thanks

PR #833, opened 2017-11-16
Add Plans page to Org Admin section

briri:
- added org admin plans page #434


briri:
rebased with latest changes to CDL-MVP

briri:
Need to refactor emails and confirmation for completion

Issue #786, opened 2017-10-30
Create page breaks for long tables


stephaniesimms:
Implement the following rule for any table (e.g., of plans, templates, guidance, user accounts, etc) that exceeds 10 records. See screenshot of UX wireframe below:

- create a new page with links for sequential ordering: 1, 2, 3 ... Next >, Last >>
- add a "View All" link on lower left. if user clicks link, expand the table to display all records.
- change link to say "View less" when complete list is visible. if user clicks this link, revert back to paginated display.

![screen shot 2017-10-30 at 12 09 02 pm](https://user-images.githubusercontent.com/12703638/32190543-5e5eaaea-bd6b-11e7-877e-96afa601bf1e.png)

briri:
@stephaniesimms and @sjDCC this has been deployed to stage for your review. It is currently only applied to the organisational plans in the 2nd table on the my dashboard page. If all looks good we can open a ticket to add it in for other tables throughout the site.

I think it should start out with pagination by default. Right now it starts out as expanded with a 'View Less' link.

stephaniesimms:
looks good @jollopre - a few changes:
- desired behavior is to only use pagination if the table exceeds 10 records. so no need to display a "View less" link as in screenshot 1 (below) since there is only one plan.
- if a table does contain > 10 records then the default should be to present it w/pagination. the user can then click "View all" to expand and view the complete table.
- for users w/no Org affiliation (listed as "Other") there should be no Org table on the My Dashboard page (screenshot 2). remove "Other Plans"

![screen shot 2017-11-01 at 2 46 48 pm](https://user-images.githubusercontent.com/12703638/32299732-e31efe5c-bf13-11e7-925a-b0f7cda2d51b.png)

![screen shot 2017-11-01 at 2 50 06 pm](https://user-images.githubusercontent.com/12703638/32299787-1a73d45e-bf14-11e7-9aef-173e70fa0f89.png)

sjDCC:
This sounds good. I also wanted to check what the conditions are for wider university plans displaying. Is institutional visibility the default or do they only appear when users have chosen to make them public or institutional - and they're over 50% complete etc?
I ask as quite a few show in the second tables which will probably be a surprise to DMPonline users. If we don't have 'private' as the default setting we should probably consult before release.

briri:
[...]

PR #836, opened 2017-11-20
Issue 786 (DO NOT MERGE)

jollopre:
- paginable module to accomodate any model that needs paginated views. DMPRoadmap/roadmap#786
- 
- paginable concern. DMPRoadmap/roadmap#786
- 
- scope and local keywords for paginable to make it more generic. DMPRoadmap/roadmap#786
- 
- accessing to default_per_page instance variable from Kaminari. DMPRoadmap/roadmap#786
- 
- integration test for paginable/plans_controller. DMPRoadmap/roadmap786
- 
- commented out test. DMPRoadmap/roadmap#786


app/controllers/plans_controller.rb
```diff
@@ -1,17 +1,13 @@
 class PlansController < ApplicationController
   require 'pp'
   helper SettingsTemplateHelper
-
+  helper PaginableHelper
```
briri: Can you not just include the Concern directly?

jollopre: We need a helper since the concern method (e.g. paginable_renderise) is only available as an instance method for the controller paginable/plans_controller but it is used in the view (views/plans/index.html) so therefore a helper is needed.

app/controllers/paginable/plans_controller.rb
```diff
@@ -0,0 +1,24 @@
...
+  end
+  # GET /paginable/plans/organisationally_or_publicly_visible/:page
+  def organisationally_or_publicly_visible
+    # TODO authorize
+    if params[:page] == 'ALL'
+      plans = Plan.organisationally_or_publicly_visible(current_user)
+    else
+      plans = Plan.organisationally_or_publicly_visible(current_user).page(params[:page])
+    end
+    paginable_renderise(partial: 'organisationally_or_publicly_visible', scope: plans)
```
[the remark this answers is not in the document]
jollopre: We need to create controller actions for any model we want to paginate. That's the reason I created the namespace paginable/ since any link used for paginable stuff (e.g. View All, View Less, First, 1...N, Last) when clicked, sends an AJAX request to an specific action under the scope of paginable controllers. 

I'm going to try to be as DRY as possible (e.g. not checking ALL in both places paginable_renderise and in each controller action...

app/controllers/plans_controller.rb
briri: ah ok. didn't realize that

PR #842, opened 2017-11-20
Fixes html issue with feedback confirmation email

briri:
Added a feedback_confirmation.html.erb to allow Rails to properly handle HTML content being provided from the DB. #726


app/mailers/user_mailer.rb
```diff
@@ -64,20 +64,21 @@ def feedback_notification(recipient, plan, requestor)
   end
   
   def feedback_confirmation(recipient, plan, requestor)
-    @user = requestor
+    user = requestor
```
jollopre: nice to see instance variables going away

Issue #746, opened 2017-10-18
CSS refinements


briri:
Now that the Bootstrap overrides are in place on roadmap-stg we should begin another round of review. @dsisu, @sjDCC and @stephaniesimms please take a look at the site now and record any refinements to color, fonts, borders, spacing, alignment, etc. in this ticket.

stephaniesimms:
- Action menus in tables are cut off on right and bottom of page when user clicks to open the menu (screenshots 1-3)
- When resizing pages on different browsers/devices, the footer wraps and obscures things at bottom of page (e.g., Save, Cancel buttons on Template editing pages, screenshots 4-5) 
- Template editing pages/workflow in Admin area need help - not sure what the desired behavior is but they are extremely confusing (create separate issue for this)

1- Guidance page - Actions menu cut off at bottom of table
<img width="1173" alt="screen shot 2017-10-18 at 12 35 42 pm" src="https://user-images.githubusercontent.com/12703638/31790232-aad9ee48-b4c9-11e7-9170-35051e41d343.png">


2- Guidance page - Actions menu cut off at bottom and side of table, need more space between table and footer
<img width="1189" alt="screen shot 2017-10-18 at 12 36 31 pm" src="https://user-images.githubusercontent.com/12703638/31790247-ae0b4c10-b4c9-11e7-9d49-a094fa4a9cc9.png">


3- Templates page - same problems with Actions menu and footer
<img width="1219" alt="screen shot 2017-10-18 at 12 37 37 pm" src="https://user-images.githubusercontent.com/12703638/31790254-b0961adc-b4c9-11e7-88de-ffcfce89b468.png">


4-5 Templates page - Save, Cancel buttons obscured by footer when resizing page
<img width="800" alt="screen shot 2017-10-19 at 10 59 21 am" src="https://user-images.githubusercontent.com/12703638/31790470-6b5eadca-b4ca-11e7-99de-c307893b8c1c.png">

<img width="747" alt="screen shot 2017-10-19 at 10 59 35 am" src="https://user-images.githubusercontent.com/12703638/31790476-6fa908d0-b4ca-11e7-9fb5-2b49332da30b.png">

briri:
Looks like `btn-default` and `btn-primary` have different top margin settings. They should be the same
![screen shot 2017-10-27 at 11 12 35 am](https://user-images.githubusercontent.com/1204467/32118877-315e8c64-bb08-11e7-892d-bf4c65732027.png)

[...]

PR #847, opened 2017-11-21
CSS Refinements (#746)

bhavi:
Remove the Filter Box from table views

### DIFF
--- a/app/views/public_pages/template_index.html.erb
+++ b/app/views/public_pages/template_index.html.erb
@@ -16,7 +16,7 @@
       <div class="table-responsive">
       <table class="table table-hover tablesorter">
         <thead>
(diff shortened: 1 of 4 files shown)


app/views/public_pages/template_index.html.erb
```diff
@@ -16,11 +16,11 @@
       <div class="table-responsive">
       <table class="table table-hover tablesorter">
         <thead>
-          <% if @templates.count > TABLE_FILTER_MIN_ROWS %>
+          <% if @plans.count > TABLE_FILTER_MIN_ROWS %>
```
briri: Think this is a copy-paste error. Update to `@templates.count` and  `colspan="3"` and `public_templates_path, placeholder: _('Filter templates')}` below

Issue #728, opened 2017-10-16
Add modal for Admin to notify user when "done" commenting


stephaniesimms:
UX wireframe @rhu001 http://h90c3m.axshare.com/admin_finish_review_dialog_box.html

Once admin follows notification link to the write plan page for the plan that they are going to provide feedback on, whenever they attempt to leave the write plan tab, they will see the modal box pop up asking them if they are "done" commenting and ready to send notification back to user or if they just want to save existing comments and return later. 

Since comments are autosaved, revise the modal text as follows:
"No, I still have some more work to do. I'll finish providing feedback at a later time."

When an admin clicks "Send" the system should trigger the following email:

Email title: [TOOL NAME]: Expert feedback has been provided for [PLAN TITLE]

Hello [USER NAME]

[COMMENTER NAME] has finished providing feedback on the plan [“PLAN TITLE”]. To view the comments, please visit the My Dashboard page in [TOOL NAME] and open your plan. 

All the best
The [TOOL NAME] team

You may change your notification preferences on your profile page. Please do not reply to this email. If you have any questions or need help, please contact us at [HELPDESK EMAIL] or visit [CONTACT US URL]

briri:
See #727 for Dev notes and the workflow for the Feedback process

briri:
Need to clarify text on modal.

stephaniesimms:
Since comments are autosaved, you can revise the modal text as follows:
"No, I still have some more work to do. I'll finish providing feedback at a later time."

jollopre:
FYI @stephaniesimms, comments are not auto-saved, you have to click save/update otherwise they are not persisted.

stephaniesimms:
ok @jollopre - if there will be no autosave functionality for comments then the modal text is fine as is. no changes necessary.

briri:
A JS solution for this would not be ideal since we're doing auto saves and do not have a good way to trigger a modal popup. Wiring one to the comment's auto-save JS would be too noisy.

Spoke with @stephaniesimms and we decided to just add a 'Complete' link to the Org Admin -> Plans notification section (see screenshot). The Org Admin accesses the plan via this page, so they will see it before starting the feedback process. The 'Complete' link has a tooltip that reads: 'Notify the plan owner that I have finished providing feedback'
[...]

PR #856, opened 2017-11-22
Org Admin provide feedback

briri:
- Add Plans page to Org Admin section #434
- Add 'Complete' action to feedback notification area or Org Admin plans page #728
- Refactored old methods used to check a user's access level for a plan (e.g. Plan.readable_by?) They did not allow for multiple roles
- Added new Email preferences for `[users][feedback_requested]` and `[users][feedback_provided]`
- Refactored emails to use the _email_signature partial
- Fixed Notes Controller which was not assigning a role for the test user on the test plan

**Run db:migrate and update you branding.yml with the new email prefs**

### DIFF
--- a/app/models/plan.rb
+++ b/app/models/plan.rb
@@ -196,36 +196,69 @@ class Plan < ActiveRecord::Base
   #  emails org admins and org contact 
   #  adds org admins to plan with the 'reviewer' Role
   def request_feedback(user)
-    val = Role.access_values_for(:reviewer, :commenter).min
-    self.feedback_requested = true
+    Plan.transaction do
+      begin
+        val = Role.access_values_for(:reviewer, :commenter).min
+        self.feedback_requested = true
     
-    # Share the plan with each org admin as the reviewer role
-    admins = user.org.org_admins
-    admins.each do |admin|
-      role = Role.new(user: admin, access: val)
-      self.roles << role unless self.users.include?(admin)
-    end 
+        # Share the plan with each org admin as the reviewer role
+        admins = user.org.org_admins
+        admins.each do |admin|
+          self.roles << Role.new(user: admin, access: val)
+        end 
 
-    if self.save!
-      # Send an email confirmation to the owners and co-owners
-      self.owner_and_coowners.each do |owner|
-        UserMailer.feedback_confirmation(owner, self, user).deliver_now
+        if self.save!
+          # Send an email confirmation to the owners and co-owners
+          deliver_if(recipients: self.owner_and_coowners, key: 'users.feedback_requested') do |r|
+            UserMailer.feedback_confirmation(r, self, user).deliver_now
+          end
+          # Send an email to all of the org admins as well as the Org's administrator email
+          if user.org.contact_email.present?
+            admins << User.new(email: user.org.contact_email, firstname: user.org.contact_name)
+          end
+          deliver_if(recipients: admins, key: 'admins.feedback_requested') do |r|
+            UserMailer.feedback_notification(r, self, user).deliver_now
+          end
+          true
+        else
+          false
+        end
+      rescue Exception => e
+        Rails.logger.error e
+        false
       end
-  
-      # Send an email to all of the org admins as well as the Org's administrator email
-      if user.org.contact_email.present?
-        admins << User.new(email: user.org.contact_email, firstname: user.org.contact_name)
-      end
-
-      deliver_if(recipients: admins, key: 'admins.feedback_requested') do |r|
-        UserMailer.feedback_notification(r, self, user).deliver_now
-      end
-      true
-    else
-      false
     end
   end
 
+  ##
+  # Finalizes the feedback for the plan:
+  #  emails confirmation messages to owners
+  #  sets flag on plans.feedback_requested to false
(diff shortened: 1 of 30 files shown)


app/models/plan.rb
```diff
@@ -314,8 +349,7 @@ def add_guidance_to_array(guidance_array, guidance_group, theme, guidance)
   # @return [Boolean] true if user can edit the plan
 	def editable_by?(user_id)
     user_id = user_id.id if user_id.is_a?(User)
-    role = roles.where(user_id: user_id).first
-    return role.present? && role.editor?
+    has_role(user_id, :editor)
```
briri: Refactored these because the feedback process assigns the 'reviewer' role to all org admins but they may have also had another role on the plan. The `first` in the old code was preventing these functions from working accurately in that situation

app/models/user.rb
```diff
@@ -263,13 +263,19 @@ def self.from_omniauth(auth)
   #
   # @return [JSON] with symbols as keys
   def get_preferences(key)
-    if self.pref.present? && self.pref.settings[key.to_s].present?
-      return self.pref.settings[key.to_s].deep_symbolize_keys
-    elsif Pref.default_settings
-      return Pref.default_settings[key.to_sym] || Pref.default_settings[key.to_s]
-    else
-      return nil
+    existing = self.pref.settings[key.to_s].deep_symbolize_keys
```
briri: Refactored this because it does not allow for new email preferences to be added to the defaults. The hash returned to the controller is used to build the page. The hash looks like: `{'users': {'key1':'val1', 'key2':'val2'},'admins': ...}`. The check was always just pulling the data from the DB and ignoring new prefs for the 'users' key. 
Instead updated to add new defaults to the existing prefs. We would need to refactor if we ever allow deeper nesting within this hash

app/models/user.rb
```diff
@@ -263,12 +263,23 @@ def self.from_omniauth(auth)
   #
   # @return [JSON] with symbols as keys
   def get_preferences(key)
-    if self.pref.present? && self.pref.settings[key.to_s].present?
-      return self.pref.settings[key.to_s].deep_symbolize_keys
-    elsif Pref.default_settings
-      return Pref.default_settings[key.to_sym] || Pref.default_settings[key.to_s]
+    defaults = Pref.default_settings[key.to_sym] || Pref.default_settings[key.to_s]
```
briri: Refactored this because it does not allow for new email preferences to be added to the defaults. The hash returned to the controller is used to build the page. The hash looks like: `{'users': {'key1':'val1', 'key2':'val2'},'admins': ...}`. The check was always just pulling the data from the DB and ignoring new prefs for the 'users' key.
Instead updated to add new defaults to the existing prefs. We would need to refactor if we ever allow deeper nesting within this hash

app/models/plan.rb
```diff
@@ -196,37 +196,70 @@ def get_guidance_group_options
...
+            admins << User.new(email: user.org.contact_email, firstname: user.org.contact_name)
+          end
+          deliver_if(recipients: admins, key: 'admins.feedback_requested') do |r|
+            UserMailer.feedback_notification(r, self, user).deliver_now
+          end
+          true
+        else
+          false
+        end
+      rescue Exception => e
```
jollopre: rescue from Exception it is considered bad practice, you should handle only exceptions that you know how to recover (e.g. ActiveRecord::RecordInvalid, etc)

app/models/plan.rb
```diff
@@ -196,37 +196,70 @@ def get_guidance_group_options
...
 
-      deliver_if(recipients: admins, key: 'admins.feedback_requested') do |r|
-        UserMailer.feedback_notification(r, self, user).deliver_now
+  ##
+  # Finalizes the feedback for the plan:
+  #  emails confirmation messages to owners
+  #  sets flag on plans.feedback_requested to false
+  #  removes org admins from the 'reviewer' Role for the Plan
+  def complete_feedback(org_admin)
+    Plan.transaction do
```
jollopre: nice using here transaction to be safe about the feedback_requested flag.I guess is needed since multiple admins can review a plan simultaneously, otherwise it might be overkill...

app/models/plan.rb
jollopre: well spotted

app/models/plan.rb
```diff
@@ -725,6 +749,15 @@ def visibility_allowed?
 
   private
 
+  # Returns whether or not the user has the specified role for the plan
+  def has_role(user_id, role_as_sym)
+    if user_id.is_a?(Integer) && role_as_sym.is_a?(Symbol)
```
jollopre: be aware that checking types is an expensive operation, if this method is used only within this class, I would make it private and remove the type checking

app/models/user.rb
jollopre: I guess this new logic is trigger anytime we access to user.get_preferences for someone who has pref in db which might be inefficient. You might need to re-consider this later on.

app/models/plan.rb
briri: Ok. we can look at readdressing later. This was intended to capture both ActiveRecord failures as well as Mailer failures, log the specific failure and rollback any DB activity that occurred so that the feedback wasn't left in a semi-complete state.

Issue #857, opened 2017-11-22
Unable to change plan collaborator privileges


stephaniesimms:
A handful of related bugs to report:
1 - When I select a new level of privileges in the dropdown under Manage Collaborators I cannot save the change (screenshot 1). I assume it should autosave.
2 - As a co-owner I cannot remove collaborators. I receive the error message in Screenshot 2. Co-owners should have the same privileges as plan owners. They should only be prevented from removing themselves and the plan owner, but should be authorized to remove other collaborators. 
3 - This error message should be displayed on the Share plan page in cases where a user does try to perform an action that isn't allowed. Currently, it redirects me to My Dashboard.

![screen shot 2017-11-22 at 1 41 14 pm](https://user-images.githubusercontent.com/12703638/33151225-03a869b4-cf8c-11e7-8028-3a8e6cb93c71.png)

![screen shot 2017-11-22 at 1 41 33 pm](https://user-images.githubusercontent.com/12703638/33151227-05ca6724-cf8c-11e7-8d55-141d20880648.png)

sjDCC:
This is related to ticket #840 which asks for autosave to be reinstated so permissions can be revised

jollopre:
The PR #861 is addressing the list of issues encountered. Summarising:

- Autosaving when combobox (Coowner, Editor, Read only) changes
- Remove action is permitted for owners and coowners
- Action (update permission and destroy permission, i.e. combobox and remove link respectively) are only permitted for owner and coowners. Note, they cannot perform these actions on themselves.

sjDCC:
All looks good:
1) autosave has been added and reload jumps to top to make user aware of notification
2) co-owners can be removed, but not yourself
3) messages display on share plan page and don't redirect

Closing out

PR #861, opened 2017-11-24
Issue 857

jollopre:
- Autosave to change permissions on a plan for collaborators. DMPRoadma…p/roadmap#840
- destroy role action allowed for users that can administer a plan. DMPRoadmap/roadmap#857
- simplified logic to handle permissions for collaborators in a plan. DMPRoadmap/roadmap#857

### DIFF
--- a/app/models/role.rb
+++ b/app/models/role.rb
@@ -41,8 +41,29 @@ class Role < ActiveRecord::Base
       return 1
     end
   end
+
+  # Sets access_level according to bit fields defined in the column access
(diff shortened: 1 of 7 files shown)


app/models/role.rb
```diff
@@ -41,8 +41,29 @@ def access_level
       return 1
     end
   end
+
+  # Sets access_level according to bit fields defined in the column access
+  # TODO refactor according to the hash defined above (e.g. 1 key is :creator, 2 key is :administrator, etc)
```
briri: Yes, this is a much better location for this method

Issue #545, opened 2017-07-27
Replace "Shared?" column in plans table


stephaniesimms:
UX wireframe @JEK-III - http://h90c3m.axshare.com/#g=1&p=my_dashboard__plans___ideal
It was helpful to be able to see which plans have been shared w/others at a glance (screenshot below). Can you cram that column back into the plans table @JEK-III ? or recommend another approach?

![screen shot 2017-07-27 at 8 15 29 am](https://user-images.githubusercontent.com/12703638/28678634-bda030b6-72a5-11e7-8e9c-cdebcd3fac07.png)

JEK-III:
Crammed!

stephaniesimms:
i think just a simple Yes or No is best here @bhavi - you can remove the parenthetical w/number of users plans has been shared with. the use case is to see which plans are being co-authored at a quick glance.

![screen shot 2017-11-30 at 7 03 48 am](https://user-images.githubusercontent.com/12703638/33437455-dab68b64-d59c-11e7-9e57-587a03e2f4d2.png)

briri:
Yes, reviewer role counts towards being shared

stephaniesimms:
working as expected now. closing issue.

PR #864, opened 2017-11-27
Add shared column in Plans table on Dashboard (#545)

bhavi:
Issue #545

### DIFF
--- a/app/models/plan.rb
+++ b/app/models/plan.rb
@@ -642,6 +642,15 @@ class Plan < ActiveRecord::Base
   end
 
   ##
+  # returns the shared roles of a plan, excluding the creator
+  def shared
+    role_values = Role.access_values_for(:commenter)
+                      .concat(Role.access_values_for(:editor))
+                      .concat(Role.access_values_for(:administrator)).uniq! - Role.access_values_for(:creator)
+    Role.where(plan: self, access: role_values)
+  end
+
+  ##
   # the owner and co-owners of the project
   #
   # @return [Users]
--- a/app/views/paginable/plans/_privately_visible.html.erb
+++ b/app/views/paginable/plans/_privately_visible.html.erb
@@ -15,6 +15,7 @@
               <th><%= _('Role') %></th>
               <th class="text-center sorter-false"><%= _('Test') %></th>
               <th><%= _('Visibility') %></th>
+              <th><%= _('Shared') %></th>
               <th class="sorter-false">
                 <span aria-hidden="false" class="sr-only"><%= _('Actions') %></span>
               </th>
@@ -41,6 +42,13 @@
                 </td>
                 <td class="plan-visibility">
                   <%= plan.visibility === 'is_test' ? _('N/A') : raw(display_visibility(plan.visibility)) %>
+                </td>
+                <td>
+                  <% unless plan.shared.empty? %>
+                    <%= _("Yes(#{plan.shared.length})") %>
+                  <% else %>
+                    <%= _('No') %>
+                  <% end  %>
                 </td>
                 <td>
                   <div class="dropdown">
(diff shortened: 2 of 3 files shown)


app/models/plan.rb
```diff
@@ -617,6 +617,21 @@ def owner
...
   end
 
+  ##
+  # returns the shared roles of a plan, excluding the creator
+  def shared
+    role_values = Role.access_values_for(:commenter)
+                      .concat(Role.access_values_for(:editor))
+                      .concat(Role.access_values_for(:administrator)).uniq! - Role.access_values_for(:creator)
+    shared_roles = []
+    self.roles.each do |role|
```
briri: You're not using the `role_values` you collected above efficiently! `self.roles` is going to run a DB query for every role.
Use `Role.where(plan: self, access: role_values)` instead. It will hit the DB once and return any roles that are not 'creator'. No need to loop through the roles association

app/views/paginable/plans/_privately_visible.html.erb
```diff
@@ -42,6 +43,13 @@
                 <td class="plan-visibility">
                   <%= plan.visibility === 'is_test' ? _('N/A') : raw(display_visibility(plan.visibility)) %>
                 </td>
+                <td>
+                  <% unless plan.shared.empty? %>
+                    <%= _("Yes(#{plan.shared.size})") %>
```
briri: Use `.length` instead. `.size` or `.count` hit the DB whereas `.length` just counts the objects in the association array

test/unit/plan_test.rb
```diff
@@ -274,6 +274,18 @@ class PlanTest < ActiveSupport::TestCase
...
 
+  # ---------------------------------------------------
+  test "returns the shared roles" do
+    #create admin role
+    admin = Role.create(user: User.first, plan: @plan, access: 14)
+    #create commenter role
+    reader = Role.create(user: User.last, plan: @plan, access: 8)
+
+    # assert that the plan is shared with above roles and doesnt include owner
+    assert_equal(@plan.shared.size, 2)
```
briri: again, use `.length` here

briri:
also note that the new test is failing in Travis

app/models/plan.rb
```diff
@@ -641,6 +641,15 @@ def owner
     User.joins(:roles).where('roles.plan_id = ? AND roles.access IN (?)', self.id, vals).first
   end
 
+  ##
+  # returns the shared roles of a plan, excluding the creator
+  def shared
+    role_values = Role.access_values_for(:commenter)
```
jollopre: If you want to express "any role that is not creator" you could use the following statement:

Role.where(plan: self).where(Role.not_creator_condition) which would be less verbose

bhavi: Well, we also need to get a count of the users its shared with  (as per the  wireframe)

briri: `Role.not_creator_condition` returns a String for SQL querying. The results are the same for me when I run yours and Jose's to get the list of shares:
```ruby
Role.access_values_for(:creator)
# Array of Integers --> [1, 3, 5, 7, 9, 11, 13, 15, 17, 19, 21, 23, 25, 27, 29, 31]
Role.not_creator_condition
# String --> (roles.access not in (1,3,5,7,9,11,13,15,17,19,21,23,25,27,29,31))
```

jollopre: @bhavi, note you might need to exclude the reviewer role too, in that case the query should be: Role.where(plan: self).where(Role.not_creator_condition).where(Role.not_reviewer_condition).any?

otherwise a plan will be displayed as shared when someone request feedback on his/her plan

dsisu:
Please revisit Jose's comments. Thanks!

PR #879, opened 2017-11-30
added missing index for questions_themes

briri:
While working with the dmptool migration data I noticed that one of the join tables was missing an index.


xsrust:
I'm not sure we need an index on themes for that table.
We generally query it by question to get a list of themes, not the other way around.

Issue #840, opened 2017-11-20
Add autosave to share plan collaborators table


sjDCC:
The share plan collaborators table needs to be on auto-save
![autosave](https://user-images.githubusercontent.com/6169940/33034878-2c14867e-ce21-11e7-9375-a5414f55bcd1.PNG)

At present this doesn't seem to be the case so there's no way to change permission without removing and re-granting access

jollopre:
Related to issue #857 and fixed in PR #861

sjDCC:
The autosave works now thanks. I noticed that removing rights jumps you to top of page so you see notice about saved changes (as popup to confirm choice is raised) but just changing permissions doesn't obviously reload page so you can easily miss notice that changes have saved unless you scroll up.

![autosave](https://user-images.githubusercontent.com/6169940/33481945-a33205ea-d68e-11e7-80d0-55fa36bd6809.PNG)

Do we want to reload and jump to top? It would be helpful here, but annoying perhaps on other autosave pages so happy to close this out if not desired.

jollopre:
@sjDCC, we can jump to the top of the page when a change is made at any collaborator.

sjDCC:
Ok, it's probably worth doing that to make sure people notice it's saved and see the notification.

Do you want to pull back to 'in progress' to implement?

jollopre:
yes, I will make that tweak

sjDCC:
Looks perfect, thanks @jollopre

PR #881, opened 2017-12-01
Issue 840

jollopre:
scrollTo element applicable for when plan permission changes. DMPRoadmap/roadmap#840


jollopre:
To be consistent, we should apply same principle to each AJAXABLE event, i.e. scrolling the window to the position in which the message (response from the server) has been added

briri:
looks good Jose. Agreed that we should add this to other ajaxable events (where applicable). User testing/feedback will let us know if its too jarring.

Issue #885, opened 2017-12-01
Add lists to Tinymce


briri:
Add lists back to the Tinymce editor. Bulleted and numbered lists are referenced in the toolbar setting but 'lists' is missing from the plugins entry for the Tinymce editor config.

Also add a max-height value to prevent the box from appearing very tall before it receives focus.

sjDCC:
Checked and all displays well. Thanks @briri Closing out

Issue #874, opened 2017-11-28
Create template layout adjustments


stephaniesimms:
Layout adjustments to standardize template creation/editing page:

Phase layout:
- reduce "Order of display" box to half column width
- shorten TinyMCE editor box to standard size (it's too long)
- Move Save and Cancel buttons to lower right
![phase-layout](https://user-images.githubusercontent.com/12703638/33338812-0a240d3a-d42c-11e7-8a2b-969543be9f3a.png)

Section layout:
- reduce "Order of display" box to half column width
- Description box should be standard TinyMCE editor box (same as phase editor)
- Buttons should be "Save" and "Cancel" across all sections (change "Delete" to "Cancel" here)
- Move Save and Cancel buttons to lower right
- Align Save/Cancel buttons with Add Question button below
![section-layout](https://user-images.githubusercontent.com/12703638/33338907-5edb7156-d42c-11e7-818b-ef8adc602ec4.png)

Question layout:
- place all form headings above the field/editor for consistency 
- reduce "Question number" box to half column width
- reduce "Answer format" dropdown to half column width
- reduce "Themes" dropdown to half column width
- reduce width of all TinyMCE editors to be same/standard for all areas of page (Phase, Section, Question)
![question-layout](https://user-images.githubusercontent.com/12703638/33339071-cbb01944-d42c-11e7-99c5-d09b1988fb2d.png)

Section/Question layout:
- align all fields in a single column

![section-question](https://user-images.githubusercontent.com/12703638/33340141-1dbb62fe-d430-11e7-95f8-a156499781b0.png)

Button alignment:
- align all buttons on lower right below entry fields
 
![button-alignment](https://user-images.githubusercontent.com/12703638/33340205-565199da-d430-11e7-8f89-8fe08900ee36.png)

Edit Question:
- when I click button to Edit Question, I'm taken to a screen w/a different layout from the Add Question page. these should be the same layout w/standardized field and editor sizes.
 
![edit-question](https://user-images.githubusercontent.com/12703638/33340324-b01c9802-d430-11e7-8e5e-852fbe3ce11d.png)

Edit Question Buttons:
- remove "Delete button" - should only have Save and Cancel buttons here

![edit-q-buttons](https://user-images.githubusercontent.com/12703638/33340455-f8a6e49c-d430-11e7-866e-2f54929f3217.png)

stephaniesimms:
standardized widths and button placements look much better @bhavi - just one change for now:
[...]

PR #886, opened 2017-12-01
added lists to plugins and set max container height

briri:
Tinymce references bulleted/numbered lists in its toolbar declaration but was not loading the plugin for them. Added the plugin to the JS config. #885

Also set a max height on the container in an attempt to correct an issue we have begun seeing where the editor box's height makes it span beyond the full page height until you click on it (it receives focus) at which point it resizes to the normal height #874

### DIFF
--- a/lib/assets/javascripts/utils/tinymce.js
+++ b/lib/assets/javascripts/utils/tinymce.js
@@ -4,6 +4,7 @@ import tinymce from 'tinymce/tinymce';
 import 'tinymce/themes/modern/theme';
 // Plugins
 import 'tinymce/plugins/table';
+import 'tinymce/plugins/lists';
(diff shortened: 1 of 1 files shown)


lib/assets/javascripts/utils/tinymce.js
```diff
@@ -36,6 +37,7 @@ export const defaultOptions = {
   paste_remove_styles_if_webkit: true,
   paste_remove_spans: true,
   paste_strip_class_attributes: 'all',
+  max_height: 350,
```
jollopre: Perhaps this is solving the problem but the root cause for seeing a huge editor that is resized automatically later on is due to using display:none instead of visibility:hidden to hide a tinymce editor.

Issue #546, opened 2017-07-27
Add "Plan Overview" tab to plan writing pages


stephaniesimms:
UX wireframe @JEK-III - http://h90c3m.axshare.com/#g=1&p=write_plan

This follows on related issue #536. We want to retain the overview summary of all questions in a template but move it from the current location at the bottom of plan/project details to its own tab. The plan overview content and functionality (i.e., show/hide at the phase level only) is substantially the same as the current DMPonline application. But replace the table layout with a structured list that (with appropriate HMTL) so it's easier to navigate with a screen-reader and attach labels to input fields.

Note blue notepad functional specs for multiphase plans (e.g. Horizon2020):
- Hide phase title and open/close control in single-phase plans; always show summary.

JEK-III:
@stephaniesimms 
Oops– that annotation came along for the ride when I copy-pasted the Write Plans tab. We definitely only want one Plan Overview tab containing all the phases.

stephaniesimms:
thanks for clarification @JEK-III - makes sense now.

JEK-III:
The wireframes on axshare have now been updated.

stephaniesimms:
just a few tweaks to request:
- The wireframe is designed for a complex H2020 template w/multiple phases. For templates that only have one phase, please remove the phase title and open/close accordion control and just display the full summary of questions. 

Formatting suggestions for the summary: 
- Number section titles and make them bold (in screenshot below 1. Data areas and data types, 2. Standards and metadata...)
- increase spacing between numbered sections
- anything else @sjDCC ? i know DMPonline has a table here but the accessibility test nixed tables bc screenreaders can't handle them.

![screen shot 2017-12-07 at 12 08 43 pm](https://user-images.githubusercontent.com/12703638/33736624-28a7d7e6-db48-11e7-9e34-e2e907c087db.png)

sjDCC:
Mostly sounds good but please don't number sections. 

We let template owners assign numbers in the question text as needed. This is because everyone uses a different system. Some funders specify numbers and letters (e.g. AHRC 2a, 2b,) while others like MRC start numbering at 0 and have sub-points (e.g. 4.1, 4.2), but most make things easy and just use bullets ;-)

If we apply numbers it gets really confusing so we intentionally leave them blank.

jollopre:
- Plan overview considers now plans with just one section.
- Margin bottom (e.g. 5px) added to each section in order to visually separate them.

[...]

PR #889, opened 2017-12-04
Issue 546

jollopre:
- plan overview controller action and view associated. DMPRoadmap/roadmap#546
- functional test for overview action. DMPRoadmap/roadmap#546

### DIFF
--- a/app/controllers/plans_controller.rb
+++ b/app/controllers/plans_controller.rb
@@ -3,7 +3,7 @@ class PlansController < ApplicationController
   require 'pp'
   helper PaginableHelper
   helper SettingsTemplateHelper
-  after_action :verify_authorized
+  after_action :verify_authorized, except: [:overview]
 
   def index
     authorize Plan
@@ -354,6 +354,17 @@ class PlansController < ApplicationController
      end
     rescue Exception
       redirect_to share_plan_path(plan), alert: alert
+    end
+  end
+
+  def overview
+    begin
+      plan = Plan.overview(params[:id])
+      authorize plan
+      render(:overview, locals: { plan: plan })
+    rescue ActiveRecord::RecordNotFound
+      flash[:alert] = _('There is no plan associated with id %{id}') %{ :id => params[:id] }
+      redirect_to(action: :index)
     end
   end
 
--- a/app/models/plan.rb
+++ b/app/models/plan.rb
@@ -70,6 +70,11 @@ class Plan < ActiveRecord::Base
       .where(['NOT EXISTS (SELECT 1 FROM roles WHERE plan_id = plans.id AND user_id = ?)', user.id])
       .order(:title => :asc)
   }
+
+  # Retrieves plan, template, org, phases, sections and questions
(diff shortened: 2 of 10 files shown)


app/controllers/plans_controller.rb
```diff
@@ -3,7 +3,7 @@ class PlansController < ApplicationController
   require 'pp'
   helper PaginableHelper
   helper SettingsTemplateHelper
-  after_action :verify_authorized
+  after_action :verify_authorized, except: [:overview]
```
briri: I think you may have forgotten to remove this exception. You're calling authorize in the method.

jollopre: This expect has been placed intentionally. I don't need pundit to verify that authorize has been called which will fail when trying to find a plan for a non-existing id, raising ActiveRecord::RecordNotFound and therefore never executing authorize which Pundit complains. The three possible test cases I can think of has been added at test/functional/plans_controller_test.rb

briri: by default, Rails displays the detailed error message only in development mode in these RecordNotFound scenarios. We have configured roadmap-stg to do the same to make it easier for us to UAT and record bugs. In production though Rails sends the user to a generic 404 page (see screenshot). Since most of our controllers do not have specific handling for RecordNotFound it would be better to do something more generic like update the default 404 page to fit within the styling of the site (we should do this anyway) or add something in the application_controller.

screen shot 2017-12-05 at 10 48 29 am
404 screen when running the app w/ `bundle exec rails s -e production`

bhavi: I think moving the Record Not found exception handling code out of the controller method into the application controller might be better and make it reusable too. For example using Rails "rescue_from" method,
https://stackoverflow.com/questions/13434523/general-rescue-throughout-controller-when-id-not-found-ror

briri: @jollopre if you could please remove the RecordNotFound check in the overview method and the pundit exception I'll create a separate ticket for this and add it to the bugfix column. After having seen the above version of what the default Rails 404 page is doing I think we should add it as an MVP task. Would be better to have a themed 404 page than the one above.

jollopre: Rails in production is automatically rescuing from ActiveRecord::RecordNotFound and instead of throwing 5XX status code (which means something not available now, basically), it changes to be 404 (which means I don't have that resource). By removing that line, you have a misleading response since and that's the reason I explicitly recover from that to tell the user that plan didn't exist. I think we should keep this like it is until we don't have a proper handler in place acting at the application_controller.rb level and rescuing from there any Expection which is appropriated message.

jollopre:
merge conflicts resolved.

app/controllers/plans_controller.rb
briri: ok. will accept this as is but please do not introduce more of these RecordNotFound rescues in future PRs. It just adds to the amount of work that will need to be done when addressing #899

Issue #845, opened 2017-11-21
Add visibility to Template details page for Funders


briri:
This issue was uncovered while working with migrated DMPTool data. 

Some funders like the USGS have normal funder templates but they also have users who create DMPs and have templates for internal use. Right now the public facing 'DMP Templates' page displays any published funder template. We need a way for a funder's Org Admin to identify whether the template is for users applying for a grant with their organisation or if the template is for their internal users. 

The templates table has a `visibility` (INT) column that is no longer in use after we converted the codebase over to publishable versions. We can use this column to identify these internal templates. By default we would set the value to 'zero' (organisational visibility). Funder Org Admin's could then change this value via a checkbox to 'one' (public visibility). (We should run a rake task that sets any existing funder templates to 'one' automatically).

The public facing 'DMP Templates' page should then be updated to only collect the funder templates whose `visibility = 1`. The create plan page would also need to be updated to consider this value in the ajax endpoint in the `templates_controller` that collects available templates for the research org and funder combo (research org templates `visibility = 0` and funder templates `visibility = 1`. The Org Admin templates page would also need to be updated to check for `visibility = 1` when generating the list of customizable templates for non-funder orgs.

sjDCC:
So if I'm understanding the use case right @briri, the logic that we'd need to apply on the create plan page is that:

1) If user is affiliated with funder and selects funder from the organisation dropdown, present the internal funder template

2) If the user is affiliated with funder and selects funder from the organisation dropdown AND funder dropdown, present the internal funder template, not public one? Currently the logic does the opposite as funder templates take precedence over institutional ones so we'd need an exception here for when the org affiliation is a funder

3) If a user affiliated with a non-funder selects a funder in the org dropdown, present the funder guidance not internal funder template as these templates are just for internal funder use?

I assume this is related to ticket #665 to ensure funders display in the org dropdown

stephaniesimms:
@briri's assumptions above are correct. we tested on DMPonline and existing logic/workflow should meet the need w/#665 . there's no need to keep USGS org templates private. the use case relates to the difference between intramural and extramural research at US federal agencies. a lot of researchers are employed directly by the agencies and have internal requirements for data mgmt (Org Templates) but don't need to apply for funding (Funder Template meant for external researchers).

sjDCC:
Ok great, even better if no logic needs to be changed in the 'create plan' workflow.

[...]

PR #892, opened 2017-12-04
Added ability to set template visibility for funders

briri:
per issue #845
- enabled use of existing templates.visibility field
- updated edit/show template details page to show checkbox 
- updated public DMP Templates page to look at visibility when generating list
- added rake task to automatically set funder templates and the default template to publicly_visible and all others to organisationally_visible

### DIFF
--- a/app/controllers/templates_controller.rb
+++ b/app/controllers/templates_controller.rb
@@ -274,6 +274,14 @@ class TemplatesController < ApplicationController
 
       @template.description = params["template-desc"]
       @template.links = JSON.parse(params["template-links"]) if params["template-links"].present?
+
+      # If the visibility checkbox is not checked and the user's org is a funder set the visibility to public
+      # otherwise default it to organisationally_visible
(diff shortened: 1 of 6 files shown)


app/controllers/templates_controller.rb
```diff
@@ -274,6 +274,16 @@ def admin_update
...
       @template.links = JSON.parse(params["template-links"]) if params["template-links"].present?
+
+      if current_user.org.funder?
+        if params[:template_visibility].nil?
+          @template.visibility = Template.visibilities[:publicly_visible]
+        else
+          @template.visibility = Template.visibilities[:organisationally_visible]
+        end
+      else
+        @template.visibility = Template.visibilities[:organisationally_visible]
```
jollopre: This else seems redundant to me. If I understood well, only funders are able to change the visibility of a template.

briri: It is preventing the value from being nil. Will refactor though to make the code less verbose

Issue #893, opened 2017-12-05
Default answers being saved on plan creation


sjDCC:
If default answers exist (see for example NERC full plan which has tables to complete), these are autosaved on plan creation so a brand new plan shows as being partially complete

![default-answer](https://user-images.githubusercontent.com/6169940/33599045-c6a55a48-d99c-11e7-9830-e560d407925b.PNG)

stephaniesimms:
tested and now I have to click Save button to save default answers. can you confirm that this is the desired behavior @sjDCC? if so this issue can be closed.

jollopre:
@stephaniesimms, if we auto-save a question with the default text, the questions types that are just text (e.g. textfield, textarea) will count in the section and progress statuses and I think that was the reason @sjDCC raised the issue.

We could exclude questions that have default answer from the counts but I think that would be more confusing.

stephaniesimms:
@jollopre I agree w/all you mentioned above and think the issue should be resolved now. I just wanted @sjDCC to confirm that she also agrees before closing the issue.

sjDCC:
This is all working fine now. Previously it was saving on creation. Now if a question is answered in another section and autosaves, it doesn't save the default answers too so all good. Removing from board

PR #896, opened 2017-12-05
Issue 893

jollopre:
- removed after_create callback for answers since no longer is needed after introducing transactions. DMPRoadmap/roadmap#893
- unit test for answer#is_valid?. DMPRoadmap/roadmap#893
- answer#update based on current_user. input text for answer based on either answer.text or question.default_text. DMPRoadmap/roadmap#893
- renamed answers#update to answers#create_or_update and added new tests. DMPRoadmap/roadmap#893

### DIFF
--- a/app/controllers/answers_controller.rb
+++ b/app/controllers/answers_controller.rb
@@ -1,20 +1,35 @@
 class AnswersController < ApplicationController
-  after_action :verify_authorized
   respond_to :html
 
-	# PUT/PATCH /answers/[:id]
-  def update
+	# POST /answers/create_or_update
+  def create_or_update
     p_params = permitted_params()
+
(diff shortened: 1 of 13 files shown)


jollopre:
@briri, the first rescue introduced is needed. We could debate the rescue from other places when it is involved just one condition (e.g. using find). But in this case, we can't leave it like that. If that's not checked, we could create an answer for a question that does not really belong to the plan passed. If you look at line https://github.com/DMPRoadmap/roadmap/blob/CDL-MVP/app/controllers/answers_controller.rb#L10, when find an answer for plan, question fails, we create answer in the rescue statement. 

We could also do the check in that rescue clause from the line I pointed, but it seems to me cleaner to check that first.

Issue #890, opened 2017-12-04
Clicking the 'Cancel' button when editing a section's details deletes the section


briri:
Please complete the following fields as applicable:

**Expected behaviour:**
Clicking the 'Cancel' button on the 'edit section details' view ends up deleting the entire section.

**Actual behaviour:**
Up for discussion (@sjDCC and @stephaniesimms). When you expand the section you are in edit mode by default. Should we simply remove the cancel button or have it collapse the section?

**Steps to reproduce:**
Log in as Org/Super Admin. Go to the Admin -> Templates page. Select one of your org's templates. Click on the phase tab. Expand one of the sections. Click 'Cancel'

sjDCC:
Realise I've responded about questions not sections, but same principle applies.

I also wonder if we should have a popup when user clicks delete to check they actually want to delete the whole section or question

stephaniesimms:
yes we appear to be missing a button. there should be 3: Save, Delete, Cancel when you click to edit a Section or Question. 

and yes we should add a modal to ask a user if they are sure they want to Delete something (same as the modal for confirming actions to Remove a plan or template).

stephaniesimms:
all the buttons are there now and work as expected, but please change the order @briri to be consistent w/ DMPonline:

Save - Delete - Cancel (flip the last two, screenshot below)

![screen shot 2017-12-11 at 3 37 55 pm](https://user-images.githubusercontent.com/12703638/33859557-53dd46e6-de89-11e7-9d7d-ef8d3acca472.png)

stephaniesimms:
button order is good now. but no confirmation message asking me if I'm sure I want to delete this question. please add the confirmation step @briri

sjDCC:
Hi @briri  - I just managed to delete two sections without the modal being issued to confirm selection. See: https://roadmap-stg.cdlib.org/org/admin/templates/phases/5492/admin_show

Perhaps this is because it was an existing legacy template?

briri:
Hey @sjDCC I get the confirmation message for existing sections and new ones I've created. I tried it as several different orgs. 
What org owns the ticket you're having trouble with?

sjDCC:
Hi @briri I was editing one called "Copy of UoE DMP" owned by the University of Edinburgh. I've gone itnto edit that again, and other templates, and the modal is displaying. Perhaps it was a browser issue yesterday or something else in the setup? I'm at home on laptop now.

All seems to be working though so suggest closing out and putting this down to some quirk.

Issue #813, opened 2017-11-13
Create Super Admin Templates page


stephaniesimms:
UX wireframe @rhu001 http://h90c3m.axshare.com/#g=1&p=dmp_all_templates__super_admin_view___org_template

Create "Templates" page in Super Admin dropdown menu. This is where Super Admins can manage funder and default templates (first tab) and Organisation templates (second tab). 
- add Table for each tab w/layout and content in screenshots below
- functional spec: Template names are links that take the user to the template editing page
- add Search box above table and pagination if either table exceeds 10 items

Templates page in Super Admin dropdown menu
![screen shot 2017-11-13 at 11 27 03 am](https://user-images.githubusercontent.com/12703638/32744818-9fbc08fc-c865-11e7-98bc-a5c717b3300e.png)

Funder & Default Template (rename heading)
![screen shot 2017-11-13 at 11 21 27 am](https://user-images.githubusercontent.com/12703638/32744663-23fc06fe-c865-11e7-9918-a3ca49a24e85.png)

Organisation Templates
![screen shot 2017-11-13 at 11 25 10 am](https://user-images.githubusercontent.com/12703638/32744750-6e4e7200-c865-11e7-862f-0a6ba74cc5c8.png)

sjDCC:
Tabbed approach looks fine to me @stephaniesimms We could apply this to the general org-admin templates page too for consistency but I don't think that's essential for MVP

dsisu:
@stephaniesimms Please, let's focus on features essential only for MVP. Also, whenever we plan to add a feature it's important that we think of the use case for it, we don't just add things because they can be done. Once we have a use case, we present it to the developers and let them think about how it fits into the process.

stephaniesimms:
thanks for the confirmation @sjDCC - as we discussed last week we need to prioritize some template/guidance dev work so that this functionality is available for manual data curation. we're beginning on 4 Dec to audit all US funder templates and hired a contractor to help as it's a huge task.

briri:
This is an ajax call that could have been impacted by the https -> http confusion with the server configuration. Jim updated our puppet script. I will test today to see if the issues have been resolved

stephaniesimms:
this looks good on roadmap stage and is working as expected. closing issue but will leave it on the board for @sjDCC to confirm.

Issue #900, opened 2017-12-05
Template table revisions


stephaniesimms:
For Org Admins, tabs on the Templates page should read:
- Own Templates
- Customizable Templates - note that this table includes the default template (mark the template name w/an asterisk)
- Org Admins should not be presented w/the box to Change affiliation. They can perform all Actions on their own templates, but can only customize funder templates.

For Super Admins, tabs on the Templates page should read:
- Organisation Templates
- Funder Templates - note that this table includes the default template (mark the template name w/an asterisk)
- Add "Customise" to Actions menu for Funder Templates and enable Super Admins to add customisations to these as well as edit them (DMPTuuli use case)

Global adjustments for both Org and Super Admin views of templates page:
- Add sort functionality to Funder templates table
- Add total template counts above each table as in the DMPTool (screenshot below). Provide the total n for: "All" templates, "Published" templates, and "Unpublished" templates in each table.

![screen shot 2017-12-05 at 11 48 34 am](https://user-images.githubusercontent.com/12703638/33627417-4db649a6-d9b2-11e7-8ad4-40c06930b89e.png)

briri:
Moving 'Add total template counts above each table as in the DMPTool (screenshot below). Provide the total n for: "All" templates, "Published" templates, and "Unpublished" templates in each table.' to a separate ticket as it requires #815 to be complete

briri:
Hey @sjDCC and @stephaniesimms,
I implemented the changes requested above, but the logic behind it is getting quite complicated and fragile (e.g. show these links when the user is super admin and the org is not a funder ... etc.). 
So I decided to experiment with a suggestion @xsrust made when we originally started talking about this consolidated template admin page. The behavior in this version is identical for the 'Own templates' and 'Customizable templates' tables for both super admin and org admins. The super admin though now has an additional tab called 'All templates' (Sam's suggestion) which always shows all the templates regardless of who they are affiliated with. For the super admin I am showing 'Org name's templates' instead of 'Own templates' to help the user understand who they are editing as.
Let me know what you think of the screenshots below they show what the page would look like for an Org Admin and then a Super Admin affiliated with the same Org:

![screen shot 2017-12-22 at 11 15 47 am](https://user-images.githubusercontent.com/1204467/34310154-9c528ba0-e70a-11e7-8ef7-ed2270daa0e7.png)
Logged in as an Org Admin and viewing my own templates

[...]

PR #902, opened 2017-12-05
Make templates page usable for Org Admins and Super Admins

briri:
issue #813 
- Updated the existing Org Admin templates page
  - Switched to tabbed view
  - Added org affiliation box for Super Admin only
  - Updated logic to display appropriate Edit/Publish/etc. links when the user is a Super Admin 
  - Added JS logic to hide/show the Edit/Publish/etc. links based on the Super Admin's current affiliation
- Moved existing controllers/templates_controller.rb to controllers/org_admin/templates_controller.rb
- Moved views from views/templates to views/org_admin/templates
- Updated Routes and Policies accordingly
- Updated tests    

Also (based on feedback from the CDL template editing session):
- Added a 'View all templates' button to the top of the template editing pages to help the user get back to the all templates page #908
- Renamed the current 'Cancel' button on the edit section area to 'Delete' and added a true cancel button #890


app/policies/template_policy.rb
```diff
@@ -1,76 +1,81 @@
...
   end
 
-  def admin_template?
-    user.can_modify_templates?  &&  (template.org_id == user.org_id)
+  def new?
+    user.can_modify_templates?
   end
 
-  def admin_customize?
+  def create?
```
xsrust: super-admins without modify_template privleges will be unable to create templates, consider changing line below to:
``` user.can_modify_templates? || user.can_super_admin ```

File: app/policies/template_policy.rb [hunk 2]
```diff
@@ -1,76 +1,81 @@
...
-  def admin_new?
-    user.can_modify_templates?
+  
+  def history?
+    user.can_super_admin? || (user.can_modify_templates? && template.org_id == user.org_id)
   end
 
-  def admin_create?
-    user.can_modify_templates? && (template.org_id.nil? || (template.org_id == user.org_id))
+  def customize?
```
xsrust: super_admins cannot modify templates with current auth system. 

Dont think these display for them on the front end either?? so this may not be an issue

app/policies/template_policy.rb
```diff
@@ -1,76 +1,81 @@
...
 
-  def admin_copy?
+  # AJAX Calls
+  def copy?
+    user.can_super_admin? || (user.can_modify_templates?  &&  (template.org_id == user.org_id))
+  end
+  def publish?
+    user.can_super_admin? || (user.can_modify_templates?  &&  (template.org_id == user.org_id))
+  end
+  def unpublish?
```
xsrust: is it desired for super-admins not to be able to unpublish?

app/views/org_admin/templates/_templates_list.html.erb
```diff
@@ -0,0 +1,66 @@
...
+    <% scope.each do |template| %>
+      <tr>
+        <td>
+          <%= template.title %>
+        </td>
+          <td>
+            <%= current_user.can_super_admin? ? template.org.name : raw(template.description) %>
+          </td>
+          <td>
+            <%= (template.published? ? _('Published') : _('Unpublished')) %>
```
xsrust: Published status shouldn't be binary.
This explains the bugs @stephaniesimms highlighted in the meeting yesterday with templates appearing unpublished once you edit them.
Have a look at the logic that was in place before which correctly displays the status:
```
<% if hash[:live].nil? %>
    <%= _('Unpublished') %>
<% elsif hash[:current].dirty? %>
    <%= _('Unpublished changes') %>
<% else %>
    <%= _('Published') %>
<% end %>
```

app/views/org_admin/templates/_funder_templates_list.html.erb
```diff
@@ -0,0 +1,106 @@
...
+        <th><%= _('Template Name') %></th>
+        <th><%= _('Funder') %></th>
+        <th><%= _('Status') %></th>
+        <th><%= _('Edited Date') %></th>
+        <th class="sorter-false"></th>
+      </tr>
+    </thead>
+    <tbody>
+    <% scope.each do |template| %>
+      <% customization = customizations.select{ |t| t.customization_of == template.dmptemplate_id }.first %>
```
xsrust: Storing customization as a hash with keys being the dmptemplate_id would allow for faster lookup here:)

app/views/org_admin/templates/_funder_templates_list.html.erb
```diff
@@ -0,0 +1,106 @@
...
+                <%= _('Not Customised') %>
+              <% end %>
+            <% end %>
+          </td>
+          <td>
+            <% last_temp_updated = template.updated_at %>
+            <%= l last_temp_updated.to_date, formats: :short %>
+          </td>
+          <td>
+            <div class="dropdown<%= (!current_user.can_super_admin? || (current_user.can_super_admin? && current_org == template.org.id)) ? '' : ' hide'%>">
```
xsrust: Do we refresh the page when the user switches organization?
If so I propose not-rendering the drop downs instead of rendering them as hidden unless we un-hide them from the JS somewhere

app/views/org_admin/templates/_templates_list.html.erb
```diff
@@ -0,0 +1,66 @@
...
+          </td>
+          <td>
+            <% last_temp_updated = template.updated_at %>
+            <%= l last_temp_updated.to_date, formats: :short %>
+          </td>
+          <td>
+            <div class="dropdown<%= current_org == template.org.id ? '' : ' hide'%>">
```
xsrust: Unless we are un-hiding this from the JS, I propose we just don't render a dropdown menu for un-editable templates.

db/schema.rb
```diff
@@ -86,349 +90,391 @@
...
   add_index "users", ["org_id"], name: "index_users_on_org_id", using: :btree
 
   create_table "users_perms", id: false, force: :cascade do |t|
-    t.integer "user_id"
-    t.integer "perm_id"
+    t.integer "user_id", limit: 4
+    t.integer "perm_id", limit: 4
   end
 
+  add_index "users_perms", ["perm_id"], name: "fk_rails_457217c31c", using: :btree
```
xsrust: Not sure I see a use case for this index.  was this added by this piece of code?
I dont think we ever select from this table users based on a perm_id

app/controllers/org_admin/templates_controller.rb
```diff
@@ -0,0 +1,450 @@
...
+      end
+    end
+    
+    # DELETE /org_admin/templates/:id
+    # -----------------------------------------------------
+    def destroy
+      @template = Template.find(params[:id])
+      authorize @template
+
+      if @template.plans.length <= 0
```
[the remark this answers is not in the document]
briri: Yes that's a good observation and one I'll add to the agenda for next week. Should we remove the single template ... sort of like rolling back a version (assuming there is a prior version) or delete them all. It currently just deletes the current version.

app/controllers/org_admin/templates_controller.rb
```diff
@@ -0,0 +1,450 @@
...
+    # -----------------------------------------------------
+    def index
+      authorize Template
+      valid_orgs = current_user.can_super_admin? ? Org.non_funders : Org.find(current_user.org)
+      org_templates = get_latest_template_versions(valid_orgs).page(1)
+      funder_templates = get_latest_template_versions(Org.funders).page(1)
+
+      # If the user is an Org Admin look for customizations to funder templates
+      customizations = []
+      unless current_user.can_super_admin?
```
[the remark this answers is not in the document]
briri: Not at this point. I am going to open a separate ticket to revisit and add the additional column that contains the customization controls/status. I'd like to get this merged asap since the template objects have been moved to org_admin namespace.

app/models/org.rb
```diff
@@ -61,6 +61,7 @@ class Org < ActiveRecord::Base
   scope :managing_orgs, -> { where(abbreviation: Rails.configuration.branding[:organisation][:abbreviation]) }
   scope :funders, -> { where(org_type: 2) }
   scope :institutions, -> { where(org_type: 1) }
+  scope :non_funders, -> { where.not(org_type: 2) }
```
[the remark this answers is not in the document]
briri: yes agreed but to be honest there's no point to using a bitflag here, should have been an enum since we only allow one org_type per org. Will create a ticket to refactor after MVP

app/policies/template_policy.rb
briri: good catch

app/policies/template_policy.rb
briri: yep. will fix this

app/views/org_admin/templates/_funder_templates_list.html.erb
briri: yes. will update that

app/views/org_admin/templates/_funder_templates_list.html.erb
briri: no, its an ajax call with some JS to handle which Orgs' publish/edit/etc. links are visible

app/views/org_admin/templates/_templates_list.html.erb
briri: Ah right. will address this and move that ticket over to review

app/controllers/org_admin/templates_controller.rb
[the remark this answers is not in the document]
xsrust: ok, that sounds sensible to handle in another ticket.

app/controllers/org_admin/templates_controller.rb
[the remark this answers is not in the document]
briri: if its going to delete all the versions we'll need to do that check for plans. I think that could get very messy!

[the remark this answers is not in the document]
xsrust: That sounds like good fodder for the Dev call on Friday

app/models/org.rb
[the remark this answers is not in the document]
xsrust: Well, we're currently using bitflags so I think it makes sense to stick to the conventions.
Additionally, it sounds like your funder/institution organization (GIS?) is a good case for bitflags over enums as we may need combinations of the current options

app/policies/template_policy.rb
briri: updating all of these policies to include the `.can_super_admin?`.

briri:
Alright Sam. Made all of the suggested changes except for adding the 'Customize column' which is part of the new #900 ticket
Please review when you have a chance.

app/policies/template_policy.rb
xsrust: Concern covered by another ticket

app/views/org_admin/templates/_funder_templates_list.html.erb
xsrust: Ok, this comment is invalid then.

app/views/org_admin/templates/_templates_list.html.erb
xsrust: invalidated by another comment

app/models/template.rb
```diff
@@ -80,6 +80,15 @@ def self.org_customizations(dmptemplate_id, org_id)
     Template.where(customization_of: dmptemplate_id, org_id: org_id).order(version: :desc).valid.first
   end
 
+  def self.get_latest_template_versions(orgs)
+    # get the unique dmptemplate_ids (template version families)
+    families = Template.valid.where(org: orgs, customization_of: nil).pluck(:dmptemplate_id)
+    # return the most current versions of each template version family
+    ids = []
+    families.map{ |fam| ids << Template.current(fam).id }
```
xsrust: Thanks for moving this but it still has scaleability concerns
The map used generates a large object in memory and then throws it out the object, never using it.
One query is run per template (template.current(fam)), which is an n+1 query issue.

I again propose this gets re-factored to the following code:

```id_query = 'select id from (select id, version , max(version) over (partition by dmptemplate_id) as v from templates order by dmptemplate_id) as foo where version = v;'
ids = []
Template.find_by_sql(id_query).each do |templ|
  ids << templ.id
end
Template.includes(:org).where(id: ids).order('orgs.name, templates.title')
```
This will run 2 queries and find all the templates instead of 2+n

briri: Thanks for the example @xsrust. That works for Postgres but MySQL doesn't seem to support OVER. I refactored a bit and this gets us the same desired result (one query to collect the unique dmptemplate_ids and the other to grab the current version for each dmptemplate_id)
```sql
def self.get_latest_template_versions(orgs)
    # get the unique dmptemplate_ids (template version families)
    families = Template.valid.where(org: orgs, customization_of: nil).pluck(:dmptemplate_id)

    # return the most current versions of each template version family
    id_query = id_query = "SELECT templates.dmptemplate_id, templates.id, current.version "\
               "FROM "\
               "(SELECT MAX(version) AS version, dmptemplate_id "\
                "FROM templates WHERE dmptemplate_id in (#{families.join(', ')}) "\
                "GROUP BY dmptemplate_id) as current "\
               "INNER JOIN templates ON current.version = templates.version AND current.dmptemplate_id = templates.dmptemplate_id;"

    ids = Template.find_by_sql(id_query).collect{|t| t.id }
    Template.includes(:org).where(id: ids).order('orgs.name, templates.title')
  end
```

xsrust: Ahh, thanks.  I didnt realise MySQL didnt support OVER.  
@jollopre suggested another alternative where the subqueries are re-factored into scopes so we can re-use each sub-query in other places, detailed in https://github.com/DMPRoadmap/roadmap/issues/928, but I'm happy to merge this

xsrust:
Merging this as changes made and potential future refactoring detailed in #928

Issue #904, opened 2017-12-05
Fix double redirect error for links on Org details page


stephaniesimms:
Expected behavior: If I enter a URL on the Org details page and do not provide link text (or vice versa), I should receive an error message when I click Save that prompts me to enter something in the empty field.

Current behavior: If I click Save when either field is blank I receive the error message below.

![dbl-redirect](https://user-images.githubusercontent.com/12703638/33633959-70378d54-d9c7-11e7-8f4b-f66af190f967.png)

![screen shot 2017-12-05 at 2 16 45 pm](https://user-images.githubusercontent.com/12703638/33633964-7462b656-d9c7-11e7-885b-1a29d865db29.png)

briri:
This will be fixed by #866.

stephaniesimms:
working as expected now. please just revise the error message for clarity (screenshot below). 
new message should read: "Error: Unable to save your changes. Please provide a valid URL and Link text to display in the banner."

![screen shot 2017-12-07 at 12 37 44 pm](https://user-images.githubusercontent.com/12703638/33737680-e4f58026-db4b-11e7-99fe-cd8cfa32aed3.png)


![screen shot 2017-12-07 at 12 37 44 pm](https://user-images.githubusercontent.com/12703638/33737571-801cb26e-db4b-11e7-87d9-45646ee70833.png)

stephaniesimms:
@briri - now when I input only a URL or only link text it doesn't save, but I receive a Save Successful message (screenshot 1). I should instead receive and error message: "Error: Unable to save your changes. Please provide a valid URL and Link text to display in the banner."

I should also receive this error message if I reverse the URL and link text accidentally. Current behavior allows me to save the reversed input and displays the URL in the banner (see screenshot 2).

1 - this should be an error msg
![no-success](https://user-images.githubusercontent.com/12703638/33850896-2f073c10-de6a-11e7-86cc-459ce843b0f3.png)

2 - this should also produce an error msg
![reverse](https://user-images.githubusercontent.com/12703638/33850885-2b9e1972-de6a-11e7-8bab-732c8387660b.png)

sjDCC:
This seems to all work now so closing out - thanks @briri 

There's one related bug - a proper warning is not generated when the 'Link text' entry is too long - but I've raised this in ticket #729

Issue #895, opened 2017-12-05
Duplicate email on DMP feedback request


sjDCC:
When users request feedback on plans, all existing org-admins and the specified contact address under 'organisational details' are emailed alerts. If the single org email specified is the same as an administrator contact address, this person receives the email twice.

Some unis may not have a generic helpdesk to specify under org details, so we should make sure emails aren't duplicated if a personal address is used.

stephaniesimms:
working as expected now. closing issue.

Issue #866, opened 2017-11-27
Convert links on Org Details page to use logic introduced in PR #863


briri:
Update the Org details page and JS to make use of the new generic JSON link array handler introduced in PR #863

jollopre:
orgs_controller.rb can execute multiple times redirect_to which leads to failing request. I'm assuming once the checking of link objects is using the concern will go away but for now. Consider that, passing a link object with empty link or text will execute one redirect and after will execute the begin rescue end clause too https://github.com/DMPRoadmap/roadmap/blob/CDL-MVP/app/controllers/orgs_controller.rb#L38

briri:
Also add a migration to change orgs.links back to a VARCHAR.

stephaniesimms:
can't test this in UI. closing issue.

PR #909, opened 2017-12-05
fixes org details redirect issue and duplicate email on feedback request

briri:
Addresses redirect error that was preventing Org Details page from saving #904. This is a temporary fix. The links are still not saving. I will address that issue tomorrow when converting to use the same logic that the funder links and sample plans use in #866 

Also addresses the duplicate email issue that happens when the Org's contact email is the same as an org admins. #895

### DIFF
--- a/app/controllers/orgs_controller.rb
+++ b/app/controllers/orgs_controller.rb
@@ -17,7 +17,7 @@ class OrgsController < ApplicationController
     @org = Org.find(params[:id])
     authorize @org
     @org.logo = attrs[:logo] if attrs[:logo]
-    
+    failure = ''
     tab = (attrs[:feedback_enabled].present? ? 'feedback' : 'profile')
     
     if attrs[:links].present?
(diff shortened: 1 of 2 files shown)


app/controllers/orgs_controller.rb
```diff
@@ -27,24 +27,20 @@ def admin_update
         if json.all?{ |o| o['link'].present? && o['text'].present? }
           @org.links = json
         else
-          redirect_to "#{admin_edit_org_path(@org)}\##{tab}", alert: _('Unable to save your changes. Invalid URLs.')
+          failure = _('Unable to save your changes. Invalid URLs.')
```
jollopre: It if was a temporary fix, return after redirect_to would be enough

app/models/plan.rb
```diff
@@ -213,7 +213,7 @@ def request_feedback(user)
             UserMailer.feedback_confirmation(r, self, user).deliver_now
           end
           # Send an email to all of the org admins as well as the Org's administrator email
-          if user.org.contact_email.present?
+          if user.org.contact_email.present? && !admins.collect{ |u| u.email }.include?(user.org.contact_email)
```
jollopre: !admins.collect(&:mail) is leaner for this case

Issue #871, opened 2017-11-27
Admin Guidance page - revise publish action


stephaniesimms:
This is a stub issue - TBD once the Templates page is revised
Need to rethink the workflow for publishing Guidance
- add "Publish/Unpublish" to the Actions menus for Guidance group list and Guidance list (like on Templates page)
- change "Delete" to "Remove" in Actions menus
- Change "Published" column in Guidance group table to "Status" - status should be "Published" or "Unpublished" (same as Templates table)
- Add "Status" column to Guidance list table - place between "Guidance group" and "Last updated" columns 
- a guidance group should auto-publish once 1 piece of guidance in that group is published

![screen shot 2017-12-01 at 9 04 10 am](https://user-images.githubusercontent.com/12703638/33494386-f31e0cfa-d677-11e7-8926-3b87a027d45d.png)

![screen shot 2017-12-01 at 9 05 25 am](https://user-images.githubusercontent.com/12703638/33494585-8d3ace7c-d678-11e7-8071-c0010fb99d0c.png)

sjDCC:
I think we definitely want to be able to publish individual pieces of guidance as there may be one in question / needing revision and the institution won't want to have to deactivate the whole group to temporarily unpubliah this

I think the guidance group should autopublish once 1 piece of guidance is published though

stephaniesimms:
ok - guidance publishing logic sounds good to me @sjDCC

stephaniesimms:
@bhavi - auto-publish works as expected when I select to "Publish" a piece of Guidance from the Actions menu. but if I tick the "Published" box on the main Guidance editing page, it does not automatically publish the associated Guidance group (see screenshot 1). desired behavior is to auto-publish the Guidance group whether I publish a piece of Guidance via the main page or the actions menu.

one more thing i noticed when testing: the "Published" tickbox for a Guidance group is not visible when I create a new group. it only appears when I click to go Edit the group from the actions menu (screenshot 2). desired behavior is to provide the Published tickbox on this page at all times.

![guidance-publish-logic](https://user-images.githubusercontent.com/12703638/33865884-9351fc3a-dea8-11e7-97d6-cf71f96f7f4f.png)

![gg-publish](https://user-images.githubusercontent.com/12703638/33865999-2414c842-dea9-11e7-9944-39b5687bdf2a.png)

briri:
[...]

PR #914, opened 2017-12-06
Revise Publish action Admin Guidance page (#871)

bhavi:


briri:
looks good @bhavi thanks for including tests for these new controller actions

PR #915, opened 2017-12-06
Org Admins/Super Admins can view plans

briri:
addresses change request on #434 
- Made plan title on Admin -> Plans page a link to the plan details page
- Modified `plan.readable_by?` so that a super admin, org admin who belongs to the owner or one of the co-owners orgs can read the plan
- Added a new `plan.commentable_by?` method and updated the notes_controller and views to use it
- Removed Role.active checks from plans_policy and placed it in the Plan.has_role method

### DIFF
--- a/app/models/plan.rb
+++ b/app/models/plan.rb
@@ -94,10 +94,10 @@ class Plan < ActiveRecord::Base
   # returns the template for this plan, or generates an empty template and returns that
   #
   # @return [Dmptemplate] the template associated with this plan
-	def dmptemplate
-		#self.project.try(:dmptemplate) || Dmptemplate.new
-		self.template
-	end
+  def dmptemplate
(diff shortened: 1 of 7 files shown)


briri:
failing tests for this have been fixed in PR #916. If you merge that one I'll rebase this and resubmit

app/models/plan.rb
```diff
@@ -343,32 +343,46 @@ def add_guidance_to_array(guidance_array, guidance_group, theme, guidance)
...
-	end
+  end
+
+  ##
+  # determines if the plan is readable by the specified user
+  #
+  # @param user_id [Integer] the id for a user
+  # @return [Boolean] true if the user can read the plan
+  def readable_by?(user_id)
+    user = user_id.is_a?(User) ? user_id : User.find(user_id)
```
xsrust: I like this check for flexibilty between passing a user_id or a User object

Issue #880, opened 2017-12-01
Display funder requirements and sample plan links on DMP Templates page


sjDCC:
We've added the ability for funders to include links to their requirements and sample plans (see #819) but this need to be displayed on the public DMP Templates page. This is how it looks on DMPTool and staging.
- move the 'Download' column to immediately after the 'Template Name' and add the funder requirements links and sample plans columns at the end (final two columns on rights as per DMPTool)

![32756578-b0fef05a-c88f-11e7-8c62-61f3fc50131d](https://user-images.githubusercontent.com/6169940/33481588-222a6c9a-d68d-11e7-8af9-337cce5f7665.png)

![stage](https://user-images.githubusercontent.com/6169940/33481604-383d5e48-d68d-11e7-8e9a-e3706feb5a36.PNG)

briri:
Ok added the funder links and sample plans to the public DMP Templates page. 
Screenshot shows 2 funder links and 1 sample plan. The first funder link has no 'text' value
![screen shot 2017-12-06 at 3 08 14 pm](https://user-images.githubusercontent.com/1204467/33690197-67abfa96-da97-11e7-96d5-35d1acfa4fe6.png)

stephaniesimms:
@briri - looks great. but can you also add a tooltip like we have for DMPTool (screenshot below); it's not necessary to add the parenthetical (if available) just the tooltip. i think it's important to clarify the source of the sample plans.

Tooltip text: "Sample plans are provided by a funder, an institution or a trusted party."

![screen shot 2017-12-11 at 11 40 57 am](https://user-images.githubusercontent.com/12703638/33850313-54065db8-de68-11e7-8996-138e2d8e8309.png)

stephaniesimms:
this looks good @briri - closing issue. just a note here that it needs to be wired up w/#903 to populate the links in this table. assuming i should create a new issue for it.

PR #916, opened 2017-12-06
added funder links and sample plans to public templates page

briri:
- Added funder links and sample plans to the public DMPTemplates page #880 
- Updated `templates.links` field to be TEXT instead of VARCHAR (was not large enough to store more than one of each link type)
- Added default to `models/template.rb` since a BLOB/TEXT field is not able to have a default value in MySQL

### DIFF
--- a/app/helpers/template_helper.rb
+++ b/app/helpers/template_helper.rb
@@ -1,8 +1,8 @@
 module TemplateHelper
-  def links_to_a_elements(links)
+  def links_to_a_elements(links, separator = ', ')
     a = links.map do |l|
       "<a href=\"#{l['link']}\">#{l['text']}</a>"
     end
-    a.join(", ")
+    a.join(separator)
   end
 end
--- a/app/views/public_pages/template_index.html.erb
+++ b/app/views/public_pages/template_index.html.erb
@@ -18,7 +18,7 @@
         <thead>
          <% if @templates.count > TABLE_FILTER_MIN_ROWS %>
             <tr>
-              <th colspan="3" class="sorter-false">
+              <th colspan="5" class="sorter-false">
                 <%= render(partial: "shared/table_filter",
                            locals: {path: public_templates_path, placeholder: _('Filter templates')}) %>
               </th>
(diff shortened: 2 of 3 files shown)


db/migrate/20171206230107_convert_template_links_to_text.rb
```diff
@@ -0,0 +1,10 @@
+class ConvertTemplateLinksToText < ActiveRecord::Migration
+  def up
+    # BLOB/TEXT fields cannot have a default value - changing to set default in app/models/template.rb
+    change_column :templates, :links, :text, default: nil
```
jollopre: I don't see a case right now to change the type from VARCHAR to TEXT. As far as I'm aware both postgresl and mysql can have a variable length VARCHAR which should satisfy the requirements to store the links object as string.

app/views/public_pages/template_index.html.erb
```diff
@@ -18,27 +18,48 @@
...
           <% @templates.each do |template| %>
             <tr>
               <td><%= template.title %></td>
-              <td><%= template.org.name %></td>
               <td class="text-center">
                 <%= link_to _('DOCX'), template_export_path(template.dmptemplate_id, format: :docx), target: '_blank' %>
+                <br>
                 <%= link_to _('PDF'), template_export_path(template.dmptemplate_id, format: :pdf), target: '_blank' %>
               </td>
+              <td><%= template.org.name %></td>
```
jollopre: please, can you use the helper I wrote (https://github.com/DMPRoadmap/roadmap/blob/CDL-MVP/app/helpers/template_helper.rb) will reduce verbosity and it has test cases in place.

app/models/template.rb
```diff
@@ -170,6 +170,7 @@ def set_creation_defaults
       self.visibility = 1
       self.is_default = false
       self.version = 0 if self.version.nil?
+      self.links = { funder:[], sample_plan:[] } if self.links.nil?
```
jollopre: the default is set at db level on the migration file, if you don't change the type to TEXT, this line is not needed

db/migrate/20171206230107_convert_template_links_to_text.rb
jollopre: Unless we have a string greater than 65535 which probably we never have...

briri: Hmm interesting. If you go in and add 3 funder links and 2 sample plans you get
screen shot 2017-12-07 at 8 34 50 am

All of the varchars in our DBs here are limited to 255 (which was an older limit for MySQL pre 5.0). We're on 5.7 right now which, as you say allows 65k (depending on total row size), so we should be able to increase them on our end.

Its likely something carried over from the original DB we started with at the beginning of the project. I'll have to do some research to determine the best way for us to remove those limitations. In the interim, I'll back out the migration and move of the default to the code.

jollopre: @briri if running your migration solves the problem quickly that's fine but it would be nice to keep VARCHAR since we never going to have huge content for that field.

briri:
removed switch from VARCHAR to TEXT. Going to investigate converting all VARCHARs over from the old pre 5.0 255 limit to 65k. Will document this as other institutions upgrading an existing DMPonline_v4 DB may run into similar issues if they remain on MySQL.

Also updated template_helper's `links_to_elements method` to allow a separator to be passed in

app/helpers/template_helper.rb
```diff
@@ -1,8 +1,8 @@
 module TemplateHelper
-  def links_to_a_elements(links)
+  def links_to_a_elements(links, separator = ', ')
```
jollopre: good one! Even more reusable!

briri:
schema file still had TEXT instead of VARCHAR after reverting back. Tests are fixed now

Issue #869, opened 2017-11-27
Add Themes to Super Admin area


stephaniesimms:
both services need to be able to manage & edit themes as part of data migrations for MVP. 
- Add "Themes" page to Super Admin menu (screenshot 1)
- Create themes table (screenshot 2) that lists 14 [theme names](https://github.com/DMPRoadmap/roadmap/wiki/Themes) in one column & default DCC/UC3 guidance in second column. Note that second column should be labeled "Guidance." 
- Themes page includes a button to "Add new theme" > this takes user to a theme editing page (described in next bullet)
- when a user clicks on a theme name (or wants to add a new theme), the link should take them to an editing page (screenshot 3)
- on the theme editing page, user can edit the Theme name and/or the Guidance that accompanies that theme. Note that the larger box should be labeled "Guidance" (and not "Theme description").

![themes-menu](https://user-images.githubusercontent.com/12703638/33294506-c43f56a4-d385-11e7-842c-039927bd2185.png)

![theme-table](https://user-images.githubusercontent.com/12703638/33294578-288e21da-d386-11e7-80cb-f416042f9d2e.png)

![edit-theme](https://user-images.githubusercontent.com/12703638/33294603-3e167aca-d386-11e7-8762-0e0134410f22.png)

briri:
will bring this into the 'Admin' navigation menu as part of #902

stephaniesimms:
Themes page is an excellent addition! some things I noted when testing:
- we need a way to Remove themes. @sjDCC and @jollopre  do you have suggestions about the best way to do this?
- the pagination links aren't working; nothing happens when I click View All or any of the links on the right to page through (will address these in another issue). I tried adding a New Theme but I can't verify if it worked because I'm assuming it was added to the end of the table.

jollopre:
@stephaniesimms, to remove the themes, there are plenty options:

- Add a third column (Remove) on themes table
- Add a checkbox as first column that permits executing multiple actions simultaneously, where the actions can be defined on the table header (seen in many interfaces, like those to address email stuff)
- Add button delete when editing a theme record (probably the safest since hides as much as possible this action).

Can you add a new issue for delete theme according to whatever you and @sjDCC decide, please.

I will fix the pagination problem in this issue. Regarding adding new theme, we can go to sort by created_at DESC any time a new theme is added to the table.

jollopre:
- Pagination problem fixed
- Themes are now descendent sorted by updated_at

[...]

PR #922, opened 2017-12-08
Issue 869

jollopre:
- Fixed pagination for themes after changing policy
- Sorted themes by updated_at desc. DMPRoadmap/roadmap#869

### DIFF
--- a/app/models/theme.rb
+++ b/app/models/theme.rb
@@ -15,12 +15,7 @@ class Theme < ActiveRecord::Base
 
   validates :title, presence: {message: _("can't be blank")}
 
-  # EVALUATE CLASS AND INSTANCE METHODS BELOW
-  #
-  # What do they do? do they do it efficiently, and do we need them?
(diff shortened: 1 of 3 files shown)


app/models/theme.rb
```diff
@@ -15,12 +15,7 @@ class Theme < ActiveRecord::Base
 
   validates :title, presence: {message: _("can't be blank")}
 
-  # EVALUATE CLASS AND INSTANCE METHODS BELOW
-  #
-  # What do they do? do they do it efficiently, and do we need them?
-
-
-
+  scope :updated_at_desc, -> { self.all.order(updated_at: 'DESC') }
```
briri: Don't think a scope adds much value for us here. Also, for future reference, you don't need to use the `self.all` it is implied. Could just be: `scope :updated_at_desc, -> { order(updated_at: :desc) }`
No need to change this one now though, it works.

Issue #910, opened 2017-12-05
Always display Org name in the header


stephaniesimms:
To avoid confusion for Super Admins changing their Org affiliation, always display the Org name in the branding/header.

stephaniesimms:
this is working as expected although see new issue #932 to create an exception for "Other." closing this issue.

PR #925, opened 2017-12-08
Fixed Org Links & Banner now shows Org Name if no logo

briri:
- Fixed issues with Links on Org details page by switching to use logic used for Template Links (funder and sample plans). #866
- Updated banner to show Org name if no logo is available #910
- Added rake tasks to seed default values for links in both Orgs and Templates tables

### DIFF
--- a/app/controllers/orgs_controller.rb
+++ b/app/controllers/orgs_controller.rb
@@ -8,6 +8,7 @@ class OrgsController < ApplicationController
     @org = Org.find(params[:id])
     authorize @org
     @languages = Language.all.order("name")
+    @org.links = {"org": []} unless @org.links.present?
   end
 
   ##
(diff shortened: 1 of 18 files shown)


app/controllers/orgs_controller.rb
```diff
@@ -17,26 +18,13 @@ def admin_update
...
-          @org.links = json
-        else
-          failure = _('Unable to save your changes. Invalid URLs.')
-        end
-      else
-        failure = _('Unable to save your changes. Invalid URLs.')
-      end
-      attrs.delete('links')
+    if params[:org_links].present?
+      @org.links = JSON.parse(params[:org_links])
```
jollopre: nice refactor! We might need to to rescue if any parse error is arisen. Also in templates_controller.rb

app/validators/org_links_validator.rb
```diff
@@ -0,0 +1,13 @@
+class OrgLinksValidator < ActiveModel::Validator
+  include JSONLinkValidator
+  def validate(record)
+    links = record.links
+    if links.is_a?(Hash)
+      if !links.has_key?('org')
+        record.errors[:links] << _('A key "org" is expected for links hash') %{ :key => k }
```
jollopre: the hash key is never use, probably a copy/paste error.

app/views/layouts/_branding.html.erb
```diff
@@ -8,25 +8,31 @@
...
-      <% if user_signed_in? && !current_user.org.nil? %>
-          <ul class="nav navbar-nav">
-            <% current_user.org.links.each do |url| %>
-              <% unless url['link'].blank? %>
-                <li><%= link_to (url['text'].blank? ? url['link'] : url['text']), url['link'] %>
+      <% if user_signed_in? && !current_user.org.nil? && current_user.org.links.present? %>
+        <% if current_user.org.links['org'].present? %>
+            <ul class="nav navbar-nav">
+              <% current_user.org.links['org'].each do |url| %>
+                <% unless url['link'].blank? %>
```
jollopre: We might need to update the concern json_link_validators.rb to prevent blank string but meanwhile this check is good to have, thanks!

lib/tasks/bugfix.rake
```diff
@@ -58,4 +59,20 @@ namespace :bugfix do
...
+  desc "Set all orgs.links defaults"
+  task set_org_link_defaults: :environment do
+    Org.all.each do |org|
+      org.links = '{"org":[]}'
+      org.save!
+    end
+  end
+  
+  desc "Set all template.links defaults"
+  task set_org_link_defaults: :environment do
```
jollopre: not sure if this task is needed, when the migration that adds links field to template tables run, it should update any record to the default automatically.

Issue #928, opened 2017-12-11
Refactor SQL Query into scoped subqueries


xsrust:
Refactoring a large query into seperate, scoped subqueries which can be re-used individually within the code

```
# Gets dmptemplate_ids for a set of organisation ids passed whose migrated is false and customization_of is nil
  scope :families, -> (org_ids){ valid.where(org_id: org_ids, customization_of: nil).distinct.pluck(:dmptemplate_id)}
  # Gets max version for a set of dmptemplate_ids passed
  scope :dmptemplate_ids_with_max_version, -> (dmptemplate_ids){
    select(“MAX(version) AS version”, :dmptemplate_id).where(dmptemplate_id: dmptemplate_ids).group(:dmptemplate_id)
  }
  # Gets the most current versions of each template version family
  scope :current_templates, -> (dmptemplate_ids) {
    select(“templates.id, templates.dmptemplate_id, current.version”)
      .from(dmptemplate_ids_with_max_version(dmptemplate_ids), :current)
      .joins(“INNER JOIN templates ON current.version = templates.version AND current.dmptemplate_id = templates.dmptemplate_id”)
      .pluck(“templates.id”)
  }

# Call - (note the array is of Org_id)
Template.current_templates(Template.families([340998856, 187885252, 766049908, 201969613, 491834794]))
```

briri:
closing this one since there's no UAT for it.

PR #931, opened 2017-12-11
Issue 928

jollopre:
using activerecord for subqueries too. Case for latest version of templates. DMPRoadmap/roadmap#928


briri:
merged. if others have further refactoring that would be beneficial we can create an additional ticket

PR #935, opened 2017-12-11
Fix for contact_us_url on devise mailers

briri:
The `contact_us_url` keyword was causing failures when the value was not defined in the branding.yml. Updated to use a different variable name


app/views/devise/mailer/invitation_instructions.html.erb
```diff
@@ -2,7 +2,7 @@
   tool_name = Rails.configuration.branding[:application][:name]
   link = accept_invitation_url(@resource, :invitation_token => @token)
   helpdesk_email = Rails.configuration.branding[:organisation][:helpdesk_email]
-  contact_us_url = Rails.configuration.branding[:organisation][:contact_us_url] || contact_us_url
+  contact_us = (Rails.configuration.branding[:organisation][:contact_us_url] || contact_us_url)
```
jollopre: thanks for changing it, the rename is for sure safer!

PR #942, opened 2017-12-13
Add validation to Links (DO NOT MERGE)

briri:
Hey @jollopre,
Ticket #902 is a request to add validation to links dynamically. We need to:
1. verify that the URL is valid
2. verify that the text is present if a URL was entered (if they filled out one field they must fill out both)
3. verify that the URL is present if text was entered (if they filled out one field they must fill out both)

The current ariatiseForm.js doesn't work well with adding elements to the form this way. Please take a look at these proposed changes before I submit the final PR. I created a new JS called validation that borrows a lot of code from ariatiseForm (which it will eventually replace ... or we can update ariatiseForm with this code) but exposed some more methods that allow the entire form or individual elements to have their validation enabled/disabled dynamically. It also introduces a validate method that could be used to trigger field level validations as an input element loses focus.

Let me know what you think

### DIFF
--- a/lib/assets/javascripts/utils/validation.js
+++ b/lib/assets/javascripts/utils/validation.js
@@ -140,9 +140,6 @@ const invalid = (el) => {
 const addValidationMessage = (el) => {
   if (isString($(el).attr('id'))) {
     const id = $(el).attr('id');
-
-console.log(isString($(el).attr('aria-describedby')));
-
     if (!isString($(el).attr('aria-describedby'))) {
       $(el).after(blockHelp(`help-${id}`, getValidationMessage(el)));
       $(el).attr('aria-describedby', `help-${id}`);
--- a/lib/assets/javascripts/utils/validation.js
+++ b/lib/assets/javascripts/utils/validation.js
@@ -1,0 +1,226 @@
+import { Tinymce } from './tinymce';
+import { isObject, isString, isBoolean } from './isType';
+import * as constants from '../constants';
+import * as validator from './isValidInputType';
+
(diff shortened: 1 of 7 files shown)


lib/assets/javascripts/utils/validation.js
```diff
@@ -0,0 +1,223 @@
...
+    return ($(ctx).attr('aria-required') && $(ctx).attr('aria-required') === 'true');
+  }
+  return false;
+};
+const checkValidations = (el) => {
+  const type = getValidationTypeForElement(el);
+  const value = getValue(type, el);
+
+  // If the field is required or it has a value
+  if (isRequired(el) || isString(value)) {
```
jollopre: why are we checking if isString here?

lib/assets/javascripts/utils/validation.js
```diff
@@ -0,0 +1,223 @@
...
+      invalid(el);
+      return false;
+    }
+  }
+  return true;
+};
+
+export const isValidatable = (ctx) => {
+  if (isObject(ctx)) {
+    return isString($(ctx).attr('data-validatable'));
```
jollopre: why we adding data-validatable?

briri: ah function is leftover from earlier. I guess linter didn't throw an 'unused const' error because its being exported.

briri:
thanks for reviewing @jollopre. Will remove those 2 items, rebase with current CDL-MVP, do some further testing and then resubmit the PR

Issue #939, opened 2017-12-12
Add functionality to delete/remove themes


stephaniesimms:
Super Admins need a way to remove the themes. After discussion we agreed the easiest/best way is to add a Delete button when editing a theme record. So in the screenshot below, insert a new button on lower right in the order: Save - Delete - Cancel. When a user clicks "Delete" include the confirmation message step "Are you sure you want to delete..."
 
![screen shot 2017-12-12 at 7 15 28 pm](https://user-images.githubusercontent.com/12703638/33920076-da1f0680-df70-11e7-8fd6-9b0a140d9ec8.png)

sjDCC:
Yes, agree. #3 sounds best

stephaniesimms:
all working as expected. but the banner msg made me LOL. please revise the msg text @jollopre to the following:

"Notice: Successfully deleted your theme."

sjDCC:
This looks good and works well.

I noticed that we have the existing themes in Roadmap, not the new ones. This is probably because it's DMPonline data. Do we need to add a separate ticket to migrate to / implement new ones? Here's a summary of the mapping from one to another in case that helps.

[DMP-theme-migration-2018.docx](https://github.com/DMPRoadmap/roadmap/files/1568501/DMP-theme-migration-2018.docx)

Issue #933, opened 2017-12-11
Clicking cancel deletes Guidance from the list 


stephaniesimms:
When I select "Remove" from the Guidance actions menu I receive an "Are you sure..." message prompt as expected. However, my guidance is removed/deleted when I click the cancel button (screenshots below). Desired behavior is to only remove/delete guidance if user clicks Ok button.

![screen shot 2017-12-11 at 12 03 30 pm](https://user-images.githubusercontent.com/12703638/33851375-b5e174de-de6b-11e7-9038-27d25122ff10.png)

![screen shot 2017-12-11 at 12 06 59 pm](https://user-images.githubusercontent.com/12703638/33851425-d4836b54-de6b-11e7-8dbe-d0eb283451be.png)

sjDCC:
Works fine now. Only removes entry when you confirm you wish to do so. Closing out

Issue #870, opened 2017-11-27
Add modal: are you sure you want to remove plan?


stephaniesimms:
It's easy for a user to "Remove" a plan accidentally by selecting the action item on the My Dashboard page (screenshot 1). When I select Remove, a modal flashes on the screen, then disappears and I receive the message in the second screenshot.

Desired behavior is to provide the user with a modal that asks: "Are you sure you want to remove this plan?" The user can click a button for OK or Cancel to complete or cancel the action. 

![remove](https://user-images.githubusercontent.com/12703638/33295208-ec7eab6c-d388-11e7-99f3-62a84ec90f59.png)

![notice-removed](https://user-images.githubusercontent.com/12703638/33295228-28006dc4-d389-11e7-9bac-cce3173622ce.png)

sjDCC:
This works fine now. Thanks @briri Closing out

PR #947, opened 2017-12-13
Issue 939

jollopre:
destroy action for themes. DMPRoadmap/roadmap#939


briri:
Here's a screenshot of the issue referenced above
screen shot 2017-12-13 at 10 12 48 am

jollopre:
I decided to remove lib/assets/javascripts/utils/linkHelper.js since this functionality is best handed through our third-party JS file (e.g. https://github.com/rails/jquery-ujs/blob/master/src/rails.js) part of vendor.js. 

The reason to this removal is because any link with data-remote attribute was augmented with a hidden form that was triggered anytime the link was clicked without taking into account confirmation dialogs (e.g. Cancel for cancelling the action, etc).

Moreover, as jquery-ujs, offers AJAX support out of the box by just annotating a form, button or link with remote: true. I have made changes in some files according to the custom events triggered by jquery-ujs to showcase how it works. Files changed:

- lib/assets/javascripts/views/plans/index.js
- lib/assets/javascripts/views/plans/share.js

To be consistent, we should adapt also the files but I think a new issue for that would be better. Below the remaining files to change:
- lib/assets/javascripts/views/answers/edit.js
- lib/assets/javascripts/views/notes/index.js
- lib/assets/javascripts/views/org_admin/templates/index.js
- lib/assets/javascripts/views/plans/new.js

jollopre:
I attach here a link to the unobtrusive library that we have in place in case new work has to be added https://github.com/rails/jquery-ujs/wiki

briri:
merging after discussion with @jollopre since this resolves the issue with the cancel button not working on confirmation messages. Will create a separate ticket to apply the changes to other ajax calls and make sure issues #870 and #933 are resolved.

Issue #911, opened 2017-12-06
Display full list of Themes & make it easier to select multiple


stephaniesimms:
It's difficult to scroll through the list of themes and make selections for guidance or templates. The current tooltip instructs users to select multiple themes w/CTRL but Mac users have to use Command - all users have to scroll up and down repeatedly to search and select. To improve usability we should display all 14 themes clearly. Remove None from the list in related issue #907

let's try laying them out in three columns (5x5x4) with tick boxes

![screen shot 2017-12-05 at 4 04 46 pm](https://user-images.githubusercontent.com/12703638/33637483-0cd12d38-d9d6-11e7-8d03-96168fd9d771.png)

briri:
Switched to display the themes as checkboxes
![screen shot 2017-12-13 at 12 04 40 pm](https://user-images.githubusercontent.com/1204467/33959945-f33dd74e-dffd-11e7-8fc1-eaf9d23332f5.png)

stephaniesimms:
I think this works to have all themes laid out w/checkboxes but will wait for @sjDCC to confirm. 

One thing that requires attention is the tooltip that seems to be creating display problems when i hover over the themes (see screenshot below) - especially w/the long list of 28 some groups of themes disappear depending on where i move focus. I think it would be better to implement a question mark/tooltip next to the word Themes here. we have popover/tooltip issues elsewhere and need to find a holistic solution.  

Proposed text based on DMPonline tooltip:
"Select one or more themes that are relevant to this question. This will display your generic institution-level guidance, as well as that from other sources e.g. the %{organisation_abbreviation} or any Schools/Departments that you provide guidance for."

![screen shot 2017-12-14 at 2 09 09 pm](https://user-images.githubusercontent.com/12703638/34017005-c6b656fc-e0d8-11e7-980b-6d12631fe7da.png)

sjDCC:
Sounds good @stephaniesimms Tooltip is at high-level, not individual themes

sjDCC:
The tooltip doesn't encourage now as much as it did before but it's a completely different style. Instead of propping up when you hover-over / scroll through this question area, you need to click the help icon which reloads the page and jumps you to the top. You only see the tooltip when you scroll back down.

![hover-over](https://user-images.githubusercontent.com/6169940/34525626-82bf647e-f097-11e7-9072-aaf9d24fd6d8.png)


Can it be implemented in the same style as the others i.e. black box offset to right which pops up automatically when question area is active?

[...]

PR #948, opened 2017-12-13
added checkbox based theme selector

briri:
#911 
Add a checkbox based theme selector for the Guidances and Questions pages


app/views/org_admin/shared/_theme_selector.html.erb
```diff
@@ -0,0 +1,28 @@
+<%# locals: all_themes %>
+<div class="form-group">
+  <%= f.label _('Themes'), for: :theme_ids, class: 'control-label' %>
+  <div class="row">
+    <% if all_themes.length > 0 %>
+      <% 
+        cntr = 0
+        nbr_of_cols = (all_themes.length.to_f / MAX_NUMBER_THEMES_PER_COLUMN.to_f).ceil
+        col_size = (12 / (nbr_of_cols > 4 ? 3 : nbr_of_cols)).round
```
jollopre: This could be problematic for instances with a number of themes less than the MAX_NUMBER_THEMES_PER_COLUMN. That would mean the size of the column would be 1 and therefore the theme label would appear cut.

jollopre: Using the modulo here the syntax is simplified and cover every case. This would be my proposal:

```
<div class="form-group">
  <%= f.label _('Themes'), for: :theme_ids, class: 'control-label' %>
  <div class="row">
    <% if all_themes.length > 0 %>
      <% group_nbr = -1 %>
      <% all_themes.each_with_index do |theme, i| %>
        <% if i % MAX_NUMBER_THEMES_PER_COLUMN == 0 %>
          <div class="col-md-4">
          <% group_nbr += 1 %>
        <% end %>
        <div class="checkbox" data-toggle="tooltip" data-html="true" title="<%= theme.description %>">
          <%= f.label :theme_ids, raw("#{f.check_box :theme_ids, {multiple: true}, theme.id, nil} #{theme.title}") %>
        </div>
        <% if (i - group_nbr * MAX_NUMBER_THEMES_PER_COLUMN) == (MAX_NUMBER_THEMES_PER_COLUMN - 1) %>
          </div>
        <% end %>
      <% end %>
    <% else %>
      <p><%= _('No themes have been defined. Please contact your administrator for assistance.') %></p>
    <% end %>
  </div>
</div>
```

 do you mind to try it?

briri: sure. will give it a try.

briri: tested both with different array lengths and they both produce the same results. The label does not get clipped when there is only one column. Going to leave as is.

briri:
merging this one in since our contractor has started and will benefit from this simplification of theme selection

briri:
submitting an additional PR afterward to remove some old guidance views/partials that are no longer used

PR #949, opened 2017-12-13
Modal dialog for confirmation messages

briri:
Discovered that existing confirmation messages were performing a delete even if the user clicked 'Cancel'. Addresses #870, #933
- Added a modal confirmation dialog that allows user to Cancel or Continue
- Updated Templates, Plans, Guidance Groups, and Guidance to use it

@jollopre you will need to apply to the theme page for #939


briri:
Closing after conversation with @jollopre. The confirmation box's cancel button will begin working if we convert over to the rails/jquery-ujs in PR #939.

Issue #953, opened 2017-12-14
Error saving template when not a funder Org


briri:
Receive an error on the empty Links hash when trying to create a new template as a non-funder org.
![screen shot 2017-12-14 at 9 54 47 am](https://user-images.githubusercontent.com/1204467/34006955-0006d4be-e0b5-11e7-9a54-8c18d6da3c4e.png)

sjDCC:
All seems to work fine now. Could create a template for Edinburgh. Closing out.

PR #954, opened 2017-12-14
Convert 'Remove' links to rails/jquery-ujs

briri:
Converted Remove links on Plans and Templates pages over to the rails/jquery-ujs approach. 
- Fixes issues with cancel button on remove confirmation #870 and #933
- Also fixed a bug #953 that caused template creation to fail if user did not belong to a funder org
- Also fixed issue with identifying customizations on templates page

### DIFF
--- a/app/controllers/org_admin/templates_controller.rb
+++ b/app/controllers/org_admin/templates_controller.rb
@@ -13,13 +13,13 @@ module OrgAdmin
 
       # If the user is an Org Admin look for customizations to funder templates
       customizations = {}
-      unless current_user.can_super_admin?
+      if current_user.can_org_admin?
         funder_templates.each do |funder_template|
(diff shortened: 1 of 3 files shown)


app/controllers/org_admin/templates_controller.rb
```diff
@@ -53,6 +53,7 @@ def create
       @template = Template.new(params[:template])
       @template.org_id = current_user.org.id
       @template.description = params['template-desc']
+      @template.links = (params["template-links"].present? ? JSON.parse(params["template-links"]) : {"funder": [], "sample_plan": []})
```
jollopre: rather than using here the ternary, if you use like that:
```
 @template.links = JSON.parse(params["template-links"]) if params["template-links"].present?
```
will only update template.links with whatever comes from the client otherwise will default to whatever is defined in JUST one place. Anyway, when you create a new Template it should take the defaults from the db unless they have not been removed in a newer commit.

Issue #938, opened 2017-12-12
Annotations layout and functionality issues


stephaniesimms:
some additional things noted when testing related issue #791 - as an Org Admin customizing templates : 
- when I click to Add Annotations, the text editors for Example Answer and Guidance are tiny (screenshot 1). they should be the same size as other TinyMCE editors on the Question editing page, e.g. (screenshot 2)  
- Save-Delete-Cancel buttons should appear beneath the editors, aligned to the right. Currently there is only Save-Cancel.
- after I create an annotation, I get a weird array of buttons (screenshot 3). there should only be Save-Delete-Cancel.
- if I click either of the duplicate buttons to "Delete Example Answer" I get the error message in screenshot 4

1
![add-annotations](https://user-images.githubusercontent.com/12703638/33866588-734c3fd2-deac-11e7-84c4-0b1edbcb303d.png)

2
![screen shot 2017-12-11 at 7 51 21 pm](https://user-images.githubusercontent.com/12703638/33866686-fad0d4c2-deac-11e7-8319-9842a22a657f.png)

3
![del-annotations](https://user-images.githubusercontent.com/12703638/33866597-7bdb6b50-deac-11e7-9c78-2190b77d16aa.png)

4
![delete-error](https://user-images.githubusercontent.com/12703638/33866607-85258da8-deac-11e7-9cb9-2566ae3ed93f.png)

sjDCC:
I'm not able to test this one as the 'Customise' action isn't displaying. I thought it might be a browser issue but the same happens on Firefox and Chrome. I have all the relevant Org Admin permissions
![customise](https://user-images.githubusercontent.com/6169940/34525319-6220d320-f096-11e7-99c2-194a19010d32.PNG)

It's not possible to do it as a super admin either as no actions are shown unless I change my affiliation to a funder (which then lets me edit the template rather than customise)
![super-admin](https://user-images.githubusercontent.com/6169940/34525344-7e9e419a-f096-11e7-9f4f-52055f7ff2c1.PNG)

briri:
Yes, this is a bug for sure. It has been addressed as part of #900. I'll have someone review it today so that we can get the fix out on the server.

stephaniesimms:
bug is fixed @briri - thanks. after some testing i noted the following issues:
- after clicking to add Annotations I only have Save - Cancel buttons (screenshot). there should be 3 buttons: Save - Delete - Cancel
[...]

PR #956, opened 2017-12-14
consolidated edit/add view and controller action for annotations

briri:
addressed issues with the annotations sections when updating a customization #938
- Consolidate separate edit/add pages and controller methods
- Also updated that the plans 'remove' link needed to be converted to jquery-ujs attributes


briri:
closing to address issue with annotations display on plans page

PR #958, opened 2017-12-15
Issue 546

jollopre:
plan overview for plans with one section. 
margin bottom for separating sections. DMPRoadmap/roadmap#546


app/views/phases/_overview.html.erb
```diff
@@ -0,0 +1,25 @@
+<%# locals: { phase } %>
+<div class="row">
```
briri: nice refactor to prevent DRY

PR #968, opened 2017-12-15
added popover style tooltip for theme selection

briri:
addresses final issue on #911 
- added generic popover partial and JS
- updated guidance and question pages to use popover for theme selector

### DIFF
--- a/app/views/shared/_popover.html.erb
+++ b/app/views/shared/_popover.html.erb
@@ -1,0 +1,8 @@
+<%# available locals: message, placement %>
+<% if message.present? %>
+  <a href="#" class="btn btn-default" data-toggle="popover" 
+     title="<%= _('More information') %>" data-content="<%= message %>" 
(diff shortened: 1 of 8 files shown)


lib/assets/javascripts/utils/popoverHelper.js
```diff
@@ -0,0 +1,8 @@
+import 'bootstrap-sass/assets/javascripts/bootstrap/popover';
+
+$(() => {
```
jollopre: options for the popover can be passed directly in the html link too (e.g. data-animation="fade" data-placement="right").

app/views/shared/_popover.html.erb
```diff
@@ -0,0 +1,8 @@
+<%# available locals: message, placement %>
```
jollopre: nice to see this generic partial for popovers. I think would be worth to annotate it somewhere so we can re-use in other places too.

PR #969, opened 2017-12-15
consolidated edit/add view and controller action for annotations

briri:
Updates for #938 
- Consolidated edit/add annotation view and controller actions
- Updated tests
- Updated annotation show partial to work properly for template editing and plan answer pages
- Fixed deletion buttons
- Blanking out an annotation is now the same as deleting it


lib/assets/javascripts/views/questions/show.js
```diff
@@ -6,18 +6,15 @@ $(() => {
...
 
-  $('.edit_annotations_button').on('click', (e) => {
-    const source = e.target;
-    const target = $(source).attr('href');
-    $(target).show();
-    $(source).hide();
-    $(target).closest('.col-md-12').find('.show_annotations_div').hide();
+  $('.cancel_edit_annotations').on('click', (e) => {
+    e.preventDefault();
+    $($(e.target).attr('href')).hide();
```
jollopre: please can you extract $(e.target) as a constant (e.g. const target = $(e.target) otherwise we read the DOM element twice.

lib/assets/javascripts/views/questions/show.js
```diff
@@ -6,18 +6,15 @@ $(() => {
...
   });
 
-  $('.add_annotations_button').on('click', (e) => {
-    const source = e.target;
-    const target = $(source).attr('href');
-    $(target).show();
-    $(source).hide();
+  $('.annotations_button').on('click', (e) => {
+    e.preventDefault();
+    $($(e.target).attr('href')).show();
```
jollopre: please can you extract $(e.target) as a constant (e.g. const target = $(e.target) otherwise we read the DOM element twice.

app/views/paginable/plans/_privately_visible.html.erb
```diff
@@ -77,8 +77,8 @@
                     <% end %>
                     <% role = plan.roles.where(user_id: current_user.id).first %>
                     <% conf = (role.creator? && plan.publicly_visible?) ? _("Are you sure you wish to remove this public plan? This will remove it from the Public DMPs page but any collaborators will still be able to access it.") : _("Are you sure you wish to remove this plan? Any collaborators will still be able to access it.") %>
-                    <li><%= link_to _('Remove'), deactivate_role_path(role),
-                                    method: :put, data: { confirm: conf } %></li>
+                    <li><%= link_to _('Remove'), deactivate_role_path(role), 'data-method': 'put', rel: 'nofollow',
```
jollopre: the action roles#deactivate returns HTML and there is no handle associated for ajax:success and ajax:error in any JS file that I can see in this PR. The quickest workaround is leaving as a plain action that refreshed the entire page, the elegant would be refactoring the action to return json and add handlers in JS.

app/views/annotations/_new_edit.html.erb
```diff
@@ -0,0 +1,22 @@
+<!--edit/add annotations. question is passed as an argument-->
```
jollopre: good to see shared partials!

app/controllers/annotations_controller.rb
```diff
@@ -119,4 +49,28 @@ def init_annotation(text, question, org, type)
     return annotation
   end
 
+  def process_changes(hash, input, type)
```
jollopre: I don't see the need of this private function unless it is re-used in several actions.

app/controllers/annotations_controller.rb
```diff
@@ -2,110 +2,40 @@ class AnnotationsController < ApplicationController
...
-    @section = @question.section
-    @phase = @section.phase
-    @phase.template.dirty = true
-    if @example_answer.destroy
-      redirect_to admin_show_phase_path(id: @phase.id, section_id: @section.id, edit: 'true'), notice: success_message(_('information'), _('deleted'))
-    else
-      redirect_to admin_show_phase_path(id: @phase.id, section_id: @section.id, edit: 'true'), notice: flash[:alert] = failed_destroy_error(@example_answer, _('example answer'))
+    annotation = Annotation.find(params[:id])
+    authorize annotation
+    phase_id = annotation.question.section.phase.id
```
jollopre: with this change still needs to run 4 queries. I would propose this:
```
Annotation.joins("INNER JOIN questions ON annotations.question_id = questions.id").joins("INNER JOIN sections ON questions.section_id = sections.id").joins("INNER JOIN phases ON sections.phase_id = phases.id").where("annotations.id": params[:id]).pluck("phases.id").first
```
although it is quite verbose, just loads the phase.id and since we redirect to admin_show anyway...

app/controllers/annotations_controller.rb
briri: its called twice within the `admin_update` action

app/views/paginable/plans/_privately_visible.html.erb
briri: agreed that handling the entire transaction via JS would be more elegant but at this point I don't think we have time to refactor all of these Remove/Delete actions. Maybe something to work on post MVP.

app/controllers/annotations_controller.rb
jollopre: I did not notice it was called twice within admin_update, in that case, it makes sense to have a private method. I don't see the case to using Proc or its specialised Lambda to this purpose since the method process_changes is not using any variable defined at the scope of admin_update

app/controllers/annotations_controller.rb
```diff
@@ -119,4 +49,28 @@ def init_annotation(text, question, org, type)
     return annotation
   end
 
+  def process_changes(hash, input, type)
+    # If the input is available update the annotation otherwise remove it if it exists
+    if input.present? 
+      annotation = Annotation.find_or_create_by(hash)
```
jollopre: find_or_create does not guarantee exclusivity if two or more threads are running same line of code, in case you wanted to avoid race conditions per database connection consider using transactions.

Issue #729, opened 2017-10-16
Org branding/banner must include up to 3 links


stephaniesimms:
Related issue #726 

Org admins can provide up to 3 links on the Org details page. The branding/banner design needs to accommodate these links.

vyruss:
Add relevant code to make these links functional:
- "`links`" JSON field to `org` to give us flexibility in what org admins store there
  - Optionally display none, one or two links with both text and link customisable
  - Flexible for future use (in case we ever need more than 2 links per org).
- Code in view for layout and to make sure we only display existing links

stephaniesimms:
Note that these links should open in a new tab. Currently they open in the same window/tab.

sjDCC:
Will these links have icons like in the DMPTool? e.g.
![icons](https://user-images.githubusercontent.com/6169940/33175484-d9a9f63c-d053-11e7-982c-12d595266442.PNG)

I think it would be good to include this. The contact email could have the chat bubbles and any standard links the computer icon. Without them it may not be obvious that these are links to support as it just shows as text
![roadmap-stage](https://user-images.githubusercontent.com/6169940/33175550-01d32b1a-d054-11e7-9122-2aad498c7e2f.PNG)

vyruss:
Temporarily blocked by #866

stephaniesimms:
created issue #904 to report double redirect bug. please delete it if unnecessary/covered by #866

sjDCC:
Tested this for Edinburgh and it looks fine

![banner](https://user-images.githubusercontent.com/6169940/34524248-b78d9982-f092-11e7-8f77-e88578b370f6.PNG)

I did generate the following error when adding the links though. It's seems there's a fairly restrictive character count so it wouldn't allow my entry of "Research Data Services" for the second link
![error](https://user-images.githubusercontent.com/6169940/34524283-da44775c-f092-11e7-81e2-d2c26ea29ac8.PNG)

Ideally we would increase the length allowed to c.40 characters and better noticing is needed. Please add a tooltip to state the character count and an error message that explains why save is disallowed instead of generic warning.

briri:
ah yes, we discovered this restriction a few weeks ago. Its related specifically to MySQL and would likely be something that our collaborators would encounter. 
@bhavi, you were looking into removing the 255 char restriction on our VARCHAR table columns before the holiday break. Did you find a solution that we can run on roadmap-stg?

sjDCC:
[...]

PR #970, opened 2017-12-16
Set up org-links and contact info in org toolbar for #729

vyruss:
Fixes # 729.

Changes proposed in this PR:
- iterate over org links in org-navbar and add contact email as last link

### DIFF
--- a/lib/assets/stylesheets/overrides.scss
+++ b/lib/assets/stylesheets/overrides.scss
@@ -16,7 +16,7 @@ $highlight-background-color: #F36F24;
   font-weight: normal;
 }
 
-/* Font Smoothing in Webkit /* 
+/* Font Smoothing in Webkit /*
 body {
   -webkit-font-smoothing: antialiased;
 }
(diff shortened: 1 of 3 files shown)


vyruss:
@briri thanks for the suggestions, this should work better now

lib/assets/stylesheets/overrides.scss
```diff
@@ -262,6 +262,16 @@ $header-background: linear-gradient(#ED6406, #F59503);
     height: 102px;
     padding-top: 0;
   }
+  #org-navbar .org-links {
+    padding-top: 15px;
+  }
+  #org-navbar .org-a {
+    line-height: 36px !important;
```
briri: prefer not to use `!important` as it essentially the nature/benefit of a cascading stylesheet. You should be able to override the default Bootstrap behavior by being more specific with your selector. 
Something like `#org-navbar .navbar-nav li .org-a` should get you what you want

Issue #815, opened 2017-11-13
Standardize tables with more than 10 items


stephaniesimms:
Related #786 
Implement pagination and add a search box above any table within the system that exceeds 10 items 

Add pagination logic to the following pages:
- Public DMPs
- DMP Templates (Public templates)
- Plans (Regular user and Org Admin)
- Templates
- Guidance
- 

Search box example
![screen shot 2017-11-13 at 3 29 47 pm](https://user-images.githubusercontent.com/12703638/32754911-b57fd0f2-c887-11e7-8377-58b9a6c8bc97.png)

Pagination example
![screen shot 2017-11-13 at 3 30 18 pm](https://user-images.githubusercontent.com/12703638/32754906-b14b0d6c-c887-11e7-9928-e18b93daa852.png)

briri:
Change the existing table search box so that it looks like the search box above (change placeholder text and position).
Also make sure that all of these tables has have sortable columns (where applicable).

briri:
Using the 'TABLE_FILTER_MIN_ROWS' for both the size of the paginated table and the trigger for whether or not the search/filter box appears is problematic. Paginated tables now only ever show 10 rows so the search/filter box is always hidden.

Would probably be better to trigger the display of the search/filter box via JS and the presence of more than one 1 page.

jollopre:
I thought we will display that search box only when View all is clicked, i.e. when all the records for a model are shown, otherwise it is not displayed.

stephaniesimms:
the search box should be displayed if there are more than 10 items in a table - regardless of whether the user clicks to view all or not. it should still be possible to search the long table without having to expand it to see the full contents.

briri:
@jollopre in the DMPTool code we are passing the search/filter value in as a param . We should be able to do something similar. We could then let the pagination handle however many results are returned by the search criteria.
You can see an example of it in use on the public Templates page: https://dmptool.org/guidance 
Here's is the controller logic for that page: https://github.com/CDLUC3/dmptool/blob/7650b5e551c3ab640a97d92eb4d627f88078243e/app/controllers/static_pages_controller.rb#L103
and the model logic that does the actual search:
https://github.com/CDLUC3/dmptool/blob/7650b5e551c3ab640a97d92eb4d627f88078243e/app/models/requirements_template.rb#L79

briri:
after doing some testing I noticed that the current column sorts are behaving in the same manner so we would need to follow the same pattern as mentioned above for that as well.

[...]

PR #971, opened 2017-12-18
Issue 815 (DO NOT MERGE)

jollopre:
filterable concern to chain scope methods for paginable, searchable and sortable models. DMPRoadmap/roadmap#815

### DIFF
--- a/app/models/concerns/filterable.rb
+++ b/app/models/concerns/filterable.rb
@@ -1,0 +1,39 @@
+module Filterable
+  extend ActiveSupport::Concern
+
+  module ClassMethods
+    # Allows creating a chainable query given a hash of filter_params passed
+    # filter_params argument must be a Hash with at least paginable_all key, otherwise ArgumentError is raised
+    # It raises NoMethodError if the value of the key passed is different from ActiveRecord::Relation and it does
+    # not define a method with the same name as the key passed
+    # 
+    # User.filter({ paginable_all: current_user.org.users.includes(:roles), paginable_page: 1, search: "@gmail.com" })
+    # will retrieve every user within the same org as the current user with an email including @gmail.com and limitting results to 10
+    #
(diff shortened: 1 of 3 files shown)


briri:
I would prefer for the url to be a bit less verbose. maybe something like `paginable/users?p=1&s=@dcc.ac.uk&o=last_sign_in_at&d=desc`
p = page, s = search term, o = order by column, d = order direction
The default page 'p' would be 1 if it was not present and the default direction 'd' would be ascending. 

I'm also not sure how flexible this concern will be in the long run. I like the introduction of a 'search' scope to the model but the concern doesn't really provide any advantage here. You could just as easily place this code in the controller as something like: 
```ruby
def index
  # determine sort order (default could be defined in the model)
  sort = (params[:o].present? && User.respond_to?(params[:o]) ? { cols: params[:o], dir: (params[:d] || 'asc') } : { cols: 'email', dir: 'asc' })
  # either call the model's search scope or all
  results = (params[:s].present? && User.respond_to?(:search) ? User.search(params[:s]).order(sort[:cols], sort[:dir]) : User.all.order(sort[:cols], sort[:dir]))

  # Now do pagination
end
```

jollopre:
Thanks for your views @briri. We should not worry about the verbosity of the URL since the are not going to be available in the browser URL and we could convert to POST if they were very lengthy. The purpose of the concern is to map param argument and values to methods defined in the model and argument passed to them. The search functionality has to be explicitly defined according to a set of fields so it would be always necessary to be written, in respect to the sort we can use the powerful mechanism of adding method dynamically. Remember, methods with single functionality are always better to maintain and test. Also, adding dynamic methods for sorting means adding new columns to table will work out of the box without needing to tweak the concern. Note, this concern will be using together with the paginable. 

I close this PR and will send a new PR when all the bits I'm thinking are properly stated.

PR #973, opened 2017-12-18
added nil checks

briri:
layouts/_branding.html.erb was failing when loading some tests because the local test DB does not always have org.links defined. Added a few `.present?` checks to allow the page to load.


briri:
merging this small fix to the work done on PR #970

briri:
turned out to just be an issue with my local test db. Was fixed with a `rake db:drop RAILS_ENV=test && rake db:create RAILS_ENV=test && rake db:schema:load RAILS_ENV=test`

Issue #955, opened 2017-12-14
Add predefined filters to the top of the Templates tables


briri:
Once #815 is complete, Add 3 predefined filters above the funder and normal template tables on the templates page. The following filters should be clickable links that provide a record count and apply only to the specific table (similar to the screenshot below):
- link text: 'All (%{nbr_templates})', behavior: reloads the table removing any search criteria
- link text:'Published (%{nbr_templates})', behavior: fetches only the published templates removing any previous search criteria
- link text:'Unpublished (%{nbr_templates})', behavior: fetches only the unpublished templates removing any previous search criteria
![screen shot 2017-12-05 at 11 48 34 am](https://user-images.githubusercontent.com/12703638/33627417-4db649a6-d9b2-11e7-8ad4-40c06930b89e.png)

briri:
All 3 predefined filters/scopes that work with pagination:
![screen shot 2017-12-19 at 11 39 17 am](https://user-images.githubusercontent.com/1204467/34175453-5440cda0-e4b1-11e7-82b0-71f1f361e45c.png)

stephaniesimms:
the filters are great, but some behavior is wonky. first general comments:
- it would be helpful to highlight the selected filter somehow, either make it bold or retain the gray box when I click one of the 3 (All, Published, Unpublished)
- we also need to clarify somehow that for Org Admins "Published/Unpublished" refers to their customizations on a funder template as opposed to the template status itself (screenshot). @sjDCC thoughts?

![screen shot 2018-01-02 at 3 52 14 pm](https://user-images.githubusercontent.com/12703638/34504688-f37c7c56-efd4-11e7-9727-07edf0ff8f66.png)

things that don't work as expected:
- when I click "Published" in both tabs as a Super Admin I get the expected list, but if I click the View all link to expand the list it seems to revert back to displaying All templates
- as an Admin I should see 1 Published template (screenshot) for Univ of Edinburgh but it doesn't appear in the count or when I click "Published"
![screen shot 2018-01-02 at 3 45 29 pm](https://user-images.githubusercontent.com/12703638/34504575-275f175a-efd4-11e7-9670-0db018d2679e.png)


- when I click "Unpublished" in both tabs I get a mixed list that appears to contain All templates, with published and unpublished things. It's a different list from "All" but I'm not sure what the logic is for displaying things here.

[...]

PR #975, opened 2017-12-19
added all/published/unpublished scopes to the templates tables

briri:
issue #955 
Adds predefined scopes for 'All', 'Published' and 'Unpublished' to the templates tables


briri:
tweaked the way customizations were loaded so that it only runs a single query

PR #979, opened 2017-12-20
Admin Guidance page - revise publish action #871

bhavi:
…n guidance create/edit

Fixes # .

Changes proposed in this PR:
-

### DIFF
--- a/app/controllers/guidances_controller.rb
+++ b/app/controllers/guidances_controller.rb
@@ -42,9 +42,18 @@ class GuidancesController < ApplicationController
       guidance_params[:theme_ids].map{|t| guidance.themes << Theme.find(t.to_i) unless t.empty? }
     end
 
+    guidance_group = GuidanceGroup.where(id: guidance.guidance_group_id).first
+    unless guidance_group.nil?
+      if !guidance_group.published? || guidance_group.published.nil?
+        guidance_group.published = true
+        guidance_group.save
+      end
+    end
+    
     if guidance.save
       flash[:notice] = success_message(_('guidance'), _('created'))
       redirect_to(action: :admin_index)
+
     else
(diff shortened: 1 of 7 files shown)


app/controllers/guidances_controller.rb
```diff
@@ -42,9 +42,18 @@ def admin_create
       guidance_params[:theme_ids].map{|t| guidance.themes << Theme.find(t.to_i) unless t.empty? }
     end
 
+    guidance_group = GuidanceGroup.where(id: @guidance.guidance_group_id).first
```
briri: you should be able to switch this to `GuidanceGroup.find(@guidance.guidance_group_id)`. Its more of a preference though so ok to leave as is for now.

app/models/org.rb
```diff
@@ -60,6 +60,10 @@ class Org < ActiveRecord::Base
 
   # Predefined queries for retrieving the managain organisation and funders
   scope :managing_orgs, -> { where(abbreviation: Rails.configuration.branding[:organisation][:abbreviation]) }
+  scope :funders, -> { where(org_type: 2) }
```
briri: Please remove these 2 scopes. They are provided already by the FlagShiTzu gem. I had removed them in an earlier PR.

briri:
Also need to fix the tests

app/controllers/guidances_controller.rb
bhavi: when a guidance_group_id doesn't exist for the particular guidance, the find method throws nil error and doesn't proceed to the unless statement at all. Whereas the where condition returns an empty active relation and proceeds without the exception. 
Thus, have been using the where clause to prevent nil exception whenever need to succeed to a conditional loop (mostly coding preference, open to suggestions !)

briri:
Looks good Bhavi. Thanks for fixing the tests.

Issue #873, opened 2017-11-28
Add download button for org plans in Admin area


stephaniesimms:
Add a "Download plans" button (suggestion to place in upper right, above table) to enable Admins to download a CSV file of the user plans info.

The CSV should only contain the plan details info (from this table) and not the full content of the plans.

![screen shot 2017-11-28 at 10 31 02 am](https://user-images.githubusercontent.com/12703638/33337261-479f7474-d427-11e7-8450-0d9a0973b70c.png)

sjDCC:
Download should probably be restricted to plan details (i.e. the metadata in this table) rather than the full content of all plans

briri:
Agreed. We should probably just export the contents of the table

stephaniesimms:
this looks and works great @briri - the only thing I can see that might be useful is to add another column to the downloaded csv for "Owner name" and rename the current field "Owner email" but only if this is super easy. The Plans table lists the name under Owner and the csv lists the email/contact info. Org admins would likely want to have both.

stephaniesimms:
looks good. closing issue.

sjDCC:
Looks good to me too. Removing from board

PR #1000, opened 2018-01-04
added csv download for org_admin plans page

briri:
- Added a 'Download plans' button to the Org Admin -> Plans page that downloads the contents of the table as a CSV. #873
- Added tests for the org_admin/plans_controller.rb

### DIFF
--- a/app/controllers/org_admin/plans_controller.rb
+++ b/app/controllers/org_admin/plans_controller.rb
@@ -1,10 +1,10 @@
 module OrgAdmin
   class PlansController < ApplicationController
-    after_action :verify_authorized
-
+    # GET org_admin/plans
     def index
-      authorize Plan
-    
+      # Test auth directly and throw Pundit error sincePundit is unaware of namespacing
+      raise Pundit::NotAuthorizedError unless current_user.present? && current_user.can_org_admin?
+      
       vals = Role.access_values_for(:reviewer)
       @feedback_plans = Plan.joins(:roles).where('roles.user_id = ? and roles.access IN (?)', current_user.id, vals)
       @plans = current_user.org.plans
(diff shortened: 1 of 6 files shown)


briri:
will look into the failing tests tomorrow

test/functional/org_admin/plans_controller_test.rb
```diff
@@ -0,0 +1,84 @@
...
+    assert_authorized_redirect_to_plans_page
+  end
+  
+  test "org admin can access the plans page" do
+    sign_in @admin
+    get org_admin_plans_path
+
+    assert_response :success
+    assert assigns(:plans)
+    assert assigns(:feedback_plans)
```
jollopre: we should stop using assigns since enforces the test to know implementation details of that controller action moreover I think it is deprecated in Rails 5

app/controllers/org_admin/plans_controller.rb
```diff
@@ -1,10 +1,10 @@
 module OrgAdmin
   class PlansController < ApplicationController
-    after_action :verify_authorized
-
+    # GET org_admin/plans
     def index
-      authorize Plan
-    
+      # Test auth directly and throw Pundit error sincePundit is unaware of namespacing
+      raise Pundit::NotAuthorizedError unless current_user.present? && current_user.can_org_admin?
```
jollopre: Pundit is unaware of namespacing, however you can still create the class under policies/org_admin/plans_policy.rb. That means, the condition for the unless can be reused in other places too.  You would call the policy as OrgAdmin::PlansPolicy.new(current_user).index? after that unless

test/functional/org_admin/plans_controller_test.rb
briri: Yes agreed. @bhavi and have also noticed some issues with the way the tests use the seeds.rb file. It can create confusing and sometimes unpredicatable results when the tests run on Travis. They recently updated their env image, so I'm not sure if its do to that or not. We should move away from using the seed.rb and move towards Rails' fixtures or some mock data that is defined in the test_helper.rb
We need to have a team discussion about tests after MVP in March. I think it should be a priority.

Issue #920, opened 2017-12-07
Cancel "x" is misaligned on create plans page


stephaniesimms:
When I hover over the cancel "x" for Org and Template selections I can't click them. I have to move focus slightly down and to the left to perform the cancel action (the cursor becomes a gloved hand w/the message "Clear this field"). The x and tickboxes beside them seem to have shifted after Bootstrap and we lost the "or" between them. They need to be realigned w/attention to the "x" placement so I can activate it in the expected location.

See the second screenshot below for wireframe guidance:
- okay for "x" to appear outside entry field
- reinstate "or"
- shift tickboxes to right and align all elements (x, or, tickbox, text)

![screen shot 2017-12-07 at 1 24 25 pm](https://user-images.githubusercontent.com/12703638/33739615-92cc74a6-db52-11e7-83e8-cac3b453b839.png)
 
![screen shot 2017-12-07 at 1 32 01 pm](https://user-images.githubusercontent.com/12703638/33739781-3e396d26-db53-11e7-85cc-1a4fd8dc3c90.png)

stephaniesimms:
@briri - the alignment is still off. i have to hover a little above and to the right of the "x" to activate it. and if I click the x to clear the field, then enter new information the "x" disappears completely (screenshot below).

![screen shot 2018-01-04 at 11 10 11 am](https://user-images.githubusercontent.com/12703638/34580000-1329baac-f140-11e7-84e0-2e053219a5e5.png)

briri:
oh shoot. this is part of a pending PR. Moved to UAT prematurely

stephaniesimms:
looks good now @briri - all working as expected. closing issue.

Issue #396, opened 2017-06-05
jquery autocomplete box cannot handle & (ampersands)


stephaniesimms:
The new combobox has an issue with strings which contain an ampersand, when you select an Org like "Cowmilk & Cheese College" the returned string is "Cowmilk & Cheese College" which doesn't match the Org's ID and therefore is not selectable - see #270 (comment) and original issue for select2 dropdown boxes #124 (background - temp fix to change all & to "and")

stephaniesimms:
working as expected now. closing issue.

PR #1004, opened 2018-01-05
fixed '&' bug in autocomplete and fix Plan create styling

briri:
- Fixes autocomplete bug with '&' character (also now allowing single/double quotes) #396 
- Discovered and fixed timing issue for when template options are loaded
- Fixed alignment of controls on create plan page #920


briri:
Just a rebase with CDL-MVP, code was reviewed as part of #976

Issue #996, opened 2018-01-04
clicking on a tab makes the page jump to the top


briri:
Please complete the following fields as applicable:

**Expected behaviour:**
Clicking on a tab does not cause the page to jump to the top. If we need to reorient the window we should scroll to the top of a designated container/div (defaulting to the parent container)

**Actual behaviour:**
Clicking on a tab in the write plan's section's quidance/comments makes the page jump to the top. This is particularly annoying when working with sections at the bottom or if all sections happen to be expanded

stephaniesimms:
@briri I can't test this because I get the following error. the error occurs if I select any extra guidance on the Project Details page and then click on the Write Plan tab. there is no error if I leave the guidance blank and navigate to Write Plan.

![screen shot 2018-01-05 at 1 05 45 pm](https://user-images.githubusercontent.com/12703638/34628548-91739976-f219-11e7-9b89-e2d0cff55bd8.png)

briri:
will address this bug with loading the page. there is another ticket in progress #943 that involves finishing the implementation of the guidance tickboxes on the project details page.

stephaniesimms:
the tab behavior is working as expected now. but I'm having problems expanding/collapsing the accordions within tabbed guidance. one example is that when i click the + to expand the guidance next to Q2 in the screenshot below it activates/expands the same section beside Q1. regardless of which accordion I click I can only expand/collapse Q1. the same appears to be happening from one section to another. when I attempt to expand guidance in lower sections it activates this same area next to Section 1: Q1 or doesn't do anything at all.

desired behavior is to be able to expand any individual piece of guidance regardless of whether others are open or closed. 

![question-level-opening-bug](https://user-images.githubusercontent.com/12703638/34691406-5234c898-f471-11e7-9c01-9238c472d930.png)

stephaniesimms:
@briri i can't test this at the moment - there appears to be a bug w/displaying guidance in tabs. not sure if it's related to this issue or #944 (more details in that issue)

sjDCC:
This is still problematic for me. Often the second or third theme won't expand. Clicking these just changes the first (opens if closed or zips it up if open). The expand all open all in most cases but I had a legacy plan where this button wouldn't do anything. The later theme wouldn't open either.

![nerc-guidance](https://user-images.githubusercontent.com/6169940/34946535-21fafac4-f9ff-11e7-97a5-c9858e669557.PNG)

[...]

PR #1007, opened 2018-01-05
fixed bug for undefined counter

briri:
Fixes bug reported in #996


app/views/phases/_guidances_notes.html.erb
```diff
@@ -70,6 +70,7 @@
         <% guidance_accordion_id = 0 %>
         <% guidance_set.keys.each do |group| %>
```
jollopre: if we use each_with_index, we don't need to create an extra variable and keep the counter updated. For instance:
```
guidance_set.keys.each_with_index do |key, i| do
bla bla bla
end
```

briri: 👍

PR #1014, opened 2018-01-09
added +/- to section acordions in org-admin view

xsrust:
Fixes # .

Changes proposed in this PR:
- make section accordions wider, add +/- to section accordions


app/views/phases/_admin_show.html.erb
```diff
@@ -23,17 +23,21 @@
...
+            <div class="panel-title pull-left">
+              <%= section.title %>
+            </div>
+            <div class="panel-title pull-right">
+              <a
+                role="button"
+                data-toggle="collapse"
+                data-parent="#sections-accordion"
+                href="#collapseSection<%= section.id%>"
+                aria-expanded="false"
```
briri: I think we may want to retain the `section.id == current_section` logic here. 
going to merge and test, will submit a patch if needed

PR #1017, opened 2018-01-09
ran bundle update

briri:
The merge from CDL-MVP to development had some conflicts in the Gemfile and Gemfile.lock. I attempted to resolve in the merge itself but had issues running the app and tests afterward.

To resolve I replaced the files with the ones from the CDL-MVP branch. This seems to have resolved all issues. Another option would have been to run `bundle update` but this would have updated all of the underlying dependencies and could result in weird issues in the application. Felt it was best to hold off on updating all of the gems until we're at a stable MVP.


briri:
Most of the updates were minor revisions. We should keep an eye out for any weird issues that might be encountered as it could be due to these gems being upgraded. We can peg any that have issues to prior versions if necessary.

PR #1023, opened 2018-01-09
Final updates to templates page

briri:
Based on feedback in #955 
- changed all 'un-published' text to 'unpublished'
- Fixed issues with 'Published' and 'Unpublished' filter counts
- Hide Customizable templates table if the user's Org is a funder only
- Updated test

### DIFF
--- a/app/controllers/org_admin/templates_controller.rb
+++ b/app/controllers/org_admin/templates_controller.rb
@@ -501,18 +501,13 @@ module OrgAdmin
         
         scopes = calculate_table_scopes(templates, customizations)
 
-        # We scope based on the customizations if the user is NOT a super admin
+        # We scope based on the customizations
         if params[:scope].present? && params[:scope] != 'all'
-          if current_user.can_super_admin?
-            templates = templates.where(published: true) if params[:scope] == 'published'
-            templates = templates.where(published: false) if params[:scope] == 'unpublished'
-          else
-            scoped = templates.select do |t| 
-              c = customizations[t.dmptemplate_id]
-              (params[:scope] == 'unpublished' && (!c.present? || !c.published?)) || (params[:scope] == 'published' && c.present? && c.published?)
-            end
-            templates = Template.where(id: scoped.collect(&:id))
+          scoped = templates.select do |t| 
+            c = customizations[t.dmptemplate_id]
+            (params[:scope] == 'unpublished' && (!c.present? || !c.published?)) || (params[:scope] == 'published' && c.present? && c.published?)
           end
+          templates = Template.where(id: scoped.collect(&:id))
         end
 
       else
@@ -541,7 +536,7 @@ module OrgAdmin
       scopes = { all: templates.length, published: 0, unpublished: 0, dmptemplate_ids: templates.collect(&:dmptemplate_id).uniq }
       templates.each do |t|
         # If we have customizations use their status
-        if customizations.respond_to?(:keys)
+        if customizations.keys.length > 0
           c = customizations[t.dmptemplate_id]
           # If the template was not customized then its unpublished
           if c.nil?
--- a/app/models/org.rb
+++ b/app/models/org.rb
@@ -104,6 +104,9 @@ class Org < ActiveRecord::Base
       return "None"
   end
 
+  def funder_only?
+    self.org_type == Org.org_type_values_for(:funder).min
+  end
 
   ##
   # returns the name of the organisation
--- a/app/views/org_admin/templates/index.html.erb
+++ b/app/views/org_admin/templates/index.html.erb
@@ -25,60 +25,53 @@
     <ul class="nav nav-tabs" role="tablist">
       <% if current_user.can_super_admin? %>
         <li role="all-templates" class="active">
-          <a href="#all-templates" role="tab" aria-controls="all-templates" data-toggle="tab"><%= _('All Templates') %></a>
+          <a data-target="#all-templates" role="tab" aria-controls="all-templates" data-toggle="tab"><%= _('All Templates') %></a>
         </li>
       <% end %>
       <li role="organisation-templates" class="<%= !current_user.can_super_admin? ? 'active' : '' %>">
-        <a href="#organisation-templates" role="tab" aria-controls="organisation-templates" data-toggle="tab"><%= current_user.can_super_admin? ? _('%{org_name} Templates') % { org_name: current_user.org.name } : _('Own Templates') %></a>
+        <a data-target="#organisation-templates" role="tab" aria-controls="organisation-templates" data-toggle="tab"><%= current_user.can_super_admin? ? _('%{org_name} Templates') % { org_name: current_user.org.name } : _('Own Templates') %></a>
       </li>
-      <li role="funder-templates">
-        <a href="#funder-templates" role="tab" aria-controls="funder-templates" data-toggle="tab"><%= _('Customizable Templates') %></a>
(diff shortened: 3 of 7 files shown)


app/controllers/org_admin/templates_controller.rb
```diff
@@ -541,7 +536,7 @@ def calculate_table_scopes(templates, customizations)
       scopes = { all: templates.length, published: 0, unpublished: 0, dmptemplate_ids: templates.collect(&:dmptemplate_id).uniq }
       templates.each do |t|
         # If we have customizations use their status
-        if customizations.respond_to?(:keys)
+        if customizations.respond_to?(:keys) && customizations.keys.length > 0
```
jollopre: customization is defined as a hash above and does not get re-assigned so therefore .respond_to?(:keys) is not needed here. Can you remove it, please?

app/models/org.rb
```diff
@@ -104,6 +104,9 @@ def type
       return "None"
   end
 
+  def funder_only?
+    self.org_type == Org.org_type_values_for(:funder).min
```
jollopre: nice helper!

app/views/org_admin/templates/index.html.erb
```diff
@@ -25,15 +25,18 @@
...
       <% end %>
       <li role="organisation-templates" class="<%= !current_user.can_super_admin? ? 'active' : '' %>">
-        <a href="#organisation-templates" role="tab" aria-controls="organisation-templates" data-toggle="tab"><%= current_user.can_super_admin? ? _('%{org_name} Templates') % { org_name: current_user.org.name } : _('Own Templates') %></a>
-      </li>
-      <li role="funder-templates">
-        <a href="#funder-templates" role="tab" aria-controls="funder-templates" data-toggle="tab"><%= _('Customizable Templates') %></a>
+        <a data-target="#organisation-templates" role="tab" aria-controls="organisation-templates" data-toggle="tab"><%= current_user.can_super_admin? ? _('%{org_name} Templates') % { org_name: current_user.org.name } : _('Own Templates') %></a>
       </li>
+      <!-- If the Org is not just a funder then show the customizations table -->
+      <% if !current_org.funder_only? %>
```
jollopre: funder_only checks if the org_type is strictly funder however the org could be funder and something else too. I would make sure that org_type is not funder, i.e. the bit flag does not include funder type at all. What do you think?

app/views/org_admin/templates/index.html.erb
```diff
@@ -66,19 +69,21 @@
...
+          <h2><%= _('Customizable Templates') %></h2>
+          <% if customizable_templates.length > 0 %>
+            <%= paginable_renderise(
+              partial: 'funder_templates_list',
+              controller: 'org_admin/templates',
+              action: 'funders', 
+              scope: customizable_templates,
+              locals: {current_org: current_org.id, customizations: customized_templates, published: published, scopes: scopes[:funders]}) %>
+          <% else %>
+            <p><%= _('There are currently no customisable templates.') %></p>
```
jollopre: With my latest PR for paginable, the partial funder_templates_list will not be rendered if scope.length is zero, therefore you can remove the guard that checks for length as well as the info message.

app/controllers/org_admin/templates_controller.rb
```diff
@@ -501,18 +501,13 @@ def apply_scoping(scope, customizable = false, all = false)
...
-            templates = templates.where(published: false) if params[:scope] == 'unpublished'
-          else
-            scoped = templates.select do |t| 
-              c = customizations[t.dmptemplate_id]
-              (params[:scope] == 'unpublished' && (!c.present? || !c.published?)) || (params[:scope] == 'published' && c.present? && c.published?)
-            end
-            templates = Template.where(id: scoped.collect(&:id))
+          scoped = templates.select do |t| 
+            c = customizations[t.dmptemplate_id]
+            (params[:scope] == 'unpublished' && (!c.present? || !c.published?)) || (params[:scope] == 'published' && c.present? && c.published?)
```
jollopre: why do we have this boolean condition here? What are you trying to do? I don't seem to see It is assigned to a variable neither used as part of a conditional construct

test/functional/org_admin/templates_controller_test.rb
```diff
@@ -513,28 +513,32 @@ def verify_own_templates_table(user)
...
-      templates = Template.where("(org_id IN (?) OR is_default = ?) AND org_id != ?", Org.where(org_type: [2,3]).collect(&:id), true, user.org.id)
-      if user.can_org_admin?
-        templates.each do |template|
-          # Expect to only see published public templates
-          if template.publicly_visible? && template.published?
-            assert el.to_s.include?(template.title), "expected #{user.email}'s customizable table to have the funder (or default) template: '#{template.title}'" 
-          else
-            assert_not el.to_s.include?(template.title), "expected #{user.email}'s customizable table to NOT have the unpublished/non-public funder template: '#{template.title}' (from org: #{template.org.abbreviation})"
+    if user.org.funder_only?
+      assert_select "#funder-templates table tbody", 0, "expected a funder only Org to NOT see the customizable table"
```
jollopre: it is nice to see assert_select here to inspect the HTML resulting however I would place these tests as integration tests instead.

app/views/org_admin/templates/index.html.erb
briri: we actually only want to hide the tab if the Org is only a funder (so its ok to show for funder/organisations, etc.)

app/controllers/org_admin/templates_controller.rb
briri: good catch. will update

app/views/org_admin/templates/index.html.erb
briri: ok will do

app/controllers/org_admin/templates_controller.rb
briri: it determines which templates are selected. If the user clicked on the 'unpublished' filter the params[:scope] would be 'unpublished' so we only want customizations that are unpublished and vice versa when the scope is published.

app/views/org_admin/templates/index.html.erb
briri: we only want to hide the customizable templates table if the the org is just a funder, so if they were a funder+organization we want them to see it.

app/views/org_admin/templates/index.html.erb
briri: ok will remove

test/functional/org_admin/templates_controller_test.rb
briri: yes. we have a bunch of these assert_selects around. Lets move after MVP when we have a chance to reevaluate the tests.  We need to do a lot of work there

app/controllers/org_admin/templates_controller.rb
jollopre: ok, if that's the case, I would refactor params[:scope] outside of that loop. In addition, although more verbose, easier to read 6 months later.
```
if params[:scope] == 'unpublished'
  scoped = templates.select do |t|
    if customizations.has_key?(t.dmptemplate_id)
      !customizations[t.dmptemplate_id].published?
    else
      true
    end
  end
else if params[:scope] == 'published'
  scoped = templates.select do |t|
    customizations.has_key?(t.dmptemplate_id) && customizations[t.dmptemplate_id].published?
  end
else
  scoped = []
end
```

PR #1025, opened 2018-01-10
Issue 815

jollopre:
This PR includes search functionality by extending the layout view paginable used through the controller concern paginable.rb. 

I have refactored all the paginable controllers to chain the scope search, if and only if a param search is passed through. 

Note, this PR does not include anything related to sorting. I think we should create a separate issue for that since through the layout paginable we have no access to the columns to sort and therefore each partial view should be refactored accordingly.

### DIFF
--- a/app/models/plan.rb
+++ b/app/models/plan.rb
@@ -54,14 +54,24 @@ class Plan < ActiveRecord::Base
   # Note that in ActiveRecord::Enum the mappings are exposed through a class method with the pluralized attribute name (e.g visibilities rather than visibility)
   scope :publicly_visible, -> { where(:visibility => visibilities[:publicly_visible]).order(:title => :asc) }
 
+  # Retrieves any plan in which the user has an active role and it is not a reviewer
+  scope :active, -> (user) {
+    includes([:template, :roles]).where({ "roles.active": true, "roles.user_id": user.id }).where(Role.not_reviewer_condition)
+  }
+
   # Retrieves any plan organisationally or publicly visible for a given org id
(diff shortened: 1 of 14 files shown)


briri:
Makes sense @jollopre I'll create separate tickets for the sorting functionality

app/models/plan.rb
```diff
@@ -54,16 +54,26 @@ class Plan < ActiveRecord::Base
...
-    Plan.includes(:template)
+    includes([:template, :roles])
       .where({
         visibility: [visibilities[:organisationally_visible], visibilities[:publicly_visible]],
         "templates.org_id": user.org_id})
       .where(['NOT EXISTS (SELECT 1 FROM roles WHERE plan_id = plans.id AND user_id = ?)', user.id])
       .order(:title => :asc)
   }
 
+  scope :search, -> (term) {
```
briri: I like this. Its a great way to handle this functionality

app/views/shared/_search.html.erb
```diff
@@ -0,0 +1,11 @@
+<%= form_tag(url, method: :get, remote: true, class: 'form-inline') do %>
+  <div class="form-group">
+    <div class="input-group">
+      <span class="input-group-addon" id="search-addon">
+        <span class="glyphicon glyphicon-search" aria-hidden="true"></span>
```
briri: We should use fontawesome here since we're using it everywhere else. `class="fa fa-search"` should work

briri:
Just a small change @jollopre to use fontawesome instead of glyphicons for consistency

Issue #1009, opened 2018-01-08
Duplicate DOM element ids on Edit Profile page


briri:
Receiving the following error in the console on the Edit Profile page:
![screen shot 2018-01-08 at 11 57 52 am](https://user-images.githubusercontent.com/1204467/34689667-33967e64-f46b-11e7-88a3-a7075d60096d.png)

sjDCC:
Have tested this and the console tab remains empty when refreshing on firefox and chrome. Thanks @jollopre for explaining how to check. Closing out

![console-empty](https://user-images.githubusercontent.com/6169940/34942373-5ac3f0a0-f9ef-11e7-89b7-028aa7cbd1b9.PNG)

PR #1030, opened 2018-01-10
Issue1009

jollopre:
removed duplicate ids for elements. DMPRoadmap/roadmap#1009


app/views/users/_notification_preferences.html.erb
```diff
@@ -10,32 +10,27 @@
     <p class="form-control-static"><strong>All Users</strong></p>
     <div class="checkbox">
       <label for="prefs[users][new_comment]">
-        <%= hidden_field_tag 'prefs[users][new_comment]', false %>
```
briri: This does indeed remove the duplicate ids from the HTML but the page now no longer functions properly. Your changes do not save if these fields are removed. The controller would need to be updated as well

jollopre:
@briri there is a now a new JS logic in place to adding hidden fields for those checkboxes that become false. This solution unlike the previous one avoids polluting the DOM with unnecessary hidden inputs. 

Note also I have fixed two other things:
- redirect to password-details when somethings goes wrong
- updating URL hash according to the href of the tab clicked, previously #undefined due to non-existent data-target attribute for data-toggle tabs.

Issue #1010, opened 2018-01-08
Always show Edit template details button


stephaniesimms:
Please complete the following fields as applicable:

**Expected behaviour:**
When I click to Edit a template I should be taken to the Template details page and always see the button to "Edit template details" on the lower right

**Actual behaviour:**
Currently I see the button sometimes and not others (screenshot below). When it doesn't appear by default I have to click the Template details tab to make the button appear.

![edit-template-details-show](https://user-images.githubusercontent.com/12703638/34693133-6a02b92a-f477-11e7-9419-6485ba7e8968.png)
 
![screen shot 2018-01-08 at 1 27 28 pm](https://user-images.githubusercontent.com/12703638/34693251-c13b4a2c-f477-11e7-8d6c-697c237ac36d.png)

jollopre:
@stephaniesimms can you provide more info, i.e. how did you reach the status of showing "Template details" page without being the tab active? thanks

stephaniesimms:
sure @jollopre - the desired behavior is to always present this page w/the Template details tab active so that the "Edit template details" button is visible as in DMPonline production.  it's an unnecessary extra click for users to select the tab.

the current behavior is inconsistent - sometimes i see the button when first navigating to the page and sometimes i don't, and it doesn't always correspond w/the tab being selected (white instead of gray).

briri:
just a note about the two screenshots above, one has no tab selected. could be some logic associated with the current tab maybe @jollopre

sjDCC:
I've tested this on a few templates now and the button shows consistently. Closing out

jollopre:
I forgot the add the scenario that was failing:

1. Edit a template that is published but has no draft yet. In order to check if there is a draft, click in action history first to see that there are no draft currently.
2. Verify you end up on template editing for that newly draft template and the button to edit appears visible. 

Can you check if works @sjDCC? Thanks

sjDCC:
Have just published a template to check this and it all works fine.

Any actions on the template page take you back to the 'All templates' tab which is a bit annoying when you're editing the 'X (organisation) templates' or 'Customisable templates' content. It would be best if you were redirected back to the tab that was active on page refresh. I'll add this to ticket #1042

jollopre:
@sjDCC, thanks that redirection makes sense, I also noticed that too.

PR #1033, opened 2018-01-11
Issue 1010

jollopre:
fixed problem at templates#edit when a published template wants to be edited. DMPRoadmap/roadmap#1010

### DIFF
--- a/app/controllers/org_admin/templates_controller.rb
+++ b/app/controllers/org_admin/templates_controller.rb
@@ -67,30 +67,19 @@ module OrgAdmin
       @current = Template.current(@template.dmptemplate_id)
 
       if @template == @current
-        # If the template is published
         if @template.published?
           # We need to create a new, editable version
           new_version = Template.deep_copy(@template)
           new_version.version = (@template.version + 1)
           new_version.published = false
           new_version.save
-          @template = new_version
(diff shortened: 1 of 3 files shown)


app/controllers/org_admin/templates_controller.rb
```diff
@@ -66,31 +66,20 @@ def edit
...
       # If the template is published
       if @template.published?
         # We need to create a new, editable version
         new_version = Template.deep_copy(@template)
         new_version.version = (@template.version + 1)
         new_version.published = false
         new_version.save
-        @template = new_version
+        # In order to reflect the new template.id in the URL
+        redirect_to(action: 'edit', id: new_version.id)
```
briri: No. Removing this logic prevents the user from being able to view a historical version of a template (go to the history page for a template and select 'View' on an old version). The `flash[:notice]` message warns them that they are viewing a historical copy and unable to edit, but you have redirected them to the current version where they can edit.

jollopre:
@briri now this controller action should cover the multiple responsibility principle. Note I have removed the lines https://github.com/DMPRoadmap/roadmap/blob/development/app/controllers/org_admin/templates_controller.rb#L85-L92 since I can't find a case to be there and moreover can lead to create multiple copies if current template version -1  is published

briri:
thanks Jose

Issue #1012, opened 2018-01-08
Should not show Org Plans if user belongs to default other org


briri:
If a user is associated with the default `is_other = true` Org (e.g. 'Other' for DMPonline and 'Non Partner Institution' for DMPTool) they should NOT see the Organisational Plans table on their dashboard page

![screen shot 2018-01-08 at 1 30 49 pm](https://user-images.githubusercontent.com/1204467/34693439-60489b60-f478-11e7-98ce-36ba63c950cd.png)

stephaniesimms:
@briri - I'm not seeing an Org table at all now. to recreate/test I just created a plan for Univ of Edinburgh and set to Org visibility (screenshot below). when I sign in as a different user w/the same Org affiliation there's no Org table on My Dashboard.

![screen shot 2018-01-11 at 12 37 45 pm](https://user-images.githubusercontent.com/12703638/34846205-457440c2-f6cc-11e7-88b3-8f89b49114dd.png)

briri:
@stephaniesimms  I think it may be a case of both user's being collaborators on the plan. If it appears in the table at the top we currently don't show it in the table below.

I just tested on Roadmap stage with a plan that is not shared with anyone and it seems to be working. Please let me know if this plan shows up when you log in.
![screen shot 2018-01-11 at 2 40 46 pm](https://user-images.githubusercontent.com/1204467/34851173-bc836ab0-f6dd-11e7-912b-ced558dc4fe9.png)

stephaniesimms:
I see your plan @briri ! the previous test was a brand new plan w/a single owner and I'm still not sure why you can't see it in this same table. i just created another UoE plan w/a different user account, set it to Org visibility, and it's still not showing up... maybe if i try testing under a different Org...
 
![screen shot 2018-01-11 at 3 22 08 pm](https://user-images.githubusercontent.com/12703638/34852394-3f29851c-f6e3-11e7-995c-f09a2a9cd56d.png)

stephaniesimms:
working as expected now. closing issue.

PR #1039, opened 2018-01-12
Fixed issue with organisationally_publicly_visible scope

briri:
Fixes issue discovered in #1012 

The old query that retrieved the list of organisationally and publicly visible plans on the user's dashboard was incorrectly using `templates.org_id` when it should have been using the creator's `users.org_id`.

### DIFF
--- a/app/models/plan.rb
+++ b/app/models/plan.rb
@@ -61,10 +61,11 @@ class Plan < ActiveRecord::Base
 
   # Retrieves any plan organisationally or publicly visible for a given org id
   scope :organisationally_or_publicly_visible, -> (user) {
-    includes([:template, :roles])
+    includes(:template, {roles: :user})
       .where({
         visibility: [visibilities[:organisationally_visible], visibilities[:publicly_visible]],
-        "templates.org_id": user.org_id})
+        "roles.access": Role.access_values_for(:creator, :administrator, :editor, :commenter).min,
+        "users.org_id": user.org_id})
       .where(['NOT EXISTS (SELECT 1 FROM roles WHERE plan_id = plans.id AND user_id = ?)', user.id])
   }
 


app/models/plan.rb
```diff
@@ -61,10 +61,11 @@ class Plan < ActiveRecord::Base
...
-    includes([:template, :roles])
+    includes([roles: :user])
       .where({
         visibility: [visibilities[:organisationally_visible], visibilities[:publicly_visible]],
-        "templates.org_id": user.org_id})
+        "roles.access": Role.access_values_for(:creator, :administrator, :editor, :commenter).min,
+        "users.org_id": user.org_id})
       .where(['NOT EXISTS (SELECT 1 FROM roles WHERE plan_id = plans.id AND user_id = ?)', user.id])
       .order(:title => :asc)
   }
```
jollopre: - Removing template from includes means that views/paginable/plans/_organisationally_or_publicly_visible.html.erb will introduce the (n+1) problem, i.e. for each plan it will be launched a query to retrieve the title of the template.
- Removing roles from includes does not have any effect since the view is using plan.owner which is still performing yet another query (we need to improve this)
- Replacing "templates.org_id: user.org_id in favour of "roles.access... and users.org_id" means that we want to retrieve any plan in the db whose access rights are "creator, administrator, editor and commenter" if the user of that plan belongs to the user's org. Do we really want that? I was thinking we want to retrieve only plans that belong to the user org and are marked "organisationally or publicly visible"

sjDCC:
It sounds like this query is to create the two tables on the 'My Dashboard' page. The first table should be the users own plans and ones that have been specifically shared with them. The second table should be plans for which users in the same organisation have set the visibility to 'organisation' or 'public'. 

As I understand it, we aren't displaying **all** publicly visible plans on 'My Dashboard', just on the public webpage at /public_plans. Is that right @stephaniesimms? I seem to recall you found the DMPTool lists too long and didn't want to encourage people to copy and use these, rather just ones from their own org.

@jollopre it sounds like your first two queries are related to performance. Can we add templates back in to avoid additional queries being launched to render the table? Is the same needed for roles? Regarding the third bullet point, we do only want to retrieve plans that are organisationally or publicly visible from the same org, not all plans from that org. Private ones should remain private unless the user in question has a specific role on them (i.e. co-owner, editor etc) in which case they'd be shown in the first table.

Hope that help to clarify

app/models/plan.rb
briri: - I will add template back to the `includes` list. The issue was specifically with the `template.org_id` in the where clause. That was causing it to only show plans that are derived from templates owned by the user's org. We want plans that were created by users belonging to the user's org regardless of what template was used to create the plan. For example If we both belong to UoE and I create a Horizon 2020 (and make it organisationally visible) you should see it when you log in. Right now, no one would see it because the template belongs to Horizon2020
- The code is already running the plan.owner query before this proposed change. I'm not attempting to refactor the entire process, just address the one bug
- Yes, that is what we want to show. Plans have no direct association to Orgs so we have to go through the Roles table.

briri:
The adjusted query returns all publicly/organisationally visible plans for the current user's organization (meaning that the creator of the plan belongs to the current user's org). It was only showing plans that were created from templates that belonged to the org before. No changes were made to the first 'My plans' table.

stephaniesimms:
@briri @jollopre @sjDCC - it sounds like all is sorted now, but just confirming the requirements for the Org Plans table in My Dashboard. It should display all plans from the user's affiliated Org that are marked w/Org or Public visibility. Private plans are only displayed to their plan owners and collaborators in the main My plans table. The complete list of public plans is only available on the public page, correct.

briri:
That is correct @stephaniesimms

Issue #1026, opened 2018-01-10
Add search to the templates page


briri:
Add the search functionality to the templates model and then move the `all`, `orgs` and `funders` methods from the org_admin/templates_controller.rb over to the paginable namespace.

Refer to the plans tables.

stephaniesimms:
i can search all the templates. closing issue.

PR #1041, opened 2018-01-12
moved template pagination logic to paginable namespace

briri:
Moved old pagination logic from the org_admin/templates_controller.rb over to the paginable/templates_controller.rb #1026

Also added shared template filter methods to a concern.


app/controllers/concerns/template_filter.rb
```diff
@@ -0,0 +1,88 @@
+module TemplateFilter
+  extend ActiveSupport::Concern
+  
+  included do
+    # Applies scoping to the template list
+    def apply_scoping(scope, customizable = false, all = false)
```
jollopre: It is nice to see apply_scoping and calculate_table_scopes refactored outside the templates_controller and templates_model, however a Rails Concern is meant to be as generic as possible and these functionalities are tight to templates only. I would suggest to create a template_filter module under lib like follows:
```
module TemplateFilter
  def self.apply_scoping
  ...
  end
  def self.calculate_table_scopes
  ...
  end
```
and after under models/template.rb
```
include TemplateFilter
```
which that change, you will know that those methods belong to the Template model and are refactored in another file since you will call Template.apply_scoping instead of apply_scoping directly on Paginable::TemplatesController

briri:
Closing to refactor in the new sort functionality and correct the failing tests

PR #1048, opened 2018-01-16
Changed links datatype to TEXT for old MySQL compatibility - #729.

vyruss:
**Requires db:migrate**

For #729 .

Changes proposed in this PR:
- Changed links datatype to TEXT for old MySQL compatibility


briri:
Thanks @vyruss. @bhavi though has created a more complete script that updates the old varchar columns to use the new limit in #1053

PR #1053, opened 2018-01-16
Sql script to update the varchar(255) to latest 65k limit

bhavi:
Fixes # .

Changes proposed in this PR:
-


vyruss:
@bhavi I'm not sure this is the right solution at this moment: 
- This will break compatibility with old MySQL versions that don't support these limits
- There's no need to have 65535 character limits e.g. for surnames (it has the potential to waste space in tables: https://stackoverflow.com/a/5899369/6234472
- We should keep database changes within db migrations
Can you please hold off on this for now, so we can discuss on Friday?

PR #1054, opened 2018-01-16
Moved template pagination logic

briri:
- Moved pagination logic to the paginable namespace and updated to make use of new search/sort functionality #1026
- Moved template helper methods from concern to a lib/filter


briri:
this is a resubmission of #1041

Issue #794, opened 2017-11-02
Edit User privileges as a Super Admin


stephaniesimms:
UX wireframe @rhu001 "DMP User Accounts (Super Admin View)" 
![screen shot 2017-11-02 at 1 20 41 pm](https://user-images.githubusercontent.com/12703638/32348348-b76ff802-bfd0-11e7-8162-cb7c3c264fb6.png)

DMP Edit User Role - modal 
![screen shot 2017-11-02 at 12 57 06 pm](https://user-images.githubusercontent.com/12703638/32348221-52ae222c-bfd0-11e7-8e2a-356e160abc56.png)

- all functionality should be the same as that implemented in #761 except if a user has Super Admin privileges the modal will contain additional items as in the second screenshot above 
- Blue notepad functional spec: Clicking on Edit link should open a modal that allows admins to assign privileges to other users. For example, I'm a Super Admin and should see a list of all users in the User Accounts table. If I click the Edit link I'm presented with the modal below and option to assign any privileges to any user.
- Ticking the first box for "Org admin privileges" should automatically tick all boxes below it; these can then be deselected individually if desired. The same behavior applies if a user ticks the second box for "Super admin privileges"
- Note: add one more privilege tickbox at the end of the Org Admin column to "Use API" (missing from wireframe)

For clarity, here are the descriptors currently used in the database, grouped by the two roles; these should be mapped to the language in the modal screenshot above:

**Org admin**
Grant_permissions
Modify_templates
Modify_guidance
Change_org_details
Use_api

**Super admin**
Add_organisation
Change_org_affiliation
Grant_api_to_orgs

sjDCC:
For clarity, here are the descriptors currently used in the database, grouped by the two roles:

**Org admin**
Grant_permissions
Modify_templates
Modify_guidance
Change_org_details
Use_api

**Super admin**
Add_organisation
Change_org_affiliation
Grant_api_to_orgs

stephaniesimms:
some things are fixed (layout for modal and only reloading current page w/user displayed). but the behavior is not quite right. 

- As a Super Admin when I click to Edit privileges then select the top level "Org Admin" box to check all org admin privileges and click Save it reloads the page but doesn't display a message or reflect the changes in the user table (screenshot 1). 
- When I click to Edit the same user's privileges I see the modal with some Org admin boxes unchecked (it's always the top level and the bottom box).
[...]

Issue #761, opened 2017-10-20
Revise Edit user privileges functionality


stephaniesimms:
UX wireframe @rhu001 "DMP User Accounts (Admin View)" 
Additional privileges are visible to Super Admin users (related issue #794)
For example, an Org Admin at UCSF should see a list of UCSF users in the table. When the admin clicks the Edit link she's presented with the modal in the screenshot below and can assign Org admin privileges to any user. 

![screen shot 2017-10-30 at 12 42 48 pm](https://user-images.githubusercontent.com/12703638/32191942-f8d8f6d0-bd6f-11e7-845d-d9185b6ea4ee.png)

DMP Edit User Role - modal 
![screen shot 2017-11-02 at 12 55 59 pm](https://user-images.githubusercontent.com/12703638/32347800-0cf2f90c-bfcf-11e7-8560-209cde6c5d9c.png)

- Update page heading to "User accounts"
- Add search box above table (this will be done as part of #815)
- Table functional spec: limit visibility to 10 rows, thereafter implement pagination (also #815)
- Update column names as in wireframe
- Add column for "Current privileges" before "Edit" column
- For display purposes in the "Current privileges" column there are two buckets: 1) Org admin privileges, 2) Super admin privileges. If a user has any individual privileges in either of these buckets, then display the high-level bucket name in this column. So, e.g., if I can "Manage user templates" then this column should say that I have "Org admin privileges." If I can also "Grant API access" then this column should also say that I have "Super admin privileges"
- For users with no Org or Super admin privileges, the "Current privileges" column should be blank
- Blue notepad functional spec: Clicking on Edit link should open a modal that allows admins to assign privileges to other users (second UX wireframe screenshot above)
- Ticking the first box for "Org admin privileges" should automatically tick all boxes below it; these can then be deselected individually if desired
- Note: add one more privilege tickbox at the end of the Org Admin column to "Use API" (missing from wireframe)

For clarity, here are the descriptors currently used in the database, grouped by the two roles; these should be mapped to the language in the modal screenshot above:

**Org admin**
Grant_permissions
Modify_templates
Modify_guidance
Change_org_details
Use_api

stephaniesimms:
main functionality is working as expected @bhavi - i just have some minor changes to request:
- after user clicks either the "Save" or "Back" button on the modal please reload the main User Accounts page to the "View All" expanded list. current behavior takes me back to the paginated list and I have to click View All, then search for the user to confirm the changes to privileges.
[...]

PR #1059, opened 2018-01-17
Edit User privileges as a Super Admin (#794) and Org Admin (#761)

bhavi:
Fixes # .

Changes proposed in this PR:
-


app/views/paginable/users/_index.html.erb
```diff
@@ -1,43 +1,78 @@
...
-                <%= link_to b_label, admin_grant_permissions_user_path(user)%>
-            <% end %>
-          </td>
-        </tr>
-      <% end %>
-    <% end %>
-  </tbody>
-</table>
\ No newline at end of file
+<div class="row">
```
jollopre: This partial view get rendered within the paginable layout therefore you only have to write code here that display your data in this case the table. There are several things to change here:
- Please, remove lines 1 to 11. The header of this view should be modified at app/views/users/admin_index.html.erb which is the view rendered when users#admin_index action is requested.
- Please, remove lines 15 to 23 since the search functionality is rendered within the paginable layout and this views stays DRI.
- The header columns must call paginable_sort_link helper in order to enable the sorting functionality passing the table column as an string argument (see other examples at app/views/paginable/**/*.html.erb.

app/views/paginable/users/_index.html.erb
```diff
@@ -1,43 +1,78 @@
...
+                      <%= user.email %>
+                  </td>
+                  <td class="text-center">
+                      <% if !user.last_sign_in_at.nil? %>
+                      <%= l user.last_sign_in_at.to_date, :formats => :short %>
+                      <% end %>
+                  </td>
+                  <td class="text-center">
+                      <% if !user.roles.nil? %>
+                      <%= user.roles.length %>
```
jollopre: you might need to remove from the count those roles in which the user is a reviewer...

app/views/users/_admin_grant_permissions.html.erb
```diff
@@ -0,0 +1,72 @@
...
+                <span aria-hidden="true">&times;</span></button>
+        <h1 class="modal-title"><%= _('Edit User Privileges') %></h1>
+      </div>
+      
+      <%= form_tag( admin_update_permissions_user_path(@user), method: :put) do %>
+        <div class="modal-body" style="text-align: left">
+          <p><%= @user.name(false) %><%= _(' current user privileges are: ') %></p>
+
+          <ul class="list-group">
+            <% if current_user.can_super_admin? %>
```
jollopre: According to the snapshot from #794 first are listed org_admin privileges, so please swap the order

lib/assets/javascripts/views/users/admin_grant_permissions.js
```diff
@@ -0,0 +1,65 @@
...
+    }
+  };
+
+  const error = () => {
+    // There was an ajax error so just route the user to the sign-in modal
+    // and let them sign in as a Non-Partner Institution
+    $('a[data-target="#modal-permissions"]').tab('show');
+  };
+
+  $(paginableSelector).on('click', '.modal-window', (e) => {
```
jollopre: We are using unobtrusive JavaScript across the site which means $.ajax is abstracted for us. I think is worth to be consistent for any new requirement coming. Can you refactor from L21-31 using delegated events for success and error. You can see a similar example at lib/assets/javascripts/views/plans/index.js

lib/assets/stylesheets/overrides.scss
```diff
@@ -142,6 +142,14 @@ span.sublist {
   margin-right: 25px;
 }
 
+ul {
```
jollopre: This might break other places using ul purely. I would be more inclined to use an specific selector instead.

briri:
Thanks for making the recommended changes Bhavi

Issue #1027, opened 2018-01-10
Add sort functionality for all tables


briri:
Add sorting functionality to all of the tables. The tables used to have a JS solution for this. Use the old 'sorter-false' class to exclude columns from being sortable. For example we don't need/want the plans table's 'Actions' column to be sortable.

stephaniesimms:
@briri - I still can't sort most tables in the Admin menu (Templates, Guidance, Plans). I can sort the Users table, but not by the "Name" field. should I create a new issue to implement sort across all tables?

jollopre:
The sorting functionality is generic and it has been refactored so far for:
- Plans (my dashboard)
- Themes
- Users

As you noticed, some of the fields do not have the sorting icons since they are not directly related to a table column:
- name in users table (which involves sorting by more than one field in the database)
- role in private plans (which is an integer on the db but text on the table displayed)
- visibility in private plans (which is either test or private) and since we have the check box to determine if the plan is marked as test or not, it seems redundant field to be displayed.
- shared in private plans (does not exist a field in the db to be sorted)
- owner in organisationally/publicly plans which can be either firstname and surname or email in the users table.

I can try to come up with something to accommodate those fields but I can't guarantee to be generic and reusable across every table we display.

stephaniesimms:
thanks for the explanation re sort-ability @jollopre! It would be great if you can enable sorting by (First) Name in the users table. that's the only additional field I think we would want, but pinging @sjDCC to think on it too...

sjDCC:
Agnostic about whether the name sort is by firstname or surname. Happy to go with firstname

stephaniesimms:
I'm also agnostic. Just suggested first name because that's what comes first in the field. @jollopre - please implement sorting by First name in the Users table.

jollopre:
@stephaniesimms, PR #1062 introduces sorting mechanism for name column. The order is by users firstname on the users table.

stephaniesimms:
this looks great. closing issue.

sjDCC:
Agree - looks great. Also like how it sorts all data (i.e. you get the first 10 names beginning with A from all users not just a reordering of the first page view).

PR #1062, opened 2018-01-18
sort by firstname for list of users name column. DMPRoadmap/roadmap#1027

jollopre:


briri:
thanks for adding the tests

Issue #817, opened 2017-11-14
Date filtering error


stephaniesimms:
Filtering the plans table on My Dashboard by the last "Edited" date field produces weird outcomes (screenshots below). Make sure date format is set correctly. 

![date-sort-error](https://user-images.githubusercontent.com/12703638/32756074-1b3fa2c8-c88d-11e7-83ae-4984fc4b0f92.png)

![date-sort-error2](https://user-images.githubusercontent.com/12703638/32756077-1e5288f4-c88d-11e7-9a7b-e617f0176a70.png)

briri:
Should be fixed as part of #815

jollopre:
should be fixed as part of #1027

stephaniesimms:
this is working now. closing issue.

PR #1071, opened 2018-01-19
Updated Plans table to use plans.updated_at for edited dates

briri:
Was using `plans.last_updated` which reflects the time the template's phase was updated. Switched to use `plans.updated_at` and verified that localization is working (e.g. 12/15/2017 - US, 15/12/2017 - UK) #817

### DIFF
--- a/app/models/answer.rb
+++ b/app/models/answer.rb
@@ -1,5 +1,11 @@
 class Answer < ActiveRecord::Base
 
+  after_save do |answer|
+    # Updates the plan.updated_at attribute whenever an answer has been created/updated 
+    # check first for a nil plan so that the controller can properly handle the ActiveRecord::NotFound exception
+    answer.plan.touch unless answer.plan.nil? 
+  end
+  
   ##
   # Associations
 	belongs_to :question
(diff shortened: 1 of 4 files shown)


jollopre:
@briri, the issue talks about my Dashboard but you just changed the list of plans that every org admin can see within each organisation. Please, update also the files:

- views/paginable/plans/_organisationally_or_publicly_visible.html.erb
- views/paginable/plans/_privately_visible.html.erb

Also, as we are changing to display the updated_at from a plan means we will need any sort of callback on phase update to trigger touch in plan updated_at, otherwise this date field will only be updated when we change any details of the plan and I believe that's not enough.

briri:
yes that's a good point @jollopre I will add a callback to the answer model

briri:
Added the callback @jollopre. I found that without a nil check in the callback that an exception was thrown during the integration tests. When an answer is created without an associated plan the controller was not able to catch the RecordNotFound exception properly. 

Its very unlikely that this scenario would occur but it might be a good idea to add similar checks to the callbacks on phase, section, question as well

Issue #1067, opened 2018-01-19
Templates History needs pagination


jollopre:
Please complete the following fields as applicable:

**Expected behaviour:**
The template history should be using the paginable concern too.
**Actual behaviour:**

**Steps to reproduce:**

sjDCC:
This does work but creating versions to test it made me realise we should invert the order by default so the most recent version is on top

![capture](https://user-images.githubusercontent.com/6169940/35280580-8779e6ae-0048-11e8-8cfd-cfe743e41865.PNG)

stephaniesimms:
most recent "Draft" template appears at the top of the list now. all seems to be working as expected. closing issue.

jollopre:
When template history page is server, the sorting mechanism is explicitly passed to the records (see https://github.com/DMPRoadmap/roadmap/blob/development/app/controllers/org_admin/templates_controller.rb#L181). In contrast, when clicking in any link handled by the paginable concern, we don't have a default sorting mechanism in place for the scope model we are dealing with. 

Since ticket #1105 is going to work on sorting enhancements, I think is more appropriate to add @stephaniesimms suggestion as part of the same ticket. 

Please, feel free to close this ticket since it will be solved generically.

stephaniesimms:
ok @jollopre - closing this issue. global sorting enhancements covered in #1105

PR #1073, opened 2018-01-22
Issue 1067

jollopre:
templates history paginablerised. DMPRoadmap/roadmap#1067


jollopre:
I have introduced an after_save callback whenever a phase gets created/updated to trigger touch at its template. To be accurate, we should also include callback at section to touch the phase and callback at question to touch the question...

briri:
yes that seems like a better approach @jollopre

Issue #1042, opened 2018-01-15
Template visibility settings incorrectly displaying on institutional templates 


sjDCC:
We added template visibility settings under 'edit template details' but my understanding was that these were only to apply to funder templates, not all templates. It was added for the USGS use case where the funder has public and internal-only templates.

This is what is currently seen. 
![template-visibility](https://user-images.githubusercontent.com/6169940/34941453-b0f2a984-f9eb-11e7-9482-37322dd67222.PNG)

The guidance and tooltip don't make much sense as we only list funder templates on the public 'DMP Templates' page, not all institutional templates. We also don't want to make institutional templates private or internal-only which checking this box would indicate. If a user from another organisation wants to use the Edinburgh template (e.g. if a collaborative project was being led from there) they should be able to access the Edinburgh template, despite not belonging to that org

I also noticed that any actions on the template page take you back to the 'All templates' tab which is a bit annoying when you're editing the 'X (organisation) templates' or 'Customisable templates' content. It would be best if you were redirected back to the tab that was active on page refresh. 

![capture](https://user-images.githubusercontent.com/6169940/34945968-f42e6a2e-f9fc-11e7-8687-4d342707c51a.PNG)

stephaniesimms:
the tickbox only appears on org/funder templates (e.g. USGS). not on institutional templates. i think this is working as expected now. closing issue but leaving on board for @sjDCC to verify.

sjDCC:
The institutional templates don't have the tickbox any longer which is great. Thanks @briri 

The 'view all templates' button still takes you back to the original table view though with the 'All templates' tab selected. It would be better if this worked like a back button (as stated above) so people who were editing institutional templates or customisable templates are taken back there instead. I found I kept having an extra click when I was editing uni templates, which was annoying.

I have reopened the ticket to do this part of it, but if that action is not feasible for some reason, close it out and remove it from board please.

stephaniesimms:
just noting here that @briri confirmed he will address new request to make "View all templates" function as a back button as part of this issue. moving back to in progress.

briri:
once deployed this should fix the issue with the 'view all templates' button not remembering what tab the user was on

stephaniesimms:
working as expected now. closing issue.

sjDCC:
Looks great - thanks @briri Removing from board

PR #1075, opened 2018-01-22
Fix bugs on template pages

briri:
- Updated template show/edit so that 'visibility' is only available to Orgs that are a funder AND some other org type (e.g. USGS is funder+organisation) #1042 
- Update 'View all templates' buttons so that they return to the correct template tab #1042 
- Discovered an issue with the way Published/Unpublished was displaying on org templates tab

### DIFF
--- a/app/controllers/org_admin/templates_controller.rb
+++ b/app/controllers/org_admin/templates_controller.rb
@@ -38,6 +38,7 @@ module OrgAdmin
     # -----------------------------------------------------
     def new
       authorize Template
+      @current_tab = params[:r] || 'all-templates'
     end
     
     # POST /org_admin/templates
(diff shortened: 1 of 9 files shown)


app/controllers/org_admin/templates_controller.rb
```diff
@@ -151,19 +156,19 @@ def destroy
         if current == @template
           if @template.destroy
             flash[:notice] = success_message(_('template'), _('removed'))
-            redirect_to org_admin_templates_path
+            redirect_to org_admin_templates_path(r: current_tab)
           else
             @hash = @template.to_hash
             flash[:alert] = failed_destroy_error(@template, _('template'))
-            render org_admin_templates_path
+            render org_admin_templates_path(r: current_tab)
```
jollopre: I'm not aware of render expecting a URL as argument. Is this a copy/paste thing?

app/controllers/org_admin/templates_controller.rb
```diff
@@ -38,6 +38,7 @@ def index
     # -----------------------------------------------------
     def new
       authorize Template
+      @current_tab = params[:r] || 'all-templates'
```
jollopre: @current_tab could be refactored into a private method since it seems to be repeated across different actions.
```
def current_tab=(value=params[:r])
  @current_tab =value || 'all-templates'
end
```

app/controllers/org_admin/templates_controller.rb
briri: yep good catch. should just be doing a redirect like the others

PR #1076, opened 2018-01-23
Change long strings to text in DB

vyruss:
**DB Migration** for #729 .

Changes proposed in this PR:
- As discussed, switch strings that we know are long (JSON, URLs) to TEXT type in DB


briri:
looks good Jimmy.

We can institute a practice of making sure any `limit: #` are cleared out in future PRs from anyone running msyql. Shouldn't be a big issue since changes to schema.rb are infrequent at this point in the project

Issue #943, opened 2017-12-13
Finish Project Details page - including Guidance section


stephaniesimms:
@briri on steps to complete #491. Alright, I reviewed the ticket and here's what's left to do:

- Fix spacing between the 'mock/test project' checkbox and the input elements above and below it. It currently encroaches on 'Funder'
- Make sure that only the first word in the field labels is capitalized (except for 'ORCID iD' which is the official name for that)
- Finish implementing the guidance section. Note that it may be difficult to introduce a typeahead style autocomplete box like we use for Orgs since this list deals with a combination of Org name and Guidance Name (e.g. University of Edinburgh: Roslin Institute vs University of Somewhere Else: Somewhere Else guidance). If it proves to be very difficult I would recommend coming back to the team with a recommendation on an alternative implementation (e.g. a modal window that lists out all of the possibilities in several columns.
- Please also move the migrate:plan_data_contacts rake task to the bugfix.rake rake task (maybe also rename that file to 'upgrade.rake') and add it as a task that gets run for v1.0. This extracts phone numbers and emails out of the old DMP contact info field and places them in the appropriate new columns and would need to be run by anyone upgrading.

Functional specs for Guidance section (blue notepad in screenshot):
- This list should always include:
[x] plan funder
[x] plan organization
[  ] user's organization if different from plan
[x] DCC/CDL generic guidance
- Plan funder should be selected and disabled– users can't turn it off.

![screen shot 2017-12-12 at 7 27 23 pm](https://user-images.githubusercontent.com/12703638/33920389-87238670-df72-11e7-911c-6192d48dc913.png)

briri:
@bhavi Please disregard the instruction above that says 'Plan funder should be selected and disabled– users can't turn it off.'
The write plan page always shows question level guidance from the funder (aka annotations). The user cannot toggle those on or off. They should be able to uncheck the funder from the list on this page because it will allow them to turn off the themed guidance from the funder.

stephaniesimms:
changes are looking good. two things i noticed that need attention:
- there's no restriction on the number of guidance boxes i can tick. restrict total guidance selections to 6 per the instructions.
[...]

PR #1077, opened 2018-01-23
Guidance section in Plan edit page(#943)

bhavi:
- reimplement the guidance selection functionality on the project details page #943
- all possible guidance selections are now only loaded on-demand when the user opens the modal
- add functional test for the select guidance list modal


briri:
Looks good Bhavi

PR #1084, opened 2018-01-25
Fix Shibboleth logins

briri:
The application_controller.rb was redirecting the user to `request.referer` after login. This creates issues when the user is authenticating via Shibboleth. The user logs in and this code sends them back to the IdP resulting in a stale SAML request.

Updated the application_controller.rb to only use `request.referer` if the referer matches the domain of the site


jollopre:
it looks fine to me.

Issue #905, opened 2017-12-05
Text editor is huge when I click to edit existing content


stephaniesimms:
When I click to edit the contents of any TinyMCE field in the system that already contains text, the editor opens up to fill ca. 3 page lengths (screenshot 1). When I place focus inside the editor, it shrinks to a normal size (screenshot 2). This happens when editing any existing content: plans, templates, the Request feedback email on the Org details page as in the example below.

Desired behavior is to enable editing with a normal sized editor - as in screenshot 2.

![screen shot 2017-12-05 at 2 22 44 pm](https://user-images.githubusercontent.com/12703638/33634606-9a358e38-d9c9-11e7-970b-a9d7eedc9948.png)

![screen shot 2017-12-05 at 2 23 02 pm](https://user-images.githubusercontent.com/12703638/33634615-9f0b0abe-d9c9-11e7-9927-1b201dd5b306.png)

stephaniesimms:
this appears to be fixed now. closing issue but i'll leave it here for @sjDCC to confirm.

sjDCC:
I haven't noticed any on a bit of playing around either so agree it's fixed. Removing from board

PR #1086, opened 2018-01-26
Issue 905

jollopre:
autoresizing tinymce to min_height defined in settings. DMPRoadmap/roadmap#905


briri:
excellent and thanks for digging through the Tinymce documentation 😁

Issue #1081, opened 2018-01-25
Add expand all/collapse all links to Template editing & customizing


stephaniesimms:
For consistency and usability, add the links to expand all/collapse all accordions on the Template editing and customizing pages. We have them for plan sections and guidance. It would be helpful for admins to have them on the template pages too to expand all sections if they edit their own templates or if they want to customize a funder template (see screenshots).

1 - Links on write plan pages
![screen shot 2018-01-24 at 4 16 52 pm](https://user-images.githubusercontent.com/12703638/35364074-0f2c2aec-0122-11e8-880d-41b969cb0441.png)

2 - Add links to Template editing page just above the first accordion (e.g., Univ of Edinburgh template)
![template-editing](https://user-images.githubusercontent.com/12703638/35364183-940ba79c-0122-11e8-959e-d53ae8146eb1.png)

3 - Add links to Customize template page just above the first accordion (e.g., for each phase of the NERC template)
![template-customize](https://user-images.githubusercontent.com/12703638/35364210-b52cb7e0-0122-11e8-9f44-b8511d904857.png)

stephaniesimms:
looks great and working as expected. closing issue.

PR #1088, opened 2018-01-26
Issue 1081

jollopre:
- expand/collapse links for template show phases and phases show/edit sections. Introduced panel default for the container that displays these views. DMPRoadmap/roadmap#1081
- phase preview content within panel. DMPRoadmap/roadmap#1081


briri:
looks ok to me

PR #1090, opened 2018-01-26
Revise  Publish action on Guidances page (#871)

bhavi:
Getting the tests to pass
Unpublish the gudiance group if none of its guidances are published
Update tooltip on Guidance Group name

Fixes # .

Changes proposed in this PR:
-


briri:
looks ok @bhavi

Issue #1056, opened 2018-01-17
Create Super Admin page to manage Orgs


stephaniesimms:
Add a page "Organizations" in the Admin menu for users with Super Admin privileges (screenshot 1). Super Admins need to be able to see a complete list of Orgs, create new Orgs, and manage basic info. When a user clicks the "Create Organization" button they should be taken to the Org details page (screenshot 2) where they can add/save a new Org. 

This same Org details page can then be edited by users from that org w/Org Admin privileges. But the Org Configuration details (Shib) section at the bottom will be grayed out and read only for Org Admins (only Super Admins can edit this info).

Make sure to implement same pagination and search box logic if the list exceeds 10 records #1055 

Two notes though based on how it currently works in DMPonline:
1. The org type should display textual categories (i.e. 'funder' 'institution') to select from. Currently we just have codes for each 0, 1 etc
2. The only mandatory field on creating an org should be 'name'. Previously you were forced to enter a contact email to save changes but this may have been removed now.

1 - Organizations page
![screen shot 2018-01-17 at 7 19 59 am](https://user-images.githubusercontent.com/12703638/35050103-df99f442-fb56-11e7-9a4d-c0c3aea6aeb8.png)

2 - Org details page
![screen shot 2018-01-17 at 7 26 10 am](https://user-images.githubusercontent.com/12703638/35050531-e88e313e-fb57-11e7-82ed-9112747dee8d.png)

sjDCC:
Sounds like it captures the main functionality to me. Two notes though based on how it currently works:
1. The org type should display textual categories (i.e. 'funder' 'institution') to select from. Currently we just have codes for each 0, 1 etc
2. The only mandatory field on creating an org should be 'name'. Previously you were forced to enter a contact email to save changes but this may have been removed now...

stephaniesimms:
looking good @briri - some things i noticed while testing that need attention:

- when creating a new Org 3 fields should be required and marked with an asterisk: Org full name, Org abbreviated name, and at least one Org type (e.g. Institution). Currently I can create an Org w/only the full name.
- I tested on both roadmap and stage and like that I can select two Org types for, e.g. USGS. but the behavior is strange here - if i tick two boxes (e.g. Funder, Org) and click Save it only saves one selection. but if I tick all three boxes and click Save it saves 2 selections. desired behavior is to allow Super Admins to make a selection of Funder, Org and save this on the first attempt.
[...]

PR #1096, opened 2018-01-27
added super admin org pages

briri:
Added new Super Admin Orgs page along with the ability to create Orgs, update the org types and shib info and delete orgs that have no associated users/templates. #1056

### DIFF
--- a/app/models/org.rb
+++ b/app/models/org.rb
@@ -25,7 +25,7 @@ class Org < ActiveRecord::Base
 
   has_many :org_identifiers
   has_many :identifier_schemes, through: :org_identifiers
-
+  
   ##
   # Possibly needed for active_admin
   #   -relies on protected_attributes gem as syntax depricated in rails 4.2
@@ -38,7 +38,7 @@ class Org < ActiveRecord::Base
                   :feedback_enabled, :feedback_email_subject, :feedback_email_msg
   ##
   # Validators
-  validates :contact_email, email: true, allow_nil: true
+#  validates :contact_email, email: true, allow_nil: true
   validates :name, presence: {message: _("can't be blank")}, uniqueness: {message: _("must be unique")}
   # allow validations for logo upload
   dragonfly_accessor :logo do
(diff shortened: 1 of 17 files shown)


app/models/org.rb
```diff
@@ -88,20 +93,14 @@ def get_locale
...
-      return "Organisation"
-    elsif self.research_institute?
-      return "Research Institute"
-    elsif self.project?
-      return "Project"
-    elsif self.school?
-      return "School"
-    end
-      return "None"
+    ret = []
```
jollopre: @briri, this method will conflict with the ActiveRecord type method used for Single Table Inheritance and we might have side effects. Can you rename it to something like "org_type_to_s"?

app/policies/org_policy.rb
```diff
@@ -6,19 +6,33 @@ def initialize(user, org)
...
   end
-
+  
   def admin_show?
     user.can_modify_org_details? && (user.org_id == org.id)
   end
 
   def admin_edit?
-    user.can_modify_org_details? && (user.org_id == org.id)
+puts "WTF!? #{user.can_modify_org_details?} #{user.org_id} == #{org.id}"
```
jollopre: please remove this puts for here

app/views/paginable/orgs/_index.html.erb
```diff
@@ -0,0 +1,50 @@
...
+        <th class="sorter-false"><span aria-hidden="false" class="sr-only"><%= _('Actions') %></span></th>
+      </tr>
+    </thead>
+    <tbody>
+      <% scope.each do |org| %>
+        <tr>
+          <td><%= org.name %></td>
+          <td><%= org.contact_email %></td>
+          <td>
+            <% types = [] %>
```
jollopre: please, can you use the method type you defined at org.rb?

app/models/org.rb
```diff
@@ -38,7 +38,7 @@ class Org < ActiveRecord::Base
                   :feedback_enabled, :feedback_email_subject, :feedback_email_msg
   ##
   # Validators
-  validates :contact_email, email: true, allow_nil: true
+#  validates :contact_email, email: true, allow_nil: true
```
jollopre: what's the case to remove this validator?

briri:
will make the changes and submit in another PR @jollopre

app/models/org.rb
briri: I was asked to change the contact email so that it isn't required in the ticket. 

This `email: true` part though was causing it to be required even though we had `allow_nil: true`. I'll try switching the order so that its `allow_nil: true, email: true` perhaps that will work. We validate the email in the JS side so its not critical although it would be good to retain it here too

app/models/org.rb
briri: yes good catch

PR #1121, opened 2018-02-05
Revert to template versioning on template details page

briri:
- Switched versioning back to the template edit page. I left the versioning logic in the template model though.
- Removed check for other user editing same phase - `Rails.cache` function

### DIFF
--- a/app/controllers/org_admin/templates_controller.rb
+++ b/app/controllers/org_admin/templates_controller.rb
@@ -68,7 +68,17 @@ module OrgAdmin
 
       @current = Template.current(@template.dmptemplate_id)
 
-      unless @template == @current
+      if @template == @current 
+        if @template.published?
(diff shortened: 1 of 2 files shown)


app/controllers/org_admin/templates_controller.rb
```diff
@@ -68,7 +68,17 @@ def edit
 
       @current = Template.current(@template.dmptemplate_id)
 
-      unless @template == @current
+      if @template == @current 
+        if @template.published?
+          new_version = @template.get_new_version
+          if !new_version.nil?
```
jollopre: just checked the method get_new_version... In an event of failing an exception will be raised (e.g. RecordInvalid, RecordNotSaved) which is not caught here and therefore this controller action will fail badly. Can you add an error handler here or remove save! in favour of save at get_new_version model, please?

Issue #1116, opened 2018-02-01
Published Org templates not available when I create a new plan


stephaniesimms:
Please complete the following fields as applicable:

**Expected behaviour:**
All of the "Published" Org templates should be available to users when creating a new plan (screenshots 1-2 below). 

**Actual behaviour:**
As an Org Admin, when I edit templates and publish the changes I receive a success message that the templates are published. But when I try to create a plan the templates appear and disappear from the dropdown list seemingly at random. Only 2 of the 6 published templates for Univ of Edinburgh are available (screenshots below), e.g.

I noticed this while testing a different bug #1063 that is now fixed - unsure if they're related

1 - Templates table shows 6 published templates for UoE
![screen shot 2018-02-01 at 9 39 10 am](https://user-images.githubusercontent.com/12703638/35693881-6d52add6-0734-11e8-9d6b-58476c5736f0.png)

2 - only 2 of 6 published UoE templates appear in the dropdown list
![screen shot 2018-02-01 at 9 38 35 am](https://user-images.githubusercontent.com/12703638/35693839-4edb05ba-0734-11e8-971e-bc1fd620b818.png)

stephaniesimms:
this is working as expected now, except for Univ of Edinburgh templates. I assume it's just bad data. will close and leave here for @sjDCC to confirm all good.

sjDCC:
Reopening this as I've done some testing with Glasgow and it still seems to be wrong.

Glasgow had 3 unpublished templates so I published 2 to test this and all three showed up in dropdown. The text notes that these are templates from the funder too, whereas it should say organisation.

![picture1](https://user-images.githubusercontent.com/6169940/36149853-47d17e1e-10b9-11e8-9837-c1c1835c4670.png)

![templates](https://user-images.githubusercontent.com/6169940/36149858-4d7cf744-10b9-11e8-9726-34b5e1c496d1.PNG)

sjDCC:
I don't know if this is linked to the logic of which templates are shown in the dropdown, but on doing testing I noticed that a newly published and then re-edited version of an Edinburgh template didn't show up. Perhaps it's because there's a fault in the logic on which templates are published and should be shown? It may be worth checking that published templates with an unpublished new version (e.g. unpublished changes) do still show. The earlier published version should remain active unless specifically unpublished by the admin.

briri:
[...]

PR #1124, opened 2018-02-05
fixed create plan logic

briri:
Fixes issue with template selection during create plan process #1116

### DIFF
--- a/app/controllers/org_admin/templates_controller.rb
+++ b/app/controllers/org_admin/templates_controller.rb
@@ -376,7 +376,7 @@ module OrgAdmin
           # Load the funder's template(s)
           templates = Template.valid.publicly_visible.where(published: true, org_id: funder_id).to_a
 
-          if org_id.present?
+          unless org_id.blank?
             # Swap out any organisational cusotmizations of a funder template
             templates.each do |tmplt|
               customization = Template.valid.find_by(published: true, org_id: org_id, customization_of: tmplt.dmptemplate_id)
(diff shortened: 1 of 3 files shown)


briri:
fixed broken test

app/controllers/org_admin/templates_controller.rb
```diff
@@ -376,7 +376,7 @@ def template_options()
           # Load the funder's template(s)
           templates = Template.valid.publicly_visible.where(published: true, org_id: funder_id).to_a
 
-          if org_id.present?
+          unless org_id.blank?
```
jollopre: if funder_is is not blank (L375) there is not need to check if org_id is not blank (L379) since you will never enter in the if (L374).

app/controllers/org_admin/templates_controller.rb
```diff
@@ -386,20 +386,15 @@ def template_options()
               end
             end
           end
+        else
+          # If no funder was selected retrieve the Org's templates
+          templates << Template.organisationally_visible.valid.where(published: true, org_id: org_id, customization_of: nil).to_a
```
jollopre: if org_id is blank, this else is executed too. Is there any problem for that query scope?

app/controllers/org_admin/templates_controller.rb
jollopre: I don't see the point of having the if statement (L374) when you have two conditional within it to check funder_id.present? and after org_id.present?.
- present? is the negation of blank?
- blank can be empty string, empty string with spaces, [], {}, false and nil.

Anyway, it works but just to note it.

PR #1130, opened 2018-02-06
Fix org links banner for #729

vyruss:
For #729.

Be aware, testing `object.present?` for empty lists will return true so it is not exactly the opposite of `.blank?`


briri:
looks good @vyruss

PR #1131, opened 2018-02-07
Make entire accordion bars clickable

briri:
- changes to HTML to make entire accordion title bar clickable
- Fix alignment of text on guidance selection section of project details page


briri:
Ah yes. good catch @jollopre

briri:
updated JS and the +/- toggle is working

PR #1133, opened 2018-02-07
fix display issue org links

briri:
The org details page was always displaying the 'Add an additional URL' link even if the limit had been reached. This patch will hide/show the link based on the current number of links displayed on the screen.

### DIFF
--- a/app/views/shared/_links.html.erb
+++ b/app/views/shared/_links.html.erb
@@ -44,7 +44,7 @@
   </div>
   <div class="row">
     <div class="col-xs-12">
-      <a href="#" class="new"><%= _('+ Add an additional URL') %></a>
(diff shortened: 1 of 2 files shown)


jollopre:
Just want to share another approach. This would be an event-based one:

```
$('.links').on('set_visibility', '.new', (e) => {
    const target = $(e.target);
    const max = maxNumberLinks(target);
    const nbrLinks = linksLength(target);
    if (nbrLinks === max) {
      target.addClass('hide'); // Hide when maximum is reached
    } else if (nbrLinks === max - 1) {
      target.removeClass('hide'); // Show when nbrlinks is max - 1
    }
  });
```

And you would trigger this custom event anytime a link is added/removed from the DOM which the following line:
```
target.trigger('set_visibility');
```
where target is the link:
```
<a href="#" class="new"><%= _('+ Add an additional URL') %></a>
```

briri:
interesting. I haven't done much with custom events. I like the approach

PR #1136, opened 2018-02-08
fixed sort issue for orgs page

briri:
addresses sorting issue on new Orgs page. An order was being set on the default scope which was conflicting with the sort selected by the user. #1056

### DIFF
--- a/app/controllers/paginable/orgs_controller.rb
+++ b/app/controllers/paginable/orgs_controller.rb
@@ -1,13 +1,11 @@
-module Paginable 
-  class OrgsController < ApplicationController
-    include Paginable
-    # /paginable/guidances/index/:page
-    def index
-      authorize(Org)
-      paginable_renderise(
-        partial: 'index',
-        scope: Org.includes(:templates, :users).all
-      )
-    end
+class Paginable::OrgsController < ApplicationController
+  include Paginable
+  # /paginable/guidances/index/:page
+  def index
+    authorize(Org)
+    paginable_renderise(
+      partial: 'index',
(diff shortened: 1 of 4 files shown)


app/views/paginable/orgs/_index.html.erb
```diff
@@ -5,7 +5,7 @@
         <th><%= _('Organisation') %>&nbsp;<%= paginable_sort_link('orgs.name') %></th>
         <th><%= _('Administrator Email') %>&nbsp;<%= paginable_sort_link('orgs.contact_email') %></th>
         <th><%= _('Organisation Type(s)') %>&nbsp;<%= paginable_sort_link('orgs.org_type') %></th>
-        <th><%= _('Templates') %>&nbsp;<%= paginable_sort_link('orgs.updated_at') %></th>
+        <th><%= _('Templates') %></th>
```
jollopre: good you removed this sortable field. Sorting by updated_at when the field name was Templates seems odd

app/controllers/paginable/orgs_controller.rb
```diff
@@ -1,13 +1,11 @@
...
-    include Paginable
-    # /paginable/guidances/index/:page
-    def index
-      authorize(Org)
-      paginable_renderise(
-        partial: 'index',
-        scope: Org.includes(:templates, :users).all
-      )
-    end
+class Paginable::OrgsController < ApplicationController
```
jollopre: This is more a style preference. We should introduce static code analyser at some point to keep consistency

Issue #944, opened 2017-12-13
Revise Guidance on Write Plan page - updated


stephaniesimms:
UX wireframes @JEK-III : http://h90c3m.axshare.com/#g=1&p=write_plan
- one guidance/comments portion per question, but refactoring the UI to be similar to the wireframe linked above and screenshot below
- Use a slider to switch between guidance and comments and create guidance tabs by organisation
- Related issue started as #494 and continued in #943 

Additional notes from @briri (please confirm that I captured everything relevant):
- All guidance sections should be collapsed on page load or a new section is expanded
- If a guidance area was expanded and is still available to the user, it should remain expanded
- The current tab guidance/comments should remain active (e.g. if the user had the comments tab open and clicked on a different answer then the comments tab should remain open)

![screen shot 2017-12-12 at 7 36 36 pm](https://user-images.githubusercontent.com/12703638/33920665-f891e346-df73-11e7-84e8-37aa9919428b.png)

briri:
Implemented according to screenshot above but using current tab color/styling:

![screen shot 2018-01-03 at 1 52 13 pm](https://user-images.githubusercontent.com/1204467/34541547-75111304-f08d-11e7-9b86-1eb14fe4563c.png)

The tabs display the GuidanceGroup name, so we will have to ensure that we curate our existing group names. I was going to use the Org abbreviation but that wouldn't work for orgs that have multiple guidance groups (e.g. Univ. of Edinburgh and Roslin Inst.)

briri:
question for both @sjDCC and @stephaniesimms:

How should we differentiate between question level guidance and themed funder guidance? The first 'DCC' tab here contains the question level guidance (entered on the template edit page). The second 'DCC' tab is themed guidance provided by the funder. Note that DCC is the template owner/funder in this example
![screen shot 2018-01-03 at 2 45 18 pm](https://user-images.githubusercontent.com/1204467/34543317-1f5278b0-f095-11e7-998c-7cbb12dc6f8b.png)
![screen shot 2018-01-03 at 2 45 29 pm](https://user-images.githubusercontent.com/1204467/34543318-1f69a5d0-f095-11e7-82ec-9368f0555d21.png)

We could always use a hard-coded value like 'Template' instead of the Org name for the first tab (assuming there is question level guidance available).

briri:
spoke with @stephaniesimms and we think the bold org name above the tabs should be removed. Its a bit redundant since the user can see which tab they've selected.

[...]

PR #1137, opened 2018-02-08
Fixes issues with Guidance selection and display

briri:
- Placement of the new `<div class="panel panel-default"><div class="panel-body">` broke the connection between the tab menu and content. #944 (change should also correct #996)
- Updated guidance selection to address issues reported in #943


briri:
will address issues with tests tomorrow

lib/assets/javascripts/views/plans/edit_details.js
```diff
@@ -20,4 +21,31 @@ $(() => {
     showHideDataContact($(e.currentTarget));
   });
   showHideDataContact($('#show_data_contact'));
+
+  // Keep the modal window's guidance selections in line with selections on the main page
+  const syncGuidance = () => {
+    const choices = $('#priority-guidance-orgs, #other-guidance-orgs').find('input[type="checkbox"]:checked')
```
jollopre: Since other-guidance-orgs is a modal you could have simplified the logic, i.e. you could have added a function handler for both containers (e.g. priority-guidance-orgs, other-guidance-orgs). Therefore if would be not needed:
- filter to remove duplicates (L29)
- Finding checkbox would be limited to the other container selector only, either #priority-guidance-orgs or #other-guidance-orgs.

Also, I would be very annoyed to have automatically disabled all my guidances if I select more than 6. I would go for a more event-based approach, perhaps custom event trigger when the MAX_NUMBER_GUIDANCE_SECTIONS is reached...

Issue #1102, opened 2018-01-30
Adjust create plan logic to present correct templates


stephaniesimms:
- If a user selects a Funder for question 3 then they should only be presented with funder templates (in most cases there is only one funder template). Current behavior includes Org templates in a dropdown menu - these should only be displayed if user ticks the box that no funder is associated. 

![default-template-logic](https://user-images.githubusercontent.com/12703638/35573440-06797cf6-058c-11e8-9e26-7a9de049f9e0.png)

sjDCC:
Part 2 of this logic is incorrect @stephaniesimms The default template does not appear in the dropdown for selection. It is only given when no funder and no org template applies.

stephaniesimms:
thanks @sjDCC - i updated the issue accordingly and pinged @briri who diagrammed it out and plans to fix this one

stephaniesimms:
all template logic seems to be working now except for one thing @briri :
- if I select USGS as my Org AND as my funder I should see a dropdown that contains both internal and the public template. current behavior only displays the public template.

![screen shot 2018-02-08 at 5 06 01 pm](https://user-images.githubusercontent.com/12703638/36006589-b16191d6-0cf2-11e8-9461-29d253dbd99b.png)

briri:
Ok. New finalized logic for the create plan template selection code:

Logic:
- Default template is used if 
  - User ticks both the 'no research org' and 'no funder' boxes 
  - User selects a Research Org that has NO templates and ticks the 'no funder' box
  - User selects a Research Org that has NO templates and a funder that has NO templates
  - User ticks the 'no Research Org' box and selects a funder that has NO templates
  - User selects a Research Org that has template(s) and a funder who has NO templates
- The research org template is used (or user is presented with list if more than one is available)
  - User selects a Research Org that has template(s) and ticks the 'no funder' box
- The funder template us used (or user is presented with list if more than one is available)
  - User ticks the 'no research org' box and selects a funder who has template(s)
  - User selects a Research Org that has NO templates and a funder who has template(s)
  - User selects a Research Org that has template(s) and a funder who has template(s)
- The user is presented with a list of both organisational and public funder templates (USGS edge case)
  - The user selects the same Org in both the Research Org and Funder boxes and that org has templates that are public and organisational

stephaniesimms:
[...]

PR #1141, opened 2018-02-09
Update create plan template selection

briri:
#1102 
adjust create plan logic so that if user selects the same Org that is both a funder and an institution/organisation (USGS edge case) as both the research org and the funder they see the internal and public templates for that org. Doing the same in every other situation only shows the selected funder's public templates.

Logic:
- Default template is used if 
  - User ticks both the 'no research org' and 'no funder' boxes 
  - User selects a Research Org that has NO templates and ticks the 'no funder' box
  - User selects a Research Org that has NO templates and a funder that has NO templates
  - User ticks the 'no Research Org' box and selects a funder that has NO templates
  - User selects a Research Org that has template(s) and a funder who has NO templates
- The research org template is used (or user is presented with list if more than one is available)
  - User selects a Research Org that has template(s) and ticks the 'no funder' box
- The funder template us used (or user is presented with list if more than one is available)
  - User ticks the 'no research org' box and selects a funder who has template(s)
  - User selects a Research Org that has NO templates and a funder who has template(s)
  - User selects a Research Org that has template(s) and a funder who has template(s)
- The user is presented with a list of both organisational and public funder templates (USGS edge case)
  - The user selects the same Org in both the Research Org and Funder boxes and that org has templates that are public and organisational

### DIFF
--- a/app/controllers/org_admin/templates_controller.rb
+++ b/app/controllers/org_admin/templates_controller.rb
@@ -387,8 +387,11 @@ module OrgAdmin
               end
             end
           end
-        else
-          # If no funder was selected retrieve the Org's templates
+        end
+        
+        # If the no funder was specified OR the funder matches the org
+        if funder_id.blank? || funder_id == org_id
+          # Retrieve the Org's templates
           templates << Template.organisationally_visible.valid.where(published: true, org_id: org_id, customization_of: nil).to_a
         end
         


app/controllers/org_admin/templates_controller.rb
```diff
@@ -387,8 +387,11 @@ def template_options()
               end
             end
           end
-        else
-          # If no funder was selected retrieve the Org's templates
+        end
+        
+        # If the no funder was specified OR the funder matches the org
+        if funder_id.blank? || funder_id == org_id
```
jollopre: we still can end up passing a blank org_id to the scope method if for any reason the params is not present...

Issue #1127, opened 2018-02-06
Error exporting to PDF


sjDCC:
I get the following error exporting to PDF. I thought it may be due to legacy data, but occurs on new plans too.
![capture](https://user-images.githubusercontent.com/6169940/35861543-02986110-0b40-11e8-8621-053f3e5ebce5.PNG)

Other export / download formats seem to work fine

briri:
FYI @stephaniesimms and @sjDCC. The docx, pdf and html download options are all working. The csv and text export options have a minor issue. I will submit the fix for them as a separate PR. Please take a look at the PDF, DocX and HTML options and let me know if they need any adjustments.
They share the same code now (each one had its own logic before) so it should be easier to maintain going forward.

stephaniesimms:
all of the formats are giving me the same error msg when i try to download plans from the "Download" tab. i can download public plan PDFs only.

![screen shot 2018-02-14 at 9 02 41 pm](https://user-images.githubusercontent.com/12703638/36241390-e0682f8a-11ca-11e8-9187-1780fdb60dca.png)

briri:
submitted a fix for this error message and also updated the public plan version so that it does not show the question text.

stephaniesimms:
i can download all file formats successfully now. i noticed a few weird formatting things w/DOCX but it may just be weird/test data. will wait for external testers to point out major issues. closing this one.

sjDCC:
The various downloads I've done have all been ok too. I wasn't scientific in checking each in turn, but agree users will flag if anything is amiss. Removing from board.

Issue #1126, opened 2018-02-06
Error message on downloading PDFs of shared institutional plans


sjDCC:
Downloading the PDF of plans presented in the shared institutional plan table under 'My Dashboard' sometimes throws a "You are not authorised to perform this action" error.

Perhaps there's a permission issue or something has changed in the visibility of these plans and wrong ones are being presented in the table? 

It only happens on some entries not all

![error](https://user-images.githubusercontent.com/6169940/35859594-a44c0ef6-0b38-11e8-9093-a787f3aaeabd.PNG)

stephaniesimms:
i tested by creating new plans (in case this is a product of bad data) but it still gives me the Error msg (see screenshot below). it seems I can't download any Org plans.

![screen shot 2018-02-14 at 8 57 19 pm](https://user-images.githubusercontent.com/12703638/36241211-ac6b5956-11c9-11e8-8cef-bf3550f51116.png)

stephaniesimms:
seems to be fixed now. i got one error but assume it's bad data. all newer plans are working. closing issue.

PR #1143, opened 2018-02-09
consolidated plan exports

briri:
Consolidated public plans and logged in plan exports to address issues in #1127 and #1126 
- moved export functionality to model concern
- moved plan view code to shared/export


jollopre:
merging the PR, @sjDCC and @stephaniesimms will test all the exporting options for each exporting type of plans.

PR #1150, opened 2018-02-13
Update the fix csv and txt exports

briri:
goes along with #1126 and #1127


app/controllers/plans_controller.rb
```diff
@@ -246,9 +246,9 @@ def export
     file_name = @plan.title.gsub(/ /, "_")
 
     respond_to do |format|
```
jollopre: I believe respond_to whitelists the formats passed to the block so if we don't want to display html, we can just remove the line format.html...

PR #1151, opened 2018-02-13
fixed issue with capitalization of guidance groups

briri:
#944 guidance group and org names were losing capitalization on the select guidance options


app/views/plans/_guidance_choices.html.erb
```diff
@@ -3,23 +3,18 @@
...
-      <label class="inline checkbox-label regular-text guidance-group-label">
-        <%= org.name %>
-      </label>
-
+      <%= org.name %>
       <ul>
         <% groups.each do |group| %>
           <li class="checkbox">
-            <span class="sublist">└─ </span> 
+            <span class="sublist">└─ </span>
```
jollopre: we can get rid of this symbol if we adopt the proposal from @vyruss raised yesterday.

Issue #1117, opened 2018-02-01
Adjust tickboxes for Sign In 


stephaniesimms:
Somewhere along the way the tickboxes got jumbled for Sign In and Create Account. Please make the following change to reflect current DMPonline production for Sign In:

- Sign In page/modal should only have 1 tickbox for "Remember email" and it should be displayed below the pw field (where it currently says Show pw). Remove "Show password" from Sign In. It should only appear on Create account.

1 - Roadmap Sign In
![screen shot 2018-02-01 at 1 17 02 pm](https://user-images.githubusercontent.com/12703638/35703878-80c7c5b8-0752-11e8-9137-f7c66b18529d.png)

2 - DMPonline Sign In
![screen shot 2018-02-01 at 1 20 14 pm](https://user-images.githubusercontent.com/12703638/35703946-b0de9074-0752-11e8-8058-80829fac5355.png)

stephaniesimms:
@briri - please move the Forgot pw? link above the Remember email tickbox as in DMPonline screenshot above.

![screen shot 2018-02-14 at 8 47 17 pm](https://user-images.githubusercontent.com/12703638/36240989-6514c354-11c8-11e8-93aa-2da9dae228c3.png)

stephaniesimms:
looks good. closing issue.

PR #1152, opened 2018-02-13
remove 'show pasword' from sign in form

briri:
#1117 removes 'show password' checkbox from signin form

### DIFF
--- a/app/views/shared/_sign_in_form.html.erb
+++ b/app/views/shared/_sign_in_form.html.erb
@@ -3,19 +3,13 @@
     <%= f.label(:email, _('Email'), class: 'control-label') %>
     <%= f.email_field(:email, class: 'form-control', "aria-required": true) %>
   </div>
-  <div class="checkbox">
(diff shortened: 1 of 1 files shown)


app/views/shared/_sign_in_form.html.erb
```diff
@@ -3,19 +3,13 @@
...
-  </div>
   <div class="form-group">
     <%= f.label(:password, _('Password'), class: 'control-label') %>
     <%= f.password_field(:password, class: 'form-control', "aria-required": true) %>
   </div>
   <div class="checkbox">
-    <label for="passwords_toggle">
-      <input type="checkbox" id="passwords_toggle" /><%= _('Show password') %>
-    </label>
+    <%= label_tag 'remember_email', raw("#{check_box_tag 'remember_email'}#{_('Remember email')}") %>
```
jollopre: since show password is not longer applicable here. We should also clean lib/assets/javascripts/views/shared/sign_in_form.js (L3 and L7)

Issue #1147, opened 2018-02-13
Add/revise descriptive text for plans tables on My Dashboard


stephaniesimms:
If a user creates a new account w/an Org affiliation and enters the tool for the first time, they may see a table w/Org plans on My Dashboard. We need to add something about what these plans are  just above the table to orient users. We should also revise the text slightly for the main plans table for consistent UI language:

Proposed text for main plans table and Org plans table @sjDCC :
"The table below lists the plans that you have created, and that have been shared with you by others. You can edit, share, download, make a copy, or remove these plans at any time."

"The table below lists the plans that users at your organisation have created and shared within your organisation. This allows you to download a PDF and view their plans as samples or to discover new research data."

![screen shot 2018-02-12 at 4 06 48 pm](https://user-images.githubusercontent.com/12703638/36127228-2a8117fa-1011-11e8-86b1-e5be53f2c333.png)

stephaniesimms:
looks good. closing issue.

PR #1155, opened 2018-02-13
added new text to my dashboard

briri:
#1147 
Updated text on My Dashboard


app/views/plans/index.html.erb
```diff
@@ -4,10 +4,9 @@
     <!-- if the user has projects -->
     <p>
       <% if @plans.count > 0 %>
```
jollopre: not sure when this got merged but we should get the habit of using length instead of count to avoid an extra query

PR #1169, opened 2018-02-15
change org.links default to empty hash

raycarrick-ed:
Fixes # .
bug encountered during setting up dev env.

Changes proposed in this PR:
- change default for orgs.links from empty array to empty hash

### DIFF
--- a/db/migrate/20180215141222_change_org_links_default.rb
+++ b/db/migrate/20180215141222_change_org_links_default.rb
@@ -1,0 +1,8 @@
+class ChangeOrgLinksDefault < ActiveRecord::Migration
+  def up
+    change_column_default :orgs, :links, '{}' 
+  end
+  def down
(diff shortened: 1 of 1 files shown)


db/migrate/20180215141222_change_org_links_default.rb
```diff
@@ -0,0 +1,8 @@
+class ChangeOrgLinksDefault < ActiveRecord::Migration
+  def up
+    change_column_default :orgs, :links, '{}'
```
jollopre: the validation is preventing to change that value for the orgs (see https://github.com/DMPRoadmap/roadmap/blob/development/app/models/org.rb#L5). You should pass { org: []} instead.

vyruss:
closing as it will be dealt with in the migrations of PR #1273

PR #1170, opened 2018-02-15
fixes for downloads

briri:
addresses issues in #1127 
- removed `.joins(:answers)` since we cannot guarantee that the plan has at least one answer before user tries to download. 
- also added check for blank answer before trying to access its options
- updated policy for organisationally visible plan download so that it compares the plan owner's org instead of the template creator's org #1126

### DIFF
--- a/app/views/shared/export/_plan.erb
+++ b/app/views/shared/export/_plan.erb
@@ -25,7 +25,7 @@
 
         <% section[:questions].each do |question| %>
           <div class="question">
-            <% if @show_sections_questions %>
+            <% if @show_sections_questions && !@public_plan %>
               <%# text in this case is an array to accomodate for option_based %>
               <% if question[:text].length > 1 %>
                 <ul>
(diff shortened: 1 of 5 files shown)


app/views/shared/export/_plan.erb
```diff
@@ -43,7 +43,7 @@
             <% blank = answer.present? ? answer.text.gsub(/<\/?p>/, '').gsub(/<br\s?\/?>/, '').chomp.blank? : true %>
```
jollopre: using answer.is_valid? would help to simplify this context. we should address this later.

app/views/shared/export/_plan.erb
```diff
@@ -43,7 +43,7 @@
...
             <% if blank && @show_unanswered %>
               <p><%= _('Question not answered.') -%></p>
-            <% else %>
+            <% elsif !blank %>
```
jollopre: rails helper method .present? would be a clearer solution instead of the baked variable blank we have defined above...

briri: it would be but I'm seeing a lot of `<p></p>` stored in the DB. I'm thinking that this is happening when the user blanks out the answer. That value always equates to present? true

jollopre: yep! If that's the case we should look at tinymce to clear out everything so the answer text becomes '' and therefore .present? is false

app/views/shared/export/_plan.erb
briri: ah yeah. adding the code to ignore the html to this is_valid would be better.
Could probably make use of the Nokogiri gem to load the answer.text value into html and then do a check on that html object's `.text()` to make sure its ignoring all markup when checking the answer

PR #1174, opened 2018-02-16
Issue 1144

jollopre:
Create account org is needed now. Fixed find nil whne swaping orgs.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -125,13 +125,14 @@ class User < ActiveRecord::Base
   # @param new_organisation_id [Integer] the id for an organisation
   # @return [String] the empty string as a causality of setting api_token
   def org_id=(new_org_id)
+    puts "$$$$ new_org_id: #{new_org_id}"
     unless self.can_change_org? || new_org_id.nil? || self.org.nil? || (new_org_id.to_s == self.org.id.to_s)
       # rip all permissions from the user
       self.perms.delete_all
     end
     # set the user's new organisation
     super(new_org_id)
-    self.save!
+    # self.save!
     # rip api permissions from the user
     self.remove_token!
   end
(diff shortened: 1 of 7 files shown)


app/models/user.rb
```diff
@@ -125,13 +125,14 @@ def identifier_for(scheme)
   # @param new_organisation_id [Integer] the id for an organisation
   # @return [String] the empty string as a causality of setting api_token
   def org_id=(new_org_id)
+    puts "$$$$ new_org_id: #{new_org_id}"
```
briri: remove this debug statement

app/models/user.rb
```diff
@@ -125,13 +125,14 @@ def identifier_for(scheme)
...
     unless self.can_change_org? || new_org_id.nil? || self.org.nil? || (new_org_id.to_s == self.org.id.to_s)
       # rip all permissions from the user
       self.perms.delete_all
     end
     # set the user's new organisation
     super(new_org_id)
-    self.save!
+    # self.save!
```
briri: delete the line. we have git ;)

lib/assets/javascripts/views/shared/create_account_form.js
```diff
@@ -1,25 +1,22 @@
 import initAutoComplete from '../../utils/autoComplete';
 import ariatiseForm from '../../utils/ariatiseForm';
 import { togglisePasswords } from '../../utils/passwordHelper';
-import { SHOW_SELECT_ORG_MESSAGE, SHOW_OTHER_ORG_MESSAGE } from '../../constants';
```
briri: We should remove those constants from the file if we don't need them anymore

app/models/user.rb
jollopre: :)

Issue #1105, opened 2018-01-30
Cannot sort columns when tables are expanded


stephaniesimms:
Paginated tables appear as expected, alphabetized by the first word in the first column. But if I click "View All" to expand the list the ordering gets shuffled (screenshot below). Then if I try to sort any of the columns, the table reverts back to a paginated view.

Desired behavior is to retain the original alphabetized order when a user clicks "View All". In addition a user should be able to sort columns while viewing the expanded list. There are only a few tables where the number of records is so large that this action will be slow (e.g. Super Admin tables of users or plans). We need to fix this to improve usability for the majority of users.

![screen shot 2018-01-30 at 10 24 49 am](https://user-images.githubusercontent.com/12703638/35583850-0caa15f6-05a8-11e8-8213-8862c7e0e273.png)

sjDCC:
Happy for this to be done and if we find it has a big performance hit, we can rethink later. 

The org users and plans tables are going to get quite long too so there may be more non-super-admin cases than we first expect.

jollopre:
This ticket will address:
- Retain View All (if enabled) when sorting is requested
- Add default sorting to paginable concern if no sorting request is made

stephaniesimms:
sorting is working beautifully for all tables now. a word of caution for @sjDCC - as a Super Admin we can't 'View All' to expand the lists for Users and Orgs (these queries overload things and blow up the server). I'll create a post MVP issue to handle them separately.

sjDCC:
Noted!

PR #1177, opened 2018-02-20
Issue1105

jollopre:
The default sorting has to be passed explicitly to paginable_renderise concern method through query_params keyword. For example:
paginable_renderise(..., query_params{ sort_field: :version, sort_direction: :asc }, ...) 

Note that query_params passed to the method are merged into the params object from the controller that included the concern and they act as a fallback ONLY for when no sort is requested through URL explicitly.

Summary of changes applied:

- page ALL is retained if latest request contains it. Sort links are now right aligned within the th. #1105
- default sorting to paginable concern specified through query_params keyword. #1105
- paginable plans including default sorting
- paginable orgs including default sorting. #1105

### DIFF
--- a/app/controllers/concerns/paginable.rb
+++ b/app/controllers/concerns/paginable.rb
@@ -7,15 +7,18 @@ module Paginable
     # controller {String} - Represents the name of the controller to handles the pagination
     # action {String} - Represents the method name within the controller
     # path_params {Hash} - A hash of additional URL path parameters (e.g. path_paths = { id: 'foo' } for /paginable/templates/:id/history/:page)
+    # query_params {Hash} - A hash of query parameters used to merge with params object from the controller for which this concern is included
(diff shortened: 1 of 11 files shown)


app/controllers/concerns/paginable.rb
```diff
@@ -86,20 +97,19 @@ def sort_link_name(sort_field)
       if direction.present? && params[:sort_field] == sort_field
         className = direction == 'ASC'? 'fa-sort-asc' : 'fa-sort-desc'
       end
-      return raw("<i class=\"fa #{className}\" aria-hidden=\"true\"></i>")
+      return raw("<i class=\"fa #{className}\" aria-hidden=\"true\" style=\"float: right; font-size: 1.2em;\"></i>")
```
briri: ok for now, but this should be in the scss

Issue #1181, opened 2018-02-21
Accordion won't open on Template details page


stephaniesimms:
When I click the +/- to expand the accordion for the phase on the main template details page ('Data management plan' in screenshot below) the + and - changes but the accordion doesn't open.

![screen shot 2018-02-20 at 7 37 05 pm](https://user-images.githubusercontent.com/12703638/36462012-b978c402-1675-11e8-9d7b-16d19372764c.png)

stephaniesimms:
@briri working now - thanks! if not too much trouble can you load the page w/accordion expanded by default? i think this was a request @sjDCC made in the templates issue sam was working on.

Issue #1180, opened 2018-02-21
'View all templates' button reloads wrong tab


stephaniesimms:
Please complete the following fields as applicable:

**Expected behaviour:**
When I click the button to 'View all templates' after editing any template for my Org (located in the second tab on main templates page) I would expect to be taken back to to my Org templates tab. Likewise, after customizing a template and clicking 'View all templates' I would expect to go back to the Customizable templates tab. The 'View all templates' button should work as a back button

**Actual behaviour:**
Current behavior when I click 'View all templates' (as a Super Admin) is to take me back to the last active tab (UCOP tab is white in screenshot below) but the table displays 'All templates.' I have to select a different tab and then reselect the UCOP Templates tab in order to see the correct list of templates. This is true for Org Admins as well - clicking 'View all templates' displays the Own templates list regardless of which tab is highlighted.

1 - Actual behavior when I click 'View all templates' after editing a UCOP/Org template
![screen shot 2018-02-20 at 7 20 25 pm](https://user-images.githubusercontent.com/12703638/36461825-8ebf3d0a-1674-11e8-9ed3-5577ea4bf955.png)

2 - Desired behavior when I click 'View all templates' after editing a UCOP/Org template
![screen shot 2018-02-20 at 7 23 26 pm](https://user-images.githubusercontent.com/12703638/36461851-b0756ab4-1674-11e8-852f-93bcc42fcdc9.png)

stephaniesimms:
this is working as expected for Super Admins - clicking 'View all templates' takes me back to the appropriate tabs.

As an Org Admin it's doing something weird when I click 'View all templates' after customizing a template. The button takes me back to a page with Own templates and Customizable templates tabs stacked (screenshot below). This only happens for Org Admins and only for the Customizable templates tab.

![screen shot 2018-02-22 at 9 55 52 am](https://user-images.githubusercontent.com/12703638/36555635-6fb72e2e-17b7-11e8-8b2b-52ec65186d5f.png)

stephaniesimms:
almost fixed. the only problem now is that as an Org Admin (only) when I first navigate to the Templates page I see blank tabs with no table content (screenshot 1). If I click either tab, then the tables appear (screenshot 2). all other behavior working as expected. please just fix the behavior to display the table content when a user first navigates to the Templates page.

[...]

PR #1189, opened 2018-02-21
fixed template pages

briri:
#1180 and #1181
- fix issue with phase accordions not opening on main template page
- fix issue with current templates table tab not being retained when editing a template


briri:
I agree. I meant it as a temporary solution for now. We should ideally refactor it along with an update of those template editing pages so that we're not loading everything all at once.

Issue #1187, opened 2018-02-21
Apply default sorting for tables


stephaniesimms:
Apply default sorting order for tables. Retain this default sorting when user clicks 'View All' to expand the tables.
- on My Dashboard: default sorting should be by plan 'Last edited' date, most recent first
- all other tables: default sorting should be alphabetical by first field (Template Name, User Name, etc)

stephaniesimms:
default sorting looks good. it's not retained when I click 'View All' to expand the tables so I have to click to sort again. will create another issue for this to address post MVP. closing this one.

PR #1190, opened 2018-02-21
updated tables so they have default sorts & fix search/sort retention

briri:
#1187 
- Apply new params local to all tables for default initial sort
- Update paginable so that params are always pulled in and attached to the base url
- Update paginable so that sort rules are retained when searching or clearing search
- Fixed tests

### DIFF
--- a/app/controllers/org_admin/plans_controller.rb
+++ b/app/controllers/org_admin/plans_controller.rb
@@ -43,7 +43,7 @@ module OrgAdmin
       
       plans = CSV.generate do |csv|
         csv << header_cols
-        org.plans.includes(template: :org).each do |plan|
+        org.plans.includes(template: :org).order(updated_at: :desc).each do |plan|
           owner = plan.owner
           csv << [
             "#{plan.title}", 
--- a/app/views/guidances/admin_index.html.erb
+++ b/app/views/guidances/admin_index.html.erb
@@ -17,7 +17,8 @@
       partial: '/paginable/guidance_groups/index',
       controller: 'paginable/guidance_groups',
       action: 'index',
-      scope: @guidance_groups) %>
+      scope: @guidance_groups,
+      query_params: { sort_field: 'guidance_groups.name', sort_direction: :asc }) %>
     <div>
       <a href="<%= admin_new_guidance_group_path %>" class="btn btn-primary"><%= _('Create a guidance group') %></a>
     </div>
@@ -36,7 +37,8 @@
         partial: '/paginable/guidances/index',
         controller: 'paginable/guidances',
         action: 'index',
-        scope: @guidances) %>
+        scope: @guidances,
+        query_params: { sort_field: 'guidances.text', sort_direction: :asc }) %>
   
     <div>
       <a href="<%= admin_new_guidance_path %>" class="btn btn-primary"><%= _('Create guidance') %></a>
--- a/app/views/org_admin/plans/index.html.erb
+++ b/app/views/org_admin/plans/index.html.erb
@@ -39,7 +39,8 @@
         partial: '/paginable/plans/org_admin',
         controller: 'paginable/plans',
         action: 'org_admin',
-        scope: @plans) %>
+        scope: @plans,
+        query_params: { sort_field: 'plans.updated_at', sort_direction: :desc }) %>
     <% end %>
   </div>
 </div>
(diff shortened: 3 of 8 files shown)


lib/assets/javascripts/utils/paginable.js
```diff
@@ -7,7 +7,13 @@ $(() => {
     // Note we update ONLY .paginable-results container although the AJAX response
     // returns the entire paginable layout html. That permits, keeping any value entered
     // in the search box (if any)
-    $(e.target).closest(paginableSelector).find('.paginable-results').html($(data).find('.paginable-results').children());
+    const paginable = $(e.target).closest(paginableSelector);
+    paginable.find('.paginable-results').html($(data).find('.paginable-results').children());
+    // Update the serahc box's action so that it retains sort settings
+    const searchPath = paginable.find('.paginable-results #search-path');
```
jollopre: html id are unique across an entire document, therefore .paginable-results can be omitted. Anyway, I don't seem to see #search-path id associated to any html attribute. Do am I missing anything here?

app/controllers/concerns/paginable.rb
```diff
@@ -48,10 +48,17 @@ def searchable?
...
     def paginable_search_link(text = _('link name'), page = 1)
+      @paginable_path_params['search'] = params[:search] if searchable?
       url = paginable_base_url(page)
-      url+= "?search=#{params[:search]}" if searchable?
       return link_to(text, url, 'data-remote': true)
     end
+    # Generates an HTML link with clear search functionality
+    # text {String} - Represents the text for the searchable link
+    # page {String | Fixnum } - Represents the page to request for a search term
+    def paginable_clear_url(page = 1)
```
jollopre: this introduces side effects, i.e. paginable_clear_url is ONLY removing search param if present but keeping any query param present in the latest URL request. For example, if the search results where sorted by any field this will be held still. Is this what we want? As a user I would expect to clear any action made after hitting "Clear search results"

app/controllers/concerns/paginable.rb
```diff
@@ -38,7 +38,7 @@ def paginable_base_url(page = 1)
     # Generates an HTML link to sort given a sort field.
     # sort_field {String} - Represents the column name for a table
     def paginable_sort_link(sort_field)
-      return link_to(sort_link_name(sort_field), sort_link_url(sort_field), 'data-remote': true)
+      return link_to(sort_link_name(sort_field), sort_link_url(sort_field).html_safe, 'data-remote': true)
```
jollopre: sort_link_name is already safe html through raw which is indeed a wrapper for html_safe method, therefore is not needed to add .html_safe here

app/controllers/concerns/paginable.rb
```diff
@@ -18,7 +18,7 @@ def paginable_renderise(partial: nil, controller: params[:controller], action: p
       @paginable_controller = controller
       @paginable_action = action
       @paginable_path_params = path_params
-      merge_query_params(query_params)
+      @paginable_path_params = @paginable_path_params.merge(merge_query_params(query_params))
```
jollopre: merge_query_params(query_params) is already merging query params into params object and since you are merging that into paginable_path_params instance variable, the variable holds effective a copy of params object and therefore becomes useless.
Please, get rid of @paginable_path_params instance variable across the concern and replace with params instead:
```
params = merge_query_params(query_params)).merge(path_params)
```

lib/assets/javascripts/utils/paginable.js
briri: its a new hidden input field that is being used to retain the sort rules after the paginable-results is refreshed. Currently, the sort field/direction are lost anytime the user searches.

app/controllers/concerns/paginable.rb
briri: If I cleared search results I would expect my sort rules to remain in place. Maybe a question for @stephaniesimms and @sjDCC.

app/controllers/concerns/paginable.rb
briri: for some reason it was coming through in the tests as `sort_direction=asc&amp;sort_field=email`. This fixed that issue and didn't seem to have any adverse affect on the html page. Perhaps the way the test is accessing the results outside of the normal UI workflow is the issue. The test is imitating the ajax call but without the original context of the page with its default sort rules.

app/controllers/concerns/paginable.rb
briri: Hmm. they didn't seem to be available. I may have been looking at the wrong thing though.

briri:
I'm going to back out the changes to the concern and the paginable and search erb files. We should ask @sjDCC and @stephaniesimms whether or not they want sort rules to be perpetuated when the user searches or clears the results.

I'll resubmit this with just the default sorts for each table. We can update the concern if necessary in a separate ticket.

stephaniesimms:
I just noticed this and yes! I definitely want the sort rules to be perpetuated when I search or clear results. this is an extremely annoying usability issue in the DMPTool currently (that we don't do this) and requires lots of unnecessary clicking. please retain the sort rules. feel free to check with @sjDCC too.

briri:
@jollopre I pulled out the changes to the concern. We can address in a separate issue/PR.
This now just contains the default sorts for each table, a default sort for the org admin plans csv download, and I also added that `role="button"` for the new cancel button on the create plan page

PR #1193, opened 2018-02-21
fixed issue with clearing all theme selections

### LINKED ISSUE #1183: Can't edit/remove all themes from a question
When I uncheck themes for any question while editing a template and click Save, it remove some but not all of the themes. Every time I uncheck and click Save at least one theme remains checked after the page reloads. Overall the behavior is erratic, sometimes saving my changes and other times retaining all of the checked themes. If I click Delete, it deletes the entire question.

Desired behavior is to allow the user to Save a question with the desired themes or uncheck all boxes and have no themes.

screen shot 2018-02-20 at 7 51 59 pm

briri:
Fixes #1183

### DIFF
--- a/app/controllers/guidances_controller.rb
+++ b/app/controllers/guidances_controller.rb
@@ -67,8 +67,10 @@ class GuidancesController < ApplicationController
     authorize guidance
     guidance.text = params["guidance-text"]
     
-    if guidance.update_attributes(guidance_params)
-
+    attrs = guidance_params
+    attrs[:theme_ids] = [] unless attrs[:theme_ids]
+    
+    if guidance.update_attributes(attrs)
       if guidance.published?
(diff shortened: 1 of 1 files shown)


app/controllers/guidances_controller.rb
```diff
@@ -67,8 +67,10 @@ def admin_update
     authorize guidance
     guidance.text = params["guidance-text"]
     
-    if guidance.update_attributes(guidance_params)
-
+    attrs = guidance_params
+    attrs[:theme_ids] = [] unless attrs[:theme_ids]
```
jollopre: Another solution could be including a hidden field theme_ids for when no themes are selected through JS associated to guidances view. We did like that for the notification preferences (e.g. https://github.com/DMPRoadmap/roadmap/blob/development/lib/assets/javascripts/views/users/notification_preferences.js#L15)

PR #1205, opened 2018-02-23
auto expand 1st phase accordion on edit template

briri:
was part of a ticket assigned to Sam


app/views/org_admin/templates/_edit.html.erb
```diff
@@ -33,13 +33,14 @@
     <div class="panel-group" id="phases_accordion" role="tablist">
       <!-- If template has phases-->
       <% if template_hash[:template][:phases].present? %>
+        <% i = 0 %>
```
jollopre: We could check for first instead of adding an extra variable and its counter
```
template_hash[:template][:phases].each do |o|
aria-expanded="<%= template_hash[:template][:phases].first == o ? 'true' : 'false' %>"
end

Issue #1200, opened 2018-02-22
Shib users can't log in if email domain contains caps


stephaniesimms:
A DMPTool tester noticed a strange phenomenon with a new DMPTool account creation (through Shib @ UVa) giving an error because of Upper/lower case. Once the user changed the email address returned by the IdP from Virginia.EDU to virginia.edu, they were able to log in.

stephaniesimms:
looking for someone who can test this...

stephaniesimms:
case sensitivity removed. UVA testing works now. closing.

PR #1206, opened 2018-02-23
allow uppercase in emails

briri:
#1200 
Fix for JS validation of emails which prevents capital letters in email addresses. We auto-populate the email returned from Shibboleth IdPs so we cannot guarantee lowercase

### DIFF
--- a/lib/assets/javascripts/utils/isValidInputType.js
+++ b/lib/assets/javascripts/utils/isValidInputType.js
@@ -10,7 +10,7 @@ import {
 */
 export const isValidEmail = (value) => {
   if (isString(value)) {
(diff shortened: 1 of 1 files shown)


vyruss:
@briri as long as we remember to always do case-insensitive matching for emails, this should be fine to merge (e.g. `jimmy@anywhere.com` must count as taken when `Jimmy@Anywhere.com` tries to sign up)

jollopre:
@vyruss, this line should check for email in the database in a case insensitive manner (e.g. https://github.com/DMPRoadmap/roadmap/blob/development/app/controllers/registrations_controller.rb#L53)

Issue #1060, opened 2018-01-17
Changing my Org when I have admin privileges throws nasty error msg


stephaniesimms:
When I clicked the x to remove my org affiliation under Edit Profile, then entered my pw in the popup to confirm my changes, I received a nasty error message (screenshot below). When I returned to my profile page the changes were saved successfully (i.e., Org field was blank). I got the same error message when I attempted to change my email address and clicked Save. And now I get the error when i try to change and Save any profile info for this account.

We should replace the nasty error with something nicer in the banner at the top of the page or a modal that notifies admin users of the implications of changing their org and/or email: 

Note: If you remove or change your organisational affiliation or email address you will lose any administrative privileges associated with this account. Are you sure you want to take this action?

![pw-update-profile-no-org](https://user-images.githubusercontent.com/12703638/35066524-c22412f6-fb85-11e7-8755-971017bd4b76.png)

briri:
I can't replicate this issue on roadmap-stg @stephaniesimms 

The system only asks for your password when you change your email address. I tried a few combinations of removing my org affiliation and changing my email to get the password popup but everything works ok.

If I remove the org affiliation and do not enter something in the other org box I get: "Error: Save Unsuccessful. Please select an organisation from the list, or enter your organisation's name."

stephaniesimms:
i'll test some more on Mon @briri - don't waste any more time on it now if you can't replicate

stephaniesimms:
I tested again and get the error message each time. It appears when I try to change any info for this account stephanie.simms@gmail.com - I also noticed that the account no longer appears in the full list of User accounts when I login as a Super Admin. I updated the original issue above @briri 

Steps to repeat:
1. user w/Org admin privileges changes email address in Edit profile page
2. clicks save
3. enters pw into modal and clicks Continue
4. then comes the nasty error message

stephaniesimms:
the behavior is much nicer now. if i try to change my email address as an Org Admin the system prevents me from doing it. we should revise the Error message (screenshot) to give clearer instructions about why this is:

"Error: Save unsuccessful. You must remove/change your Organizational affiliation to remove your administrator privileges. Only then you will be able to change your email address."

I can easily remove my admin privileges by changing my Org. We should also add a tooltip to prevent admins from accidentally doing this. @sjDCC thoughts and suggestions about messaging here?

[...]

PR #1216, opened 2018-02-23
Issue 1060

jollopre:
removed side effects when generating or removing API token. 
Removed org_id method from User and use after_update callback instead. #1060

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -65,6 +65,8 @@ class User < ActiveRecord::Base
     end
   }
 
+  after_update :when_org_changes
+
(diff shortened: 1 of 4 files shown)


app/models/user.rb
```diff
@@ -249,7 +228,7 @@ def keep_or_generate_token!
         random_token = SecureRandom.urlsafe_base64(nil, false)
         break random_token unless User.exists?(api_token: random_token)
       end
-      self.save!
+      update_column(:api_token, api_token)  unless new_record?
```
briri: thanks for removing these. Its terrible to do a save in a method like that!

PR #1222, opened 2018-02-24
added check for blank question

briri:
edge case where a test template had a blank question text was causing an error

### DIFF
--- a/app/views/shared/export/_plan.erb
+++ b/app/views/shared/export/_plan.erb
@@ -34,7 +34,7 @@
                   <% end %>
                 </ul>
               <% else %>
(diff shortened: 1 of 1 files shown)


app/views/shared/export/_plan.erb
```diff
@@ -34,7 +34,7 @@
                   <% end %>
                 </ul>
               <% else %>
-                <p><%= raw question[:text][0].gsub(/<tr>(\s|<td>|<\/td>|&nbsp;)*(<\/tr>|<tr>)/,"") %></p>
+                <p><%= raw question[:text][0].gsub(/<tr>(\s|<td>|<\/td>|&nbsp;)*(<\/tr>|<tr>)/,"") if question[:text].present? && question[:text][0].present?  %></p>
```
jollopre: I don't get why are we baking new hash structures to represent phases, sections, questions. We are not only loosing the references defined in models but also we have to figure out keys and values for this newly structures.

PR #1223, opened 2018-02-24
fixed issue with api-token rake task

briri:
Forgot to PR this last week. Fixes the api token creation rake task

### DIFF
--- a/app/models/org.rb
+++ b/app/models/org.rb
@@ -151,7 +151,7 @@ class Org < ActiveRecord::Base
   end
 
   def grant_api!(token_permission_type)
-    org.token_permission_types << token_permission_type unless org.tokenpermission_types.include? token_permission_type
+    self.token_permission_types << token_permission_type unless self.token_permission_types.include? token_permission_type
   end
 
   private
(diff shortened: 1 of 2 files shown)


app/models/org.rb
```diff
@@ -151,7 +151,7 @@ def plans
   end
 
   def grant_api!(token_permission_type)
-    org.token_permission_types << token_permission_type unless org.tokenpermission_types.include? token_permission_type
+    self.token_permission_types << token_permission_type unless self.token_permission_types.include? token_permission_type
```
jollopre: self is implicit

lib/tasks/bugfix.rake
```diff
@@ -90,18 +90,21 @@ namespace :bugfix do
...
+    Rake::Task['bugfix:add_missing_token_permission_types'].execute
+    orgs = Org.where(is_other: nil)
     orgs.each do |org|
       org.grant_api!(TokenPermissionType::STATISTICS)
     end
-    # leave tokens intact
+    users = User.joins(:perms).where("org_id IN (?) AND (api_token IS NULL OR api_token = '')", orgs.collect(&:id))
     users.each do |user|
-      user.keep_or_generate_token!
+      if user.can_org_admin?
```
jollopre: it's fine because it is just a task but the email sending should not be invoked within keep_or_generate_token! method. We need to refactor that bit

Issue #1234, opened 2018-02-28
Error when trying to create account with an existing email


briri:
Receiving the following error when I tried to create an account for an email address that has already been used.
![screen shot 2018-02-28 at 11 23 41 am](https://user-images.githubusercontent.com/1204467/36808496-124e8760-1c7a-11e8-92fd-9dc5b196a03e.png)

Like caused by the recent modifications to the login workflow. I think this line was originally put into place for users who received an invitation to create an account when a user shared a plan with them. But instead of using the link, went ahead and tried to create a new account via the homepage.

stephaniesimms:
i thought we just fixed this, or was it in dmptool only @briri ? i'm confused after testing so many things on so many servers the past weeks.

briri:
yeah fixed it in the dmptool dev code, but this file differs a bit for us so didn't push the change over to roadmap. will have to update the roadmap version.

stephaniesimms:
working as expected on roadmap-stage. i get a nice error message that says my email is already registered. closing.

PR #1237, opened 2018-02-28
fixed issue with duplicate email when creating account

briri:
#1234 
added additional check for an invitation token


app/controllers/registrations_controller.rb
```diff
@@ -52,13 +52,13 @@ def create
     else
       existing_user = User.where_case_insensitive('email', sign_up_params[:email]).first
       if existing_user.present?
-        if existing_user.accept_terms?
+        if existing_user.invitation_token.present? && !existing_user.accept_terms?
```
jollopre: not sure why was
```
existing_user.accept_terms?
```
instead of
```
existing_user.accept_terms.nil?
```
anyway adding the invitation_token makes it safer. Thanks

Issue #1240, opened 2018-03-01
blank answer text causing error in download/export


briri:
An answer with blank text causes an error when exporting/downloading

stephaniesimms:
I tested the plans that gave Bad Gateway messages before and can download them now. there's still a problem with text files but that appears to be a separate issue. closing this one.

PR #1241, opened 2018-03-01
patch for download/export blank answers

briri:
#1240

### DIFF
--- a/app/views/shared/export/_plan.erb
+++ b/app/views/shared/export/_plan.erb
@@ -40,7 +40,7 @@
             <% end %>
 
             <% answer = @plan.answer(question[:id], false) %>
-            <% blank = answer.present? ? answer.text.gsub(/<\/?p>/, '').gsub(/<br\s?\/?>/, '').chomp.blank? : true %>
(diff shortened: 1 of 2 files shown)


app/views/shared/export/_plan.erb
```diff
@@ -40,7 +40,7 @@
             <% end %>
 
             <% answer = @plan.answer(question[:id], false) %>
-            <% blank = answer.present? ? answer.text.gsub(/<\/?p>/, '').gsub(/<br\s?\/?>/, '').chomp.blank? : true %>
+            <% blank = (answer.present? && !answer.text.blank?) ? answer.text.gsub(/<\/?p>/, '').gsub(/<br\s?\/?>/, '').chomp.blank? : true %>
```
jollopre: @briri, please consider is_valid? instead (e.g. https://github.com/DMPRoadmap/roadmap/blob/development/app/models/answer.rb#L75) since answers for an option_based question the text is purely a comment.

briri:
switched to use `answer.is_valid?` instead of checking for blank. I think it would be good to move the additional checks found in `views/shared/export/_plan.erb` into that method afterword to make sure the answer.text doesn't contain things like `<p></p>` which would currently evaluate to a valid response. Too close to your release though to make that change as it would potentially impact the plan writing pages

PR #1244, opened 2018-03-02
removed joins from paginable orgs

briri:
Missed the paginable version of the query on the last PR


briri:
@jollopre I'm going to go ahead and merge this one. It goes along with PR #1242. This is the paginable version of the same query

Issue #599, opened 2017-08-22
Replace gemfile for front-end translations with the underlying Jed library


jollopre:
As part of moving all the assets to Webpack, the gem "gettext_i18n_rails_js" has to be replaced with Jed library directly. A separated issue to deal with this will allow progressing on the ticket #587

briri:
move JS constants to hidden input elements on the page so that the regular GetText can be used for translation. Then update JS to pull in the values when needed.

briri:
picking this one up @jollopre as we want to get translation files out to a few people. Will add the constants to the `views/layouts/application.html.erb` file and tie them in via the `lib/assets/constants.js`

stephaniesimms:
can't test in UI. closing issue.

PR #1253, opened 2018-03-02
moved JS constants to application.html.erb

briri:
#599 
Moved all JS constants to json string in the `js-constants` hidden input on the `/views/layouts/application.html.erb` page. Updated JS to pull in that data and updated existing calls for constant values.

This should make the JS constants translatable


xsrust:
Looks good, thanks @briri

Issue #1247, opened 2018-03-02
PDF formatting settings don't work


stephaniesimms:
None of the PDF formatting settings (font type, size, margins) are working. Everything I test produces the same default formatting.

briri:
still having issues 
![screen shot 2018-03-07 at 8 53 13 am](https://user-images.githubusercontent.com/1204467/37105933-4da1d3e2-21e5-11e8-93b1-01700698bea0.png)

stephaniesimms:
working as expected now. closing issue.

PR #1259, opened 2018-03-07
removed debug line and fixed integer check

briri:
#1247


briri:
merging this minor change as well @vyruss since its a critical issue preventing PDF download

PR #1260, opened 2018-03-08
Rda backend mvp

xsrust:
Contains all RDA work

- Added helper functions to answers for reading/writing rda answers as a hash
- added rake task to add new question type
- added display for new question type
    - note that the autocomplete box is currently not working and is hidden
- allows adding rda question from backend
- update answers controller to deal with RDA questions
- added export for RDA questions

### DIFF
--- a/app/models/answer.rb
+++ b/app/models/answer.rb
@@ -10,9 +10,9 @@ class Answer < ActiveRecord::Base
       else
         plan.touch  # Force updated_at changes if nothing changed since save only saves if changes were made to the record
       end
-    end 
+    end
   end
-  
+
   ##
   # Associations
 	belongs_to :question
--- a/app/models/answer.rb
+++ b/app/models/answer.rb
@@ -97,7 +97,7 @@ class Answer < ActiveRecord::Base
   #                'standards' is a list of <std_id>: <title> pairs
   #                'text' is the text from the comments box
   def answer_hash
-    default = {'standards' => {}, text => ''}
+    default = {'standards' => {}, 'text' => ''}
     begin
       h = self.text.nil? ? default : JSON.parse(self.text)
     rescue JSON::ParserError => e
@@ -115,8 +115,8 @@ class Answer < ActiveRecord::Base
   # nothing returned, but the status of the text field of the answer is changed
   def update_answer_hash(standards={},text="")
     h = {}
-    h[:standards] = standards
-    h[:text] = text
+    h['standards'] = standards
+    h['text'] = text
     self.text = h.to_json
   end
 end
--- a/app/models/answer.rb
+++ b/app/models/answer.rb
@@ -99,7 +99,7 @@ class Answer < ActiveRecord::Base
   def answer_hash
     default = {'standards' => {}, text => ''}
     begin
-      h = self.text.nil? ? default : JSON.parse(answer.text)
+      h = self.text.nil? ? default : JSON.parse(self.text)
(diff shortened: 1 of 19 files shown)


app/models/answer.rb
```diff
@@ -86,4 +86,37 @@ def is_valid?
...
+
+  ##
+  # Returns the parsed JSON hash for the current answer object
+  # Generates a new hash if none exists for rda_questions
+  #
+  # @return [Hash] the parsed hash of the answer.
+  #                Should have keys 'standards', 'text'
+  #                'standards' is a list of <std_id>: <title> pairs
+  #                'text' is the text from the comments box
+  def answer_hash
```
briri: This seems specific to RDA. Would be good to refactor at some point since we may end up with other answer types that need to store as hash objects.

Perhaps extending Answer would be a good approach `class RdaAnswer < Answer` and then override the `text` and `text=` methods

lib/assets/javascripts/views/answers/rda_metadata.js
```diff
@@ -0,0 +1,406 @@
...
+      url: url + id.slice(4),
+      type: 'GET',
+      crossDomain: true,
+      dataType: 'json',
+    }).done((results) => {
+      descriptions[id] = {};
+      descriptions[id].title = results.title;
+      descriptions[id].description = results.description;
+      noWaiting -= 1;
+    });
```
briri: Should also handle a `.fail()` here

lib/assets/javascripts/views/answers/rda_metadata.js
```diff
@@ -0,0 +1,406 @@
...
+        minTree[num].standards = result.ids;
+      } else {
+        minTree[num].children[child].standards = result.ids;
+      }
+      if (result.ids.length < 1) {
+        removeUnused(name);
+      }
+      noWaiting -= 1;
+      initDescriptions(result.ids);
+    });
```
briri: again, should have a `.fail()` here

lib/assets/javascripts/views/answers/rda_metadata.js
```diff
@@ -0,0 +1,406 @@
...
+    url = data.url;
+    // get api_tree structure from api
+    $.ajax({
+      url: `${url}subject-index`,
+      type: 'GET',
+      crossDomain: true,
+      dataType: 'json',
+    }).done((result) => {
+      subjectCallback(result);
+    });
```
briri: need a `.fail()` here

lib/assets/javascripts/views/answers/rda_metadata.js
```diff
@@ -0,0 +1,406 @@
...
+        if (key === standardsArray[key]) {
+          selectedStandards.append(`<li class="${key}">${key}<button class="remove-standard"><i class="fa fa-times-circle"></i></button></li`);
+        } else {
+          selectedStandards.append(`<li class="${key}">${descriptions[key].title}<button class="remove-standard"><i class="fa fa-times-circle"></i></button></li>`);
+        }
+      });
+    });
+  }
+
+  function waitAndUpdate() {
```
briri: What happens when the API request to the RDA API fails or takes an exceptionally long time? Does the user know something is happening in the background? The `noWaiting` variable doesn't get set back to zero until you receive a successful response form the API call.

You could also have probably used @jollopre `utils/debounce.js` for this which provides a timer like this along with a way to cancel it.

Issue #1245, opened 2018-03-02
Fix formatting of DOCX plans


stephaniesimms:
there's still some weird indentation going on (screenshot below). 

the cover page plan title is fine but the main plan title is still indented. so is the subtitle on the cover page "A plan created..." these titles can either be centered or left justified. whichever is easier to implement is fine - just make it consistent.

![screen shot 2018-03-02 at 10 00 17 am](https://user-images.githubusercontent.com/12703638/36914042-be15c0c2-1e00-11e8-86b1-15e5ff6d58ae.png)

stephaniesimms:
@briri text and csv download work but other file formats produce the error below:

![screen shot 2018-03-07 at 11 09 41 am](https://user-images.githubusercontent.com/12703638/37112727-3045bbd4-21f8-11e8-93d3-42edd9149eea.png)

briri:
same issue as #1247

stephaniesimms:
i can download DOCX now. formatting looks good enough. there are still small rogue indentations but i understand that these downloads are a nightmare. and since you mentioned the possibility of swapping out the download gem i think we're good enough for now. closing.

PR #1262, opened 2018-03-08
Fixed PDF fomratting issues

briri:
PDF formatting options stopped working when we moved off of the asset pipeline. Was able to fix with some adjustments to the plan_styling partial and reference to web based fonts #1245 #1247


app/views/shared/export/_plan.erb
```diff
@@ -1,9 +1,30 @@
+<%
+  font_face = (@formatting[:font_face].present? ? "#{@formatting[:font_face]}" : 'Arial, Helvetica, Sans-Serif')
+  font_size = (@formatting[:font_size].present? ? "#{@formatting[:font_size]}" : '12')
+  margin_top = '5'
```
jollopre: if these values are meant to be defaults for exporting to pdf, I don't think here is the right place

PR #1267, opened 2018-03-09
Fix for duplicate answer issue

vyruss:
Fix for nasty bug (add Note after Answer has been saved -> duplicate blank answer saved)


briri:
hot fixed both dmptool and dmponline production with this one

PR #1270, opened 2018-03-09
Remove duplicate answers Rake task

xsrust:
Adds rake bugfix: remove_duplicate_answers

Logic for merging duplicate answers:
- creates a new answer object in place of the old ones
- user/updated_at/created_at/plan/question are set from the most recently answered question
- Notes from ALL answers have their answer_id changed to that of the new answer
- Question Options
    - dropdown is taken from the most recently saved answer
    - all other sets of question options are merged
- Text
    - if only one answer has text, the new answer's text is that answer's text
    - If both answers have text, the new answer's text is a concatonation of the multiple texts
    - Concatonations are marked with: `<p><strong>ANSWER SAVED TWICE - REQUIRES MERGING<\strong><\p>` and separated by `<p><strong>-------------<\strong><\p>`
- Old answers are deleted

### DIFF
--- a/lib/tasks/bugfix.rake
+++ b/lib/tasks/bugfix.rake
@@ -107,4 +107,59 @@ namespace :bugfix do
       end
     end
   end
+
+
+  desc "Remove Duplicate Answers"
+  task remove_duplicate_answers: :environment do
+    ## Concat Duplicate Answers
+    ActiveRecord::Base.transaction do
+      plan_ids = ActiveRecord::Base.connection.select_all("SELECT a1.plan_id as plan_id FROM Answers a1 INNER JOIN Answers a2 ON a1.plan_id = a2.plan_id AND a1.question_id = a2.question_id WHERE a1.id > a2.id" ).to_a.map{|h| h["plan_id"]}.uniq
+      plans = Plan.where(id: plan_ids)
+      plans.each do |plan|
+        plan.answers.pluck(:question_id).uniq.each do |question_id|
+          answers = Answer.where(plan_id: plan.id, question_id:  question_id).order(:updated_at)
+          if answers.length > 1 # Duplicates found
+            puts "found duplicate for plan:#{plan.id}\tquestion:#{question_id} \n\tanswers:[#{answers.map{|answer| answer.id}}]"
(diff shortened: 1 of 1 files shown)


lib/tasks/bugfix.rake
```diff
@@ -107,4 +107,59 @@ namespace :bugfix do
...
+              new_answer.question_options << answers.last.question_options.first
+              puts "\t adding option answers.last.question_options.first.text" unless answers.last.question_options.first.blank?
+            end
+            answers.reverse.each do |answer|
+              if num_text == 0 && answer.text.present? # case first present text
+                new_answer.text = answer.text
+                num_text += 1
+              end
+              if num_text == 1 && answer.text.present?
+                text = "<p><strong>ANSWER SAVED TWICE - REQUIRES MERGING</strong></p>"
```
vyruss: HTML tags problem

lib/tasks/bugfix.rake
```diff
@@ -107,4 +107,59 @@ namespace :bugfix do
...
+                text += new_answer.text
+                new_answer.text = text + "<p><strong>-------------</strong></p>" + answer.text
+              end
+              answer.notes.each do |note|
+                note.answer_id = new_answer.id
```
vyruss: this `.id` doesn't exist yet (before `new_answer.save`)

lib/tasks/bugfix.rake
```diff
@@ -107,4 +107,59 @@ namespace :bugfix do
       end
     end
   end
+
+
+  desc "Remove Duplicate Answers"
+  task remove_duplicate_answers: :environment do
+    ## Concat Duplicate Answers
+    ActiveRecord::Base.transaction do
+      plan_ids = ActiveRecord::Base.connection.select_all("SELECT a1.plan_id as plan_id FROM Answers a1 INNER JOIN Answers a2 ON a1.plan_id = a2.plan_id AND a1.question_id = a2.question_id WHERE a1.id > a2.id" ).to_a.map{|h| h["plan_id"]}.uniq
```
briri: Could have probably gotten to the list of questions in one query, but if this is working its fine since its really only run once.

PR #1286, opened 2018-03-13
Select the correct phase for editing in edit plan

vyruss:


briri:
looks good to me. good catch on the preload

Issue #1220, opened 2018-02-23
Admins should not be able to comment on plans unless invited


stephaniesimms:
As an Org Admin or Super Admin I can open and view all plans in read only mode, which includes adding comments. I should only have access to comment on plans for which users have requested feedback (these plans appear in the notifications panel at the top of the Plans page). 

Correct behavior is to set permissions so admins can only view plans but disallow commenting for plans in the main table (screenshot). Plans that a user submits for feedback should appear in the Notifications panel above this table with read only permissions that permit commenting.

<img width="1014" alt="screen shot 2018-02-23 at 1 33 31 pm" src="https://user-images.githubusercontent.com/12703638/36618103-b668e956-189e-11e8-9574-e4f17138bbe4.png">

briri:
Ok will Remove the link to the plan from the table. 

Feedback seems to be working properly for me. I think the issue you see is a timing thing. 
When a user requests feedback it makes all admins for that Org a reviewer. If a user becomes an admin after the request for feedback has been made though they would not see the plan in their notification list because they were not an admin at the time the user requested feedback.

stephaniesimms:
feedback is working again, thanks! sarah created an issue to notify the Org contact email when feedback is requested, but it's fine as is for now. the most important thing is that admins can no see links to open all plans - only those for which feedback has been requested.

stephaniesimms:
this is working as expected now. closing.

sjDCC:
Hi @briri I'm reopening this issue as we had intentionally made the plan names clickable links and allowed the admins to view the contents. We had a whole long discussion about it in ticket #434 

On 5th December @stephaniesimms said:

"sorry, i got confused. @sjDCC is correct that yes, we do want to make the plan name a link to enable org admins to open & read the plan. per our discussion on the check-in call, we need to create an exception in the code that enables admins to see plan content but not comment or perform any actions (a true read-only access w/no commenting privileges). admins should only have the ability to provide comments when a plan owner has requested feedback.

related issue #876 reflects this desired functionality"

This was a desired new feature and the T&Cs allow for it. I've also promoted and demoed this

briri:
ok. Its a very easy thing to revert:
https://github.com/DMPRoadmap/roadmap/pull/1224/files

briri:
[...]

PR #1295, opened 2018-03-15
added check for user's role on a plan to determine if download tab is…

briri:
Added check to prevent the share and download tabs from displaying when an admin views a plan but does not have a role #1220

### DIFF
--- a/app/views/plans/_navigation.html.erb
+++ b/app/views/plans/_navigation.html.erb
@@ -13,14 +13,17 @@
       </a>
     </li>
   <% end %>
-  <% if plan.editable_by?(current_user) %>
-    <li role="presentation" class="<%= (isActivePage(share_plan_path(plan)) ? 'active' : '') %>">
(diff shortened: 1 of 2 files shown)


app/views/plans/_navigation.html.erb
```diff
@@ -13,14 +13,17 @@
       </a>
     </li>
   <% end %>
-  <% if plan.editable_by?(current_user) %>
-    <li role="presentation" class="<%= (isActivePage(share_plan_path(plan)) ? 'active' : '') %>">
-      <a href="<%= share_plan_path(plan) %>" role="tab" aria-controls="content"><%= _('Share') %></a>
+  <% if plan.users.include?(current_user) %>
```
jollopre: this statement fixes the problem for now because plan#readable_by?(user_id) is doing more than it should, specially lines after https://github.com/DMPRoadmap/roadmap/blob/master/app/models/plan.rb#L379. We should remove can_super_admin? ... from there and put it as part of the policy methods for plan. If not, readable_by does not really reflect a readable role from the roles table.

PR #1302, opened 2018-03-16
Exclude javascript directory which seems to break "rake gettext:find"

vyruss:


briri:
looks fine to me. We had to do this for DMPTool as well

Issue #1275, opened 2018-03-12
Enable super-admins to edit user profiles


sjDCC:
Make the entry in the name column in the 'users' table a link to their profile page so super-admins can change org affiliation before granting admin permissions. Currently it is not possible to change their org affiliation to the correct one (e.g. to a new org that has been added to enable admin use). A separate ticket is addressing the ability to view the org #1248 
![picture1](https://user-images.githubusercontent.com/6169940/37296550-36ae28f2-2613-11e8-8343-c2f387e2db49.png)

Many DMPonline admins come from overseas unis that aren't currently listed in the tool so we need to add them as an org and affiliate the user to this before assigning admin permissions. The 'edit privileges' column now just pops up a modal with admin permission settings rather than opening their user profile where affiliation can also be set. Please make the name a link where we can access their full profile and change the org affiliation before granting permissions. Adding the 'org' column in the user table is also very helpful so we can ensure they have the correct affiliation before granting rights. This will be addressed under the related ticket.

briri:
Are there other pieces of information or tasks a super admin should be able to perform for users on this page?
- [x] Change org affiliation
- [x] Change privileges?
- [ ] Change email address /name?
- [ ] Unlink institutional credentials?
- [ ] Change email notification prefs?
- [ ] Change/Reset password?
- [ ] Disable/Deactivate account (requires new field on table - Devise may offer something for this)

stephaniesimms:
@briri - the only other thing i might want to be able to do is disable/deactivate an account. i'm mainly thinking about bouncing emails, but this isn't a high priority so not something we need to solve now. for dmptool users we want to place the onus of managing accounts on the users themselves so i prefer not to be able to do all the things. but i'll let @sjDCC confirm if there's anything else she needs to be able to do as super admin.

briri:
Tasks needed:
- [x] Add active flag to users table
- [x] Update scopes in Users model to look at new flag
- [x] Override Devise authentication method in User model to prevent deactivated users from logging in
- [x] Add checkbox column to Admin -> Users page to allow admins to activate/deactivate users
- [x] Make user name column a clickable link for Super Admins so that they can edit the user's profile
- [ ] Update profile page to allow super admins access to other users' profiles
- [ ] Update mailer to skip users who are inactive
- [ ] Add unit tests for new active flag, and functional tests for controller methods

stephaniesimms:
[...]

Issue #1248, opened 2018-03-02
Add Org & Date created columns to Users table 


stephaniesimms:
- add "Organization" column to main User accounts table after Email column 
- add "Created date" column after "Org" and before "Last activity" (for date account was created)

from Perry Willett:
this is probably also just a superadmin issue, but when I get requests from new institutions to make someone an admin, I need to know if they're affiliated with the institution in the db, or if they're still a "non-partner". The old DMPTool included the institution on this page: https://dmptool.org/org/admin/users/admin_index

from dmptool admin:
I would also add that we find it useful to be able to know when an individual user created their account. Sometimes we see a surge of account creations from admins in a specific department, for example. That helps us to target certain outreach. I know in the Users list, we can sort by Last Activity, which is useful, but that account creation date also has value.

![dmptool-userlist](https://user-images.githubusercontent.com/1558172/36863163-c845f47c-1d3d-11e8-82a5-247f36a4c798.jpg)

stephaniesimms:
table looks good. closing issue.

sjDCC:
Looks good. Removing from board

PR #1303, opened 2018-03-16
Updates to Super Admin Users page

briri:
Addresses feature requests in #1275 and #1248 
- Added `users.active` flag (defaults to true)
- Added new columns to Super Admin -> Users page
  - User Name is now a link to new `super_admin/users/[:id]/edit` page
  - Organization name
  - Created date
  - Active an ajax enabled checkbox to allow admins to quickly activate/deactivate accounts
- Added new super admin user edit page (route, controller, policy, view and JS) to allow admin to update user name, email, org and language
- Added logic to User model to enable Devise to force users who are not active to be logged out
- Updated devise's default message about account being inactive
- Updated `user_mailer` so that deactivated users do not receive emails (unless its a devise email about their account or a plan is shared with them)
- Added tests

Also:
- added logic to privilege update to only show/email users if the values have changed
- Fixed failing org controller test by setting default orgs.link value if not available 
- Added 'date-column' class to stylesheet to set min-width for date columns on users table. Usable elsewhere if necessary

**Need to run `rake db:migrate`**

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -50,8 +50,8 @@ class User < ActiveRecord::Base
 
   # Retrieves all of the org_admins for the specified org
   scope :org_admins, -> (org_id) { 
-    joins(:perms).where("users.org_id = ? AND perms.name IN (?)", org_id,
-      ['grant_permissions', 'modify_templates', 'modify_guidance', 'change_org_details'])
+    joins(:perms).where("users.org_id = ? AND perms.name IN (?) AND users.active = ?", org_id,
+      ['grant_permissions', 'modify_templates', 'modify_guidance', 'change_org_details'], true)
   }
 
   scope :search, -> (term) {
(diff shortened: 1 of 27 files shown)


app/models/user.rb
```diff
@@ -67,6 +67,12 @@ def filter(query)
 
   after_update :when_org_changes
 
+  ##
+  # This method uses Devise's built-in handling for inactive users
+  def active_for_authentication?
+    super and self.active?
```
vyruss: can you please change `and` to `&&` because of this Ruby weirdness: 
https://stackoverflow.com/a/1426835/6234472

vyruss:
Otherwise, tested and seems fine. The only issue we may face in UAT is: 
When you edit a user and save, you have to click the back-arrow twice to go back to the list of users, and then you are not on the right paginable page

briri:
Updated that line of code @vyruss and also added a back button to the main users page (which doesn't retain the pagination/sort info, but is at least one less click for now)

Issue #1197, opened 2018-02-22
Add asterisk to mandatory fields


stephaniesimms:
Add asterisk to mark mandatory form fields via CSS. Try red asterisks cf. DMPtool (screenshot below).

![screen shot 2018-02-21 at 4 20 52 pm](https://user-images.githubusercontent.com/12703638/36513179-38ae48ac-1723-11e8-84d8-06ed0219e033.png)

briri:
a CSS solution won't work for this. We should be able to add some JS to the ariatiseForm.js and validation.js though to add in the asterisk. Just need to determine how best to apply it to regular input fields vs checkboxes and radio buttons. Maybe something like:
```javascript
const addAsterisk = (el) => {
  if (isObject(el) && el.attr('aria-required') === 'true') {
    if (el.attr('type') === 'checkbox') {
      el.after(' * ');
    } elseif (el.attr('type') === 'radio') {
        el.closest('.form-group').before('* ');
      } else {
        const label = el.closest('.form-group').find('label');
        if (isObject(label)) {
          label[0].before('* ');
        }
      }
    }
  }
};
```
An example of required radio buttons can be found on the share plan page and example of a required checkbox can be found on the create account page

briri:
Updated to add asterisks to any fields marked with `aria-required="true"` which drives our form validations. 
@stephaniesimms and @sjDCC it looks though like we don't require all of the fields we should be requiring. For example in the Template and Guidance pages in particular there are a lot of fields I would expect to be required (e.g. Question text, Section title, etc.) We should probably go through all of the admin pages and create a new ticket for making any of these fields required (can all be listed on one ticket).

stephaniesimms:
asterisks look good @briri - @sjDCC and I can go through all pages and create a new issue for any additional fields that should be required. 

for now, there's just a tiny html issue that crept in on the Edit Profile page. see screenshot below and remove <p> tag

![screen shot 2018-03-25 at 6 41 53 pm](https://user-images.githubusercontent.com/12703638/37883020-dc0502d2-305c-11e8-8ee4-69bceabc089d.png)

jollopre:
After adding asterisks, I've just noticed that create plans has changed. Is that the layout desired?
![untitled](https://user-images.githubusercontent.com/7393044/37895571-9d9531ac-30d9-11e8-9852-eb828412cdb5.png)

sjDCC:
[...]

PR #1306, opened 2018-03-20
Add asterisk to required fields

briri:
adds asterisks to required fields per #1197 
- consolidated question edit/new js so that ariatise and tinymce are only invoked once
- updated timing of validation checks on ariatiseForm because they were firing before Tinymce was finished transferring its contents to the textarea
- updated a few views to make sure that they had appropriate labels so that asterisks could be applied


lib/assets/javascripts/views/questions/edit.js
```diff
@@ -11,5 +11,10 @@ $(() => {
     questionEdit.hide();
     questionEdit.parent().find('.question_show').show();
   });
+  $('.new_question_cancel').on('click', (e) => {
```
jollopre: good catch with this. Please, rename the file accordingly since now handles a cancel button for edit form and and a cancel button for new form

lib/assets/stylesheets/overrides.scss
```diff
@@ -477,15 +482,24 @@ $tooltip-background: $grey;
...
-.create-plan-or {
-  padding: 5px 0 0 15px;
-}
-.create-plan-checkbox {
-  margin-left: -40px;
-  margin-top: -5px;
-}
-.create-plan-mock {
-  margin-left: -40px;
+#new_plan {
```
jollopre: I don't think we need these styles for tweaking the new_plan form elements. You could easily adjust the elements in the view using col-md-*, pull-left, pull-right, etc. Can you have a try, please?

lib/assets/stylesheets/overrides.scss
```diff
@@ -477,15 +482,24 @@ $tooltip-background: $grey;
...
+  }
+  .create-plan-checkbox {
+    margin-left: -40px;
+    margin-top: -5px;
+  }
+  .create-plan-mock {
+    margin-left: -40px;
+  }
+  .h2-label {
+    font-size: 30px;
```
jollopre: HTML h2 heading is 30px by default if I remember well.

lib/assets/stylesheets/overrides.scss
```diff
@@ -477,15 +482,24 @@ $tooltip-background: $grey;
...
+    font-weight: normal;
+  }
+  .red {
+    margin-left: 15px;
```
jollopre: it seems to me a very confusing class selector for the property styles that define. What about ml-15 instead?

lib/assets/javascripts/views/questions/edit.js
briri: renamed

lib/assets/stylesheets/overrides.scss
briri: Spent an hour trying but its going to be more work than just swapping in some Bootstrap classes and reorganizing the row/col classes. This ticket is for adding Asterisks to required fields not refactoring the html/css of pages. You can open a separate ticket if you feel its important

Issue #1239, opened 2018-03-01
New Org not displayed in Orgs list


briri:
If you create a new Org via the Super Admin page they do not appear in the list of orgs on the Super Admin Orgs page.

I believe its due to the join on the templates table (a new org would have none). This join speeds up the query but will not work for this scenario since orgs need not have a template.

`render 'index', locals: { orgs: Org.includes(:templates, :users).joins(:templates, :users) }`

cpwillett:
It's not just new orgs. I see a lot of institutions missing. There were 226 institutions in the database that we migrated, but only 76 appear in the list at https://dmptool.org/super_admin/orgs

briri:
yes, in attempt to optimize the query a join was added to the templates and users tables to get their counts. The type of join that is being used though will ignore an org if it has no users or templates.

Fix for this is coming today

stephaniesimms:
this is working as expected now. all Orgs appear in the list and I can see new Orgs when I add them. closing.

PR #1313, opened 2018-03-21
Downloads bugfix

xsrust:
Addresses #1239 and another reported issue.

User reported error msg on plan download (PDF)
- Due to not checking if an answer has text before attempting to .gsub  (case for option_based questions which would be valid with options but no text)
- 1239 is due to an issue in the logic where answers with blank text will not get exported, breaking export for option_based questions.

Changes:
- Refactored &fixed checking for blank to answer.rb
- Switched export to conditionally check for presence of options/text with special case to handle rda questions's hash answers.

NOTE: REBASE DEVELOPMENT and CONTRIBUTIONS after merging

### DIFF
--- a/app/models/answer.rb
+++ b/app/models/answer.rb
@@ -87,6 +87,18 @@ class Answer < ActiveRecord::Base
     return notes.select{ |n| n.archived.blank? }.sort!{ |x,y| y.updated_at <=> x.updated_at }
   end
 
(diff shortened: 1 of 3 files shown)


xsrust:
It's been brought to my attention that Question text for questions with options is not currently exporting.  Updating this as well in the following commit

PR #1314, opened 2018-03-21
made reviewers unable to download plans

briri:
Updated plan navigation so that reviewers cannot download plans per #1220

### DIFF
--- a/app/views/plans/_navigation.html.erb
+++ b/app/views/plans/_navigation.html.erb
@@ -19,10 +19,12 @@
         <a href="<%= share_plan_path(plan) %>" role="tab" aria-controls="content"><%= _('Share') %></a>
       </li>
     <% end %>
-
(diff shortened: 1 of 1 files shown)


app/views/plans/_navigation.html.erb
```diff
@@ -19,10 +19,12 @@
         <a href="<%= share_plan_path(plan) %>" role="tab" aria-controls="content"><%= _('Share') %></a>
       </li>
     <% end %>
-
-    <li role="presentation" class="<%= (isActivePage(download_plan_path(plan)) ? 'active' : '') %>">
-      <a href="<%= download_plan_path(plan) %>" role="tab" aria-controls="content"><%= _('Download') %></a>
-    </li>
+    <%# download should be available for any users that the plan is shared with but not reviewers %>
+    <% if !plan.reviewable_by?(current_user) %>
```
jollopre: fix the problem but we should address 'access' to stuff through policies mainly because this hides the view but the action to download is still available if they know the link.

Issue #1298, opened 2018-03-15
Expand/contract symbols (+/-) not displaying correctly


dsisu:
Please complete the following fields as applicable:

**Expected behaviour:**

Section should expand when user clicks +
Section should contract when user clicks -

**Actual behaviour:**
- Section doen't always expand when user clicks +, but the button turns to - . In other words, it sometimes takes two clicks to expand a section. When the section finally expands, the button turns to +. See two screenshots enclosed below.

![image](https://user-images.githubusercontent.com/2138956/37478119-5201d954-2871-11e8-8d66-82316e365682.png)

![image](https://user-images.githubusercontent.com/2138956/37478210-872bf93e-2871-11e8-8b32-7623f4aea7c9.png)

**Steps to reproduce:**

Create plan go to Edit (or Edit any existing plan) and try to expand contract sections. ZonMw template reproduces the error. Horizon2020 does it one and off.

briri:
yes @vyruss that seems to be the case. If you wait for the page to finish loading its not an issue. As you note, its tied to the way the page loads everything all at once instead of when the individual accordion sections are expanded.

dsisu:
For the record, another user who mentioned this was for DMP template of Hanze University of Applied Sciences.

dsisu:
@briri It's one of those things, you're not even aware the page is still loading so you click on the plus again, thinking that maybe you haven't 'hit the button'. Not a major issue but worth fixing. Thanks!

sjDCC:
Working now. Closing out.

xsrust:
I don't think that this was worked on,
I can reproduce this on -test.
You just need to expand an accordian before the page loads in all the way
![Screenshot 2019-10-01 at 09 56 48](https://user-images.githubusercontent.com/7352616/65948259-d0419480-e431-11e9-8a6c-73eb2b20abb9.png)

Issue #1285, opened 2018-03-13
TinyMCE loads content.min.css once for each text area


vyruss:
Please complete the following fields as applicable:

**Expected behaviour:** `content.min.css` loaded once

**Actual behaviour:** `content.min.css` loaded x times for x text areas which is actively slowing down our page loads for large templates (it also gives the impression that you can't click on accordions to edit)

**Steps to reproduce:** edit a plan with multiple questions

Proposed solution here perhaps?
https://stackoverflow.com/questions/47835486/tinymce-loads-the-content-min-css-for-each-textarea

briri:
related to #1298

briri:
Did some initial experimentation with the above recommendations. Setting the 'mode' on the Tinymce init method didn't seem to have any affect.
I think that spending some time to redesign the way this page is structured/served up to the UI would be a better approach. (e.g. load sections and comments areas via ajax on demand)

jollopre:
I dug into this problem while trying to fix #1365. Tinymce loads the content.min.css asynchronously for each instance created and since each textarea got associated a new iframe for tinymce, there is no way to prevent loading multiple times the content.min.css (unless inline mode is used, which we don't want).

In order to speed up the page rendering, we should load things on demand (as @briri suggested) and therefore tinymce instances would be loaded incrementally when the user requests a specific area.

mesterjagels:
Any update on this issue?

briri:
Hi @mesterjagels, no we have not addressed this issue yet. We will be investigating the new [Rails 6 ActionText](https://blog.saeloun.com/2019/10/01/rails-6-action-text.html) as a replacement for Tinymce along with a larger redesign of the way the forms are generated/loaded.

benjaminfaure:
Needs some investigation if this issue is still relevant

coderjinhui:
Please improve this behavior! This will cause the loading bug when using the slow network!

Steps:
page1 has the editor, page2 also has the editor.
When using slow network, opening page1 will load the editor, and before the editor finished loading, navigate to page2.

Then, navigate back to page1, the editor will not load forever! It will be always blank!!!

What a horrible bug!

PR #1323, opened 2018-03-22
started work on load sections individually (DO NOT MERGE)

briri:
Started playing around with ways to load section accordions on the write plan page on demand instead of all together.

Lets discuss in the dev meeting tomorrow

related to #1285 and #1298


briri:
closing this one since we're already working through the rewrite of the templates pages so that they load sections on demand. 
We should eventually copy those patterns over to the plan edit page

Issue #1214, opened 2018-02-23
Retain default sorting for tables when user clicks 'View all'


stephaniesimms:
This is an outstanding part of issue #1187 - Default sorting was completed. But we still want to retain default sorting when user clicks 'View All' to expand the tables.

- on My Dashboard: default sorting should be by plan 'Last edited' date, most recent first
- all other tables: default sorting should be alphabetical by first field (Template Name, User Name, etc)

stephaniesimms:
this doesn't appear to be working. when I click "View All" on plans tables on My Dashboard, DMP Templates, Public Plans, etc the sorting still changes. moving back to in progress.

jollopre:
@stephaniesimms I don't think the server has deployed any of this stuff yet

stephaniesimms:
clicking 'View All' retains default sort for all tables visible to logged-in users (plans, templates, etc). But this doesn't appear to be working for the DMP Templates or Public DMPs tables @jollopre

jollopre:
@stephaniesimms, any page that is 'paginable', i.e. has search, sorting and page links behaves in the same way. I have tried the two cases that you described and the sort is properly retained.

Your comment suggest 'default sorting' which indeed these two pages don't have and therefore you have to click to any sortable column to be retained. What is the default sorting you want to apply to these? 

I would create a new issue to address the default sorting of these two views, since the retaining of the sorting I believe it works properly.

stephaniesimms:
thanks for the clarification @jollopre - i thought these tables had a default set, but it appears that they do not so i'll create an issue as you suggested. all other tables look good so i'll go ahead and close this issue.

Issue #1186, opened 2018-02-21
Clicking 'View All' for Users and Orgs as Super Admin blows up server


stephaniesimms:
The queries for the Users and Organizations tables include a join to provide plan and template counts - these queries are overloading the server when a Super Admin clicks the link to 'View All' in these tables (I got a proxy error for both)

Disable View All for Super Admin users on these tables - updated per @sjDCC and my confirmation that this is a good solution

sjDCC:
Yeah, @jollopre expressed concern about how big some of these queries would be. I’m not sure we need the ‘view all’ so let’s just disable it for superadmin.

stephaniesimms:
View All link is no longer present on Users page at all. Desired behavior is to remove the link for Super Admins only and retain it for Org Admins. In the latter case it's unlikely that the number of users will cause issues and admins want to be able to expand the table to see an overview of all their users.

please reinstate the ability to View All users for Org Admins @jollopre

stephaniesimms:
looks good now on roadmap stage. closing issue.

PR #1324, opened 2018-03-23
Issue1214

jollopre:
Refactored paginable concern to accommodate proper retaining of query_params. DMPRoadmap/roadmap#1214

There are additional test cases that should be added but I plan to do it as part of #1186


briri:
this looks reasonable @jollopre. We will need to update the other views that use paginable to work with this.
Just fix the linter issues and then feel free to merge yourself.

jollopre:
@briri thanks for reviewing. I've fixed the linter problems, the API method used across the views (e.g. paginable_sort_link) has not changed at all so therefore there is no need to update any view.

PR #1325, opened 2018-03-26
Issue 1186

jollopre:
- paginable_base_url_with_query_params using double splat. paginable method instead of local variable. #1186
- view_all option for paginable_renderise method in order to restrict access to ALL data from a scope. DMPRoadmap/roadmap#1186
- fixed integration tests to accommodate the view_all option. DMPRoadmap/roadmap#1186


briri:
looks good @jollopre. Nice use of Ruby's splat argument.

Issue #1334, opened 2018-03-28
Changes to paginable JS break existing ajax form submissions


briri:
Please complete the following fields as applicable:

**Expected behaviour:**
The existing ajax functionality embedded in paginable tables should still be working after updating the paginable sort/search logic

**Actual behaviour:**
Existing ajax calls no longer work. The table disappears and no message is displayed to the user. We receive a `Uncaught TypeError: Cannot read property 'ownerDocument' of undefined` error in the browser console. 

I was able to tie this back to the change in the jquery selector here: https://github.com/DMPRoadmap/roadmap/blob/a59c966035bdc68ee0b44a288ccbb63e2afb926e/lib/assets/javascripts/utils/paginable.js#L14

We have been leveraging the Rails 'data-remote' functionality to embed small forms on each row of a table to handle ajax calls within tables. The selector in the line above is now too broad.

**Steps to reproduce:**
- Login
- Click on one of the 'test' checkboxes in the Plans table of My Dashboard.

OR

- Login as a super admin
- Go to the Users page
- Click on the 'active' checkbox or change a users privileges

PR #1335, opened 2018-03-28
Made ajax:success method for paginable more specific

briri:
Fixes issue with paginable JS `ajax:success` handler #1334


briri:
Note that this is a bug fix being PRed to master

briri:
sounds good Jose. Will deploy to roadmap-stg and test it out

PR #1336, opened 2018-03-29
Issue1334

jollopre:
- any paginable link or form uses paginable-action selector to avoid conflicting with other wider selectors. DMPRoadmap/roadmap#1334


jollopre:
@briri, I'm merging this proposal but tagging you in case you want to add some comments

briri:
ah ok. This will be a better solution long term. Thanks for updating.

Issue #1342, opened 2018-04-04
Creating a funder template makes it privately visible by default


vyruss:
Please complete the following fields as applicable:

**Expected behaviour:** When a funder creates a template, it should get visibility: Public

**Actual behaviour:** When a funder creates a template, its visibility is automatically set to Private

stephaniesimms:
as a funder (e.g., NIH) I can create a new public template and it appears in the create plan wizard. but I don't have the ability to select any guidance (see screenshot below). when I click to "See full list of guidance" there's nothing there. everything works as expected when I create a new plan with any other template, just not the newly created public one.

![screen shot 2018-04-05 at 10 12 27 am](https://user-images.githubusercontent.com/12703638/38380948-1f8f6fe6-38ba-11e8-95b4-db7655270d4e.png)

stephaniesimms:
testing again on roadmap stage. i created a new NIH template with themed questions and it appears on the public DMP Templates page. i can also create a new plan with the template, but I only see the option to select default guidance (UCOP in screenshot). the expected behavior is to see the full list of guidance.
![screen shot 2018-04-06 at 1 38 05 pm](https://user-images.githubusercontent.com/12703638/38443245-eb1e110e-399f-11e8-888f-acd03aac0b96.png)

sjDCC:
tested too and all working for me - both under create plan and public templates page

Issue #1318, opened 2018-03-22
Admin users unable to customise template


dsisu:
Please complete the following fields as applicable:

**Expected behaviour:**

Able to edit customisable funder templates

**Actual behaviour:**

If you try to add a section, for example, it appears that all works correctly but when you come back to view the template later, all the edits are missing.

**Steps to reproduce:**
Try to customise BBSRC template, for example. All funder templates at UoE appear to have not been customised so assign yourself to University of Manchester and compare behaviour in BBSRC template vs AHRC template.

The issue appears to be relating to templates that haven't been customised before. To identify these at a glance, check whether the status is set to 'not customised'. If you don't want to rely on the 'status' info only (which has been unrealiable on one occasion), also check the 'actions' available, which should include 'customise'.

I reported that Kerry Miller at UoE tried to customise BBSRC template by adding sample answers and got the same results. It's probably down to the fact that users are not seeing the latest version of the template.

dsisu:
This needs to be resolved by the 17th of April at the latest.

briri:
I tried this on roadmap-stg by logging in as myself (super admin affiliated with UCOP) and selected to customize the DOD template (no previous customizations for that one). I added an example answer and some guidance to the first question, saved the changes and then went back and published the customized template.

I then logged in as a separate user also affiliated with UCOP and created a plan (research org = UCOP and funder = DOD). I went to the Write Plan tab and see my custom guidance but do not see my example answer. 
![screen shot 2018-04-03 at 3 09 57 pm](https://user-images.githubusercontent.com/1204467/38278586-85c6d9b2-3751-11e8-9983-6516ac6d75bf.png)

stephaniesimms:
after more testing on roadmap stage, I can confirm most of what @briri observed for the DOD template. As a Pitt admin, I customized the NIH-GDS template with guidance text and example answers, then created a plan with it. I did the same thing on dmptool stage.

The screenshot below shows the following errors with how the customizations are displayed (or not):
- The first tab for the NIH-GDS funder guidance that is attached to the template is renamed "Pitt." This tab should say "NIH-GDS"
- the Pitt guidance text customization appears in this tab with the same generic "Guidance" label. It should say "Pitt Guidance" if it is supposed to appear here. Otherwise it should be under the second tab with the correct "Pitt" label that also contains generic themed guidance.
- example answers do not appear at all. they should be displayed below the text editor.

[...]

PR #1346, opened 2018-04-04
patch to fix issue with default template visibility setting

briri:
Sets the template visibility to 'Public' for new funder templates and 'Organisational' for all others. Orgs that are both funder and organisation/institution still retain the ability to change the visibility after the template is created (on the edit template details page). Their templates are Organisationally visible by default though when initially created. #1342 & #1318


briri:
ah yes. thanks for catching that

briri:
removed db.schema changes

Issue #1344, opened 2018-04-04
Customisation example answers do not appear when editing plan


vyruss:
**Expected behaviour:** When you edit a plan using a customised template, you should see either:
    * Both the funder's and the customising org's example answers
    * The customising org's example answer

**Actual behaviour:** You only see the funder's example answer and the other one is never displayed

**Screenshot:**
![image](https://user-images.githubusercontent.com/18707439/38319966-2893ee80-382b-11e8-9d60-f1cd5f36d91c.png)

stephaniesimms:
testing on dmptool stage and here is what i see @briri  (screenshot below):
- example answers now appear w/appropriate org label, but there is also a label w/the funder name and a blank space if there's no funder-provided sample answer below every question. desired behavior is to only display an example answer for the funder or the org if one has been provided.

![screen shot 2018-04-05 at 9 59 49 am](https://user-images.githubusercontent.com/12703638/38380437-8c372078-38b8-11e8-9cec-c893cea82000.png)

stephaniesimms:
this now works as expected on roadmap stage. closing issue.

sjDCC:
Yay! This is fantastic to see. We've been asked to display multiple example answers in the past so we've added a new feature here. One to promote in the release notes and blog ;-)

Looks great!
![example-answers](https://user-images.githubusercontent.com/6169940/38814600-c3b05310-4189-11e8-9a2d-e5cc000d240d.PNG)

Issue #1343, opened 2018-04-04
Add Funder & Org abbreviation before "Guidance" word in customised templates


vyruss:
Please complete the following fields as applicable:

**Expected behaviour:** Guidance when using a customised template should show the origin of the guidance

**Actual behaviour:** Funder guidance appears as issued from customising organisation 

**Screenshot:**
![image](https://user-images.githubusercontent.com/18707439/38319798-a60e4d34-382a-11e8-957c-52adb3f444b3.png)

stephaniesimms:
testing on dmptool stage and here is what i see @briri (screenshot below):
- the appropriate funder or org name appears before "guidance", but the tab is still renamed with the org name when it should retain the funder name. so in the example below, the first tab should still be "NIH" (instead it's changed to the customizing org - UCD)

![screen shot 2018-04-05 at 9 59 49 am](https://user-images.githubusercontent.com/12703638/38380521-dcaa97e2-38b8-11e8-8a12-3bad0f3e51ea.png)

stephaniesimms:
working as expected now on roadmap stage. closing issue.

sjDCC:
guidance all showing right for me now

PR #1347, opened 2018-04-04
added org abbreviation to 'Guidance' sections and added customized ex…

briri:
- Removed example answer from guidances section since its never used (query looks only for guidance type annotations)
- Added Org abbreviation to 'Guidance' label when the template is customized #1343
- Used existing model method to get customization example answer and display it underneath the template's original example answer (would be better to refactor into a single query later) #1344
- Updated 'Answer locked by another user' section to exclude reshowing example answers since they appear below the main answer

screen shot 2018-04-04 at 1 02 18 pm

### DIFF
--- a/app/views/annotations/_show.html.erb
+++ b/app/views/annotations/_show.html.erb
@@ -1,12 +1,8 @@
 <% org = template.org.abbreviation.present? ? template.org.abbreviation : template.org.name %>
-<% if example_answer.present? || guidance.present? %>
+<% if guidance.present? %>
   <dl<%= for_plan ? ' class="dl-horizontal"' : '' %>>
-    <% if example_answer.present? %>
-      <dt><%= _('Example answer') %></dt>
-      <dd><%= raw example_answer.text %><dd>
-    <% end %>
     <% if guidance.present? %>
-      <dt><%= _('Guidance') %></dt>
+      <dt><%= template.customization_of.nil? ? _('Guidance') : _('%{org} Guidance') % { org: guidance.org.short_name } %></dt>
       <dd><%= raw guidance.text %></dd>
     <% end %>
   </dl>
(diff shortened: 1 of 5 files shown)


app/views/annotations/_show.html.erb
```diff
@@ -1,12 +1,8 @@
 <% org = template.org.abbreviation.present? ? template.org.abbreviation : template.org.name %>
-<% if example_answer.present? || guidance.present? %>
+<% if guidance.present? %>
   <dl<%= for_plan ? ' class="dl-horizontal"' : '' %>>
-    <% if example_answer.present? %>
-      <dt><%= _('Example answer') %></dt>
-      <dd><%= raw example_answer.text %><dd>
-    <% end %>
     <% if guidance.present? %>
```
jollopre: this if is no longer needed

app/views/answers/_new_edit.html.erb
```diff
@@ -64,15 +64,18 @@
...
   <!--Example Answer area -->
-  <% annotation = question.first_example_answer %>
-  <% if annotation.present? && annotation.text.present? %>
-    <div class="panel panel-default">
-      <span class="label label-default">
-        <%="#{annotation.org.abbreviation} "%> <%=_('example answer')%>
-      </span>
-      <div class="panel-body">
-        <%= raw annotation.text %>
+  <% annotations = [question.first_example_answer] %>
```
jollopre: This is workable but I don't think is understandable at a glance. A more simplistic approach would be question.annotations.where(type: Annotation.types[:example_answer]) which would bring both example answers in one go. Obviously we should move data access to the controllers or model

app/views/answers/_new_edit.html.erb
```diff
@@ -1,4 +1,4 @@
-<%# locals: { question, answer, readonly, locking } %>
+<%# locals: { template, question, answer, readonly, locking } %>
```
jollopre: this partial is used also for previewing a template, did you test if still works?

PR #1351, opened 2018-04-06
Issue 1318

jollopre:
Existing template customisation appears now to be editable. DMPRoadmap/roadmap#1318


briri:
looks fine to me @jollopre 
I'm going to sync these bug fixes back with development so that they're available to us for the template versioning work

PR #1356, opened 2018-04-06
Template model updates

briri:
Hey @jollopre 

Here's a start to the Template model refactoring. I rebased with all of the latest bug fixes for template customizations.

So far I did the following:
- refactored the self-join relationships so that we now have `template.versions` and is tied to `family_id`, `customizations` is now tied to `customization_of` and `customized_from` ties back to the original template
- Started refactoring the scopes (I think the self-join changes above may make some of the old ones unnecessary)
- Started adding Unit tests

Look for your name in the code, I left you a bunch of comments on how I've been organizing things as I work


app/models/template.rb
```diff
@@ -39,84 +43,181 @@ class Template < ActiveRecord::Base
...
+    end
+  }
+  
+# Jose: I updated the relationships above so that we now have template.versions 
+#       which ties to family_id; and customizations and customized_from which are
+#       tied to customization_of.
+#
+#       We many be able to use those in many instances instead of these class
+#       scopes
+  scope :latest_version_numbers, -> (family_id = nil) {
```
jollopre: what about latest_version_per_family instead?

PR #1359, opened 2018-04-09
Fixes issue with exporting plans as text

### LINKED ISSUE #1358: Download plan as text file failing
Users are unable to export their plans as Text.  

screen shot 2018-04-09 at 8 30 15 am

briri:
Fixes #1358

### DIFF
--- a/app/views/shared/export/_plan_txt.erb
+++ b/app/views/shared/export/_plan_txt.erb
@@ -30,7 +30,7 @@
     <% section[:questions].each do |question| %>
       <% if @show_sections_questions %>
         <%# text in this case is an array to accomodate for option_based %>
-        <% if question[:text].length > 1 %>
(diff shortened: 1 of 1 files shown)


app/views/shared/export/_plan_txt.erb
```diff
@@ -30,7 +30,7 @@
     <% section[:questions].each do |question| %>
       <% if @show_sections_questions %>
         <%# text in this case is an array to accomodate for option_based %>
-        <% if question[:text].length > 1 %>
+        <% if question[:text].is_a?(Array) %>
```
jollopre: Can you change it to:
```
question[:text].respond_to(:each) 
```
using duck typing is safer because in the case that hash key had an ActiveRecord::Relation, that is not an object of Array class however it responds to each method to iterate the collection.

PR #1364, opened 2018-04-09
Fixes iteration error on templates_controller

### LINKED ISSUE #1363: Customized version of funder template not being used

**Expected behaviour:** 
The user is given the customized version of the funder template (if one is available for the research org they select)

**Actual behaviour:**
The user is sometimes given the base funder templates.

**Steps to reproduce:**
This may be specific to the deployment and its Ruby/Rails versions.
- As an org admin, login and customize a funder template
- Publish your customization
- Create a new plan (select the customizing org as the research org and then select the funder)

briri:
Fixes #1363 

Opted to work with a separate array to hold the funder templates. We can explore more elegant solutions to this as we work on the template versioning rework.

### DIFF
--- a/app/controllers/org_admin/templates_controller.rb
+++ b/app/controllers/org_admin/templates_controller.rb
@@ -375,15 +375,18 @@ module OrgAdmin
       if org_id.present? || funder_id.present?
         unless funder_id.blank?
           # Load the funder's template(s)
-          templates = Template.valid.publicly_visible.where(published: true, org_id: funder_id).to_a
+          funder_templates = Template.valid.publicly_visible.where(published: true, org_id: funder_id)
 
-          unless org_id.blank?
+          if org_id.blank?
+            templates = funder_templates.to_a
+          else
             # Swap out any organisational cusotmizations of a funder template
-            templates.each do |tmplt|
+            funder_templates.each do |tmplt|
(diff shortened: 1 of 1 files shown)


app/controllers/org_admin/templates_controller.rb
```diff
@@ -375,15 +375,18 @@ def template_options()
...
+            funder_templates.each do |tmplt|
               customization = Template.valid.find_by(published: true, org_id: org_id, customization_of: tmplt.dmptemplate_id)
               if customization.present? && tmplt.created_at < customization.created_at
-                templates.delete(tmplt)
                 templates << customization
+              else
+                templates << tmplt
               end
             end
           end
```
jollopre: An alternative solution would be using map which iterates over the elements of a list but 'preserves' the original list.
```
templates = Template.valid.publicly_visible.where(published: true, org_id: funder_id)
unless org_id.blank?
   templates = templates.map do |tmpt|
      customization = Template.valid.find_by(published: true, org_id: org_id, customization_of: tmplt.dmptemplate_id)
      if customization.present? && tmpl.created_at < customization.created_at
         customization
      else
         tmpl
      end
   end
end
```
but in any case the performance is hit because we are doing N queries. I believe we can sort this issue easier with the new template scope query methods

PR #1383, opened 2018-04-12
Issue1341

briri:
Ok, so here's what I got done today:
- moved customize code out of the controller 
- added a generate_copy method based on logic in the controller and added test for it
- made deep_copy private on the template model so that we can ensure that the modifiable/published/default/etc. flags are always correct becasue people are using the new methods
- added some test helpers to check for equality of phases/sections/questions/annotations/options that can be used with generate_version, generate_copy, customize and the various deep_copy methods
- updated a line to stop using scaffold_template test_helper in favor of the newer more complete/accurate init_[object] methods

We should move the scopes back into the main model class and remove the few that are still in there (maybe after the controller work when we're sure they're no longer being called). We should also add an `archive` method to set `archived=true` for the template and its historical copies.

### DIFF
--- a/app/controllers/org_admin/templates_controller.rb
+++ b/app/controllers/org_admin/templates_controller.rb
@@ -201,7 +201,7 @@ module OrgAdmin
     # the funder template's id is passed through here
     # -----------------------------------------------------
     def transfer_customization
-      current_customization = Template.includes(:org).find(params[:id])
+      @template = Template.includes(:org).find(params[:id])
       @current_tab = params[:r] || 'all-templates'
       authorize @template
       new_customization = @template.upgrade_customization
@@ -214,7 +214,7 @@ module OrgAdmin
     def copy
       @template = Template.find(params[:id])
       authorize @template
-      new_copy = template.generate_copy
+      new_copy = @template.generate_copy(current_user.org)
       if new_copy.save!
         flash[:notice] = "#{@template.template_type.capitalize} was successfully copied."
         redirect_to edit_org_admin_template_path(new_copy, edit: true, r: 'organisation-templates')
--- a/app/controllers/org_admin/templates_controller.rb
+++ b/app/controllers/org_admin/templates_controller.rb
@@ -53,10 +53,10 @@ module OrgAdmin
       @template.links = (params["template-links"].present? ? JSON.parse(params["template-links"]) : {"funder": [], "sample_plan": []})
 
       if @template.save
-        redirect_to edit_org_admin_template_path(@template), notice: success_message(@template.template_type, _('created'))
+        redirect_to edit_org_admin_template_path(@template), notice: success_message(template_type(@template), _('created'))
       else
         @hash = @template.to_hash
-        flash[:alert] = failed_create_error(@template, @template.template_type)
+        flash[:alert] = failed_create_error(@template, template_type(@template))
         render action: "new"
       end
     end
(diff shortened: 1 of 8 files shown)


app/controllers/org_admin/templates_controller.rb
```diff
@@ -204,80 +204,25 @@ def transfer_customization
...
-                  annotation_copy.save!
-                end
-              end
-            end
-          end
-        end
-      end
-      new_customization.save
+      new_customization = @template.upgrade_customization
+      new_customization.save!
```
jollopre: this is best handled inside that method, so that here we just need to handle any possible exception after attempting to save. I'll remove this line from here and will update family_id and version to be consistent with the newer changes you introduced.

app/controllers/org_admin/templates_controller.rb
```diff
@@ -204,80 +204,25 @@ def transfer_customization
...
-      new_copy.family_id = loop do
-        random = rand 2147483647
-        break random unless Template.exists?(family_id: random)
-      end
-
-      if new_copy.save
-        flash[:notice] = "#{@template.template_type.capitalize} was successfully copied."
-        redirect_to edit_org_admin_template_path(id: new_copy.id, edit: true, r: 'organisation-templates'), notice: _('Information was successfully created.')
+      new_copy = @template.generate_copy(current_user.org)
+      if new_copy.save!
```
jollopre: new_copy object is saved within the generate_copy I believe...

app/controllers/org_admin/templates_controller.rb
```diff
@@ -373,5 +318,10 @@ def template_options()
     def plan_params
       params.require(:plan).permit(:org_id, :funder_id)
     end
+    
+    def template_type(template)
```
jollopre: I think this method should leave in a helper for the template not here and since it is used only in the controller you can include the module instead of using the helper keyword from rails

app/models/template.rb
```diff
@@ -53,24 +53,25 @@ def default
...
-  def deep_copy(**options)
-    copy = self.dup
-    copy.version = options.fetch(:version, self.version)
-    copy.published = options.fetch(:published, self.published)
-    copy.save! if options.fetch(:save, false)
-    self.phases.each{ |phase| copy.phases << phase.deep_copy(options) }
-    return copy
-  end
-
+
```
jollopre: I'm not convinced to privatise this method, it is pretty flexible thanks to its configuration options and therefore it is safe the leave it as public since its defaults are sensible, i.e. it does not save a hierarchy. Moreover, can help to create specific unit tests without having to use generate_copy, customize, upgrade_customization... At the end in ruby even private methods are accessible (e.g. using send)

app/models/template.rb
```diff
@@ -53,24 +53,25 @@ def default
...
   # Returns whether or not this is the latest version of the current template's family
   def is_latest?
     return (self.id == Template.latest_version(self.family_id).pluck(:id).first)
   end
 
-  # Generates a new copy of self
+  # Returns a new unpublished copy of self with a new family_id, version = zero for the specified org
+  def generate_copy(org)
+    template = deep_copy(modifiable: true, version: 0, published: false, save: true)
+    template.update!({
```
jollopre: I'd rather prefer to to send all the attributes at once when we create the template so that we don't end up with temporary template records that are not yet valid. I'm going to extend the deep_copy for template

app/models/template.rb
```diff
@@ -146,6 +150,8 @@ def upgrade_customization
   # TODO: Themes & guidance?
   #
   # @return [hash] hash of template, phases, sections, questions, question_options, annotations
+
+# TODO: If there is time to update the UI to stop using hashes, remove this method
```
jollopre: agree on this. We have to remove any baked hash

test/test_helper.rb
```diff
@@ -148,6 +148,50 @@ def init_annotation(org, question, **props)
       nil
     end
   end
+
```
jollopre: good to see this helpers however I'm concern about the length this file is taking. We should split into multiple files so that it makes more maintainable the testing

test/unit/template_test.rb
```diff
@@ -86,23 +87,38 @@ def default_formatting
   end
 
   test "able to version a template" do
-    init_full_template(@basic_template)
-    assert_equal 0, @basic_template.version, 'expected the initial template version to be zero'
-    version2 = @basic_template.generate_version
+    template = init_full_template(@basic_template)
```
jollopre: I believe in ruby you pass the value of a reference as an argument, that means you modify @basic_template within the method (init_full_template) and return effectively a reference to the instance variable. Don't understand why do you want to assign the result into template since @basic_template contains the updates.

Issue #1375, opened 2018-04-11
Refactor Templates controller


briri:
Refactor the templates controller:
- update the actions to work with the newly refactored template model
- remove the `admin_` prefix from those actions
- update the routes.rb file and policies according to the above changes
- update all of the `old_route_path` references to the `new_route_path`
- update tests

related to #1340

stephaniesimms:
closing issues that can't be tested in the UI

PR #1387, opened 2018-04-16
Templates controller revisions

briri:
Addresses work in #1375
- refactored controller methods to work with new Template model methods
- Changed action for customize, transfer_customization, copy, publish and unpublish so that they are more RESTful
- Updated functional tests for the templates_controller
- Updated template_selection (mimics different create plan scenarios) integration tests
- Stubbed in new setup for the template_versioning (ensures versioning is happening at appropriate times) integration tests so that they make use of the new test_helper methods

### DIFF
--- a/app/controllers/org_admin/templates_controller.rb
+++ b/app/controllers/org_admin/templates_controller.rb
@@ -1,6 +1,7 @@
 module OrgAdmin
   class TemplatesController < ApplicationController
     include Paginable
+    include Versionable
     include TemplateFilter
     after_action :verify_authorized
 
@@ -9,6 +10,9 @@ module OrgAdmin
     def index
       authorize Template
       
+# TODO: Refactor this method and its views in next sprint if time allows so that
+#       we are no longer passing around hashes and do not need the template_filter.rb
+
       # Apply scoping
       all_templates_hash = apply_scoping(params[:scope] || 'all', false, true)
       own_hash = apply_scoping(params[:scope] || 'all', false, false)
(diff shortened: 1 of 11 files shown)


app/controllers/org_admin/templates_controller.rb
```diff
@@ -104,12 +94,8 @@ def update
       @template = Template.find(params[:id])
       authorize @template   # NOTE if non-authorized error is raised, it performs a redirect to root_path and no JSON output is generated
 
-      current = Template.current(@template.family_id)
-
       # Only allow the current version to be updated
-      if current != @template
-        render(status: :forbidden, json: { msg: _("You can not edit a historical version of this #{template_type(@template)}.")})
-      else
+      if @template.latest?
```
jollopre: you have to add here the logic to generate a new template version if the latest is published

app/controllers/org_admin/templates_controller.rb
```diff
@@ -178,33 +166,39 @@ def history
...
-      
-      @current_tab = params[:r] || 'all-templates'
-      redirect_to edit_org_admin_template_path(customisation, r: 'funder-templates')
+      begin
+        customisation = @template.customize!(current_user.org)
+        redirect_to edit_org_admin_template_path(customisation, r: @current_tab)
+      rescue Exception => e
+        flash[:alert] = _('Unable to customize that template.')
+        redirect_to org_admin_templates_path(r: @current_tab)
+      end
```
jollopre: Please, consider this change:
```
if customize?(current_user.org)
  begin
    customisation = @template.customize!(current_user.org)
    redirect_to edit_org_admin_template_path(customisation, r: @current_tab)
  rescue => e
    flash[:alert] = _('Unable to customize that template.')
    redirect_to org_admin_templates_path(r: @current_tab) 
  end
else
  redirect_to ... #perhaps to index ?
end
```
and remember that we want to rescue from exceptions that are recoverable, by default StandError or any subclass and in our case:
- ActiveRecord::StatementInvalid (not yet in code but will be raised when family_id and version already exists)
- ActiveRecord::RecordInvalid (if any validation for template fails)
- RuntimeError (the raise clauses we have right at the top of the method)

app/controllers/org_admin/templates_controller.rb
```diff
@@ -178,33 +166,39 @@ def history
...
+      @current_tab = 'funder-templates'
       authorize @template
-      # TODO add check @template.upgrade_customization?
-      new_customization = @template.upgrade_customization!
-      new_customization.save!
-      redirect_to edit_org_admin_template_path(new_customization, r: 'funder-templates')
+      begin
+        new_customization = @template.upgrade_customization!
+        redirect_to edit_org_admin_template_path(new_customization, r: @current_tab)
+      rescue Exception => e
```
jollopre: same changes proposes above

app/controllers/org_admin/templates_controller.rb
```diff
@@ -213,24 +207,21 @@ def copy
         new_copy = template.generate_copy!(current_user.org)
         flash[:notice] = "#{template_type(template).capitalize} was successfully copied."
         redirect_to edit_org_admin_template_path(new_copy, edit: true, r: 'organisation-templates')
-      rescue ActiveRecord::RecordInvalid => e
+      rescue Exception
```
jollopre: ``` rescue => e ``` again

app/views/paginable/templates/_funders.html.erb
```diff
@@ -66,16 +66,16 @@
               </button>
               <ul class="dropdown-menu">
                 <% if !customization.present? %>
-                  <li><%= link_to _('Customise'), customize_org_admin_template_path(template) %></li>
+                  <li><%= link_to _('Customise'), customize_org_admin_template_path(template), 'data-method': 'post' %></li>
```
jollopre: I like the use of AJAX here but I don't see the handlers for when the request succeeds or fails, this will leave the page unchanged unless you refresh the browser

lib/template_filter.rb
```diff
@@ -7,17 +7,15 @@ def apply_scoping(scope, customizable = false, all = false)
...
       orgs << Template.default.org unless current_user.org == Template.default.org
 
-      templates = Template.unarchived
-        .where(family_id: Template.families(orgs.map(&:id)).pluck(:family_id))
-        .where(published: true)
-        .where(visibility: Template.visibilities[:publicly_visible])
-        .includes(:org)
+      # Gather all of the unarchived published publicly visible templates
+      # for the specified orgs
+      templates = Template.includes(:org).families(orgs.collect(&:id)).
```
jollopre: If I'm not missing anything templates within this block is only used for collecting the family_ids since the else i.e. !customizable reinitialises it again, what about:
```
templates = Templates.families(orgs.map(&:id).pluck(:family_id)
```
if not, where is used the org pre-fetched?

test/integration/template_selection_test.rb
```diff
@@ -4,167 +4,105 @@ class TemplateSelectionTest < ActionDispatch::IntegrationTest
...
-    @funder_template.save
-
-    @org = @researcher.org
-    @org_template = Template.create(title: 'Org template', org: @org, archived: false)
-    # Template can't be published on creation so do it afterward
-    @org_template.published = true
-    @org_template.visibility = Template.visibilities[:organisationally_visible]
-    @org_template.save
+    # Need to clear the tables until we get seed.rb out of test_helper.rb
+    Template.delete_all
```
jollopre: I don't get why we delete_all here, can you explain the purpose to me?

briri: We're using the seeds.rb file to populate the test db but its not always clear how those records are setup (e.g. how many templates are there, which ones are published, who do they belong to, etc.) Its cleaner to just add what you need in the `setup` for each test that way you can guarantee that a change to seed.rb doesn't break your tests

app/controllers/org_admin/templates_controller.rb
briri: am holding off on that for the final ticket (although I don't think it needs to be a concern as the name of that ticket suggests). We should only version the template when 'versionable' fields have changed. I want us to clarify what those things are. For example if the description changes, or the visibility changes it probably doesn't need a new version so we'd want to see if the title and other versioanble fields are dirty.

briri:
updated the controller based on your feedback and to pull in the new Versioanble concern. I hope to get rid of that template_filter.rb entirely if we have time in the next sprint for the UI work.

lib/template_filter.rb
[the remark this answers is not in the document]
briri: I think that would be redundant since the 'families', 'published', etc. scopes are all starting from 'unarchived'. Will move the `.includes()` to the end of the chain instead which will ensure that any separate queries are based on the subset

briri:
Update the template_filter query. We can add the AJAX success/failure handlers while working on the UI tickets.

Issue #85, opened 2017-01-09
Consolidate DB migrations into one new migration


briri:
Remove all of the old DB migrations and build a new initial migration that is based on the current schema and any changes made to the DB from the onset of the Roadmap project to the end of the 1st week in January.

briri:
Started working on this but it makes sense to wait until after the DB refactor. Work was done in a branch on CDLUC3/roadmap

briri:
I still think we should do this just to clean up the migrations directory. A user cannot run `rake db:migrate` from a clean installation without encountering errors. It should be fairly easy to convert the the current schema.rb over to a migration file. @jollopre @vyruss and @xsrust what do you think?

Yu-Chieh-Henry-Yang:
I'm doing a project that use DMP Roadmap and we need to have database pass migration before going to production. Does anyone have an estimate about when the database consolidation this ticket mention will be finished?

Also, I provide a possible temporary workaround here: https://github.com/DMPRoadmap/roadmap/issues/1774 but I'm not sure if it breaks anything

Bodacious:
@briri @xsrust @Yu-Chieh-Henry-Yang This should not be necessary. The schema file (`db/schema.rb`) **_is_** a single migration that reflects the current DB structure. Creating a new, single migration would simply be a duplicate of the schema file.

The following rake tasks should be useful for almost all of the issues devs are likely to come across:

`rake db:migrate` should only be necessary when you have an existing DB that has fallen behind the schema at HEAD. If there are issues preventing those migrations from running, these should be added as issues with the **bug** label.

``` bash
rake db:reset          # Drop DB, create DB again, and load the schema to the current version
rake db:schema:load    # Load a schema.rb file into the database
rake db:create         # Creates the database from DATABASE_URL or config/database.yml for the current RAILS_ENV 
rake db:seed           # Load the seed data from db/seeds.rb (ONLY FOR DEVELOPMENT)
rake db:setup          # Create the database, load the schema, and initialize with the seed data (use db:reset to also drop the database first)
rake db:drop           # Drops the database from DATABASE_URL or config/database.yml for the current RAILS_ENV (use db:drop:all to drop all databases in the ...
rake db:migrate:status # Display status of migrations
rake db:rollback       # Rolls the schema back to the previous version (specify steps w/ STEP=n)
```

PR #1392, opened 2018-04-18
Issue 1341

jollopre:
In order to ensure thread-safe operation for:
- generate_version!
- customize!(org)

I decided to go with Optimistic Concurrence Control and its implementation requires:
- Creating an index (family_id, version) for a successful generate_version! without duplicates
- Creating an index (customization_of, version, org_id) for a successful customize!(org) without duplicates

Please, execute the following commands in the order stated:
```
rake upgrade:remove_duplicated_template_versions
rake db:migrate
```
since the templates tables has to be cleaned up before adding the above mentioned indexes. @briri, @vyruss, @xsrust, can you check the tasks created in case I missed any edge case?

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,11 +11,8 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20180405152454) do
+ActiveRecord::Schema.define(version: 20180418115318) do
 
-  # These are extensions that must be enabled in order to support this database
-  enable_extension "plpgsql"
-  
   create_table "annotations", force: :cascade do |t|
     t.integer  "question_id"
     t.integer  "org_id"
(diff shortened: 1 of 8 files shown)


db/schema.rb
```diff
@@ -174,8 +171,8 @@
     t.string   "logo_uid"
     t.string   "logo_name"
     t.string   "contact_email"
-    t.integer  "org_type",               default: 0,              null: false
-    t.text     "links",                  default: "{\"org\":[]}"
+    t.integer  "org_type",               default: 0,     null: false
+    t.text     "links",                  default: "[]"
```
briri: no, this should not change!

db/schema.rb
```diff
@@ -426,42 +428,4 @@
 
   add_index "users_perms", ["user_id"], name: "index_users_perms_on_user_id"
 
-  add_foreign_key "annotations", "orgs"
```
briri: why delete these foreign keys?

briri:
not to my knowledge @jollopre. go ahead and create one (maybe tie it to #85 which involves cleaning out the old schema migrations)

PR #1393, opened 2018-04-18
1367 versionable concern

xsrust:
Accidentally contains commit from Jose's PR as well so only merge if that PR succeeds

- Adds versionable concern
get_modifiable(obj)
 - if template DOES NOT need versioning, returns obj
 - If template DOES need versioning, returns the corresponding obj in the versioned template|
- where obj is  a Template/Phase/Section/Question/Annotation

### DIFF
--- a/app/controllers/concerns/versionable.rb
+++ b/app/controllers/concerns/versionable.rb
@@ -1,0 +1,75 @@
+module Versionable
+  extend ActiveSupport::Concern
+
+  def get_modifiable(obj)
+    if obj.respond_to? 'template'
+      template = obj.template
+    elsif obj.is_a? Template
+      template = obj
+    else
+      return nil # Throw error as wrong obj added
+    end
+    template, copy = find_or_generate(template)
+    # copy not generated - return obj
+    # copy generated and obj.is_a? template - return template
+    if copy
+      if obj.is_a? Template
+        obj = template
+      else
+        obj = find_in_space(obj,template.phases)
+      end
+    end
+    return obj
+  end
+
+  ##
+  # Creates a new verison IF template published
+  # to be called from the create/update/destroy actions of
+  #    phases/sections/questions/annotations
+  # @param template  -  Template object
+  # @return Template (new or current version)
+  def find_or_generate(template)
+    if template.generate_version?
+      new_template = template.generate_version!
+      return new_template, true
+    end
+    return template, false
+  end
+
+
+  ##
+  # Find the corresponding object
+  def find_in_space(obj, search_space)
+    # obj can be a phase/section/question/annotation
+    # case of objet being an instance of search space
+    puts search_space.first.class
+    puts search_space
+    if obj.is_a? search_space.first.class
+      puts "match"
+      if obj.respond_to?('number') && search_space.first.respond_to?('number')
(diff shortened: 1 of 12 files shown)


app/controllers/concerns/versionable.rb
```diff
@@ -0,0 +1,75 @@
...
+      template = obj.template
+    elsif obj.is_a? Template
+      template = obj
+    else
+      return nil # Throw error as wrong obj added
+    end
+    template, copy = find_or_generate(template)
+    # copy not generated - return obj
+    # copy generated and obj.is_a? template - return template
+    if copy
```
briri: Hmm. I like the approach but I don't think this bit is necessary.

The controller has the current template/phase/section/etc. why not assign whatever changes come through in the form and then version that instance (`generate_version!` will call `deep_copy` which does a `self.dup` which should carry forward the user's changes. That way we don't have to rely on this `number == number` stuff
```ruby
# In the controllers:
phase = Phase.find(params[:id])
authorize phase
...
phase.assign_attributes(phase_params) # assigns the user input to the instance without updating the record
...
# Call to your concern method to determine if it should be versioned
template = get_modifiable(phase)

# Then in your concern:
def get_modifiable(obj)
  ...  
  template = save_or_version!(template)
  ...

# Then on the Template model I guess something like this:
def save_or_version!(template)
  if template.generate_version?
    template.generate_version!
  else
    template.save!
  end
end
```

Or we could skip the concern entirely and just make sure that we do those steps in the controller:
```ruby
section = Section.find(params[:id])
... 
section.assign_attributes(section_params)
if section.template.generate_version?
  template = save_or_version(section.template) # will generate a new version with changes
else
  section.save! # no version so save the single object
end
```
We'd still need to figure out what the new id is to properly render/redirect the UI. But this way its already been versioned and the changes applied so we could search for the current title/text, etc.

jollopre: - Sam's approach attempts to locate an existing object on a draft or newly generated template. It might seem more complex but simplifies the logic vastly for the controllers. Also note that matching numbers should work if we assume non-repeated number on the collection we are looking for, however it fails when we are adding a new object because it will not be found anywhere in the hierarchy and
- Brian's approach modifies the existing and postpones the save to draft or newly generated to later. It is another option but I think is best to delegate the finding of the context to the concern instead. 

What about if we decorate the instance methods save and destroy for template, phase, section, question and annotation, i.e. each class has to decide whether to write into each immediate parent, i.e. question, section, phase or generate an entire new hierarchy and write into each newly immediate parent. Probably most of the logic around these two methods in these classes will be shared and therefore refactored later but we could delegate the decision of where to write to the model in which we want to make a change.

app/controllers/concerns/versionable.rb
```diff
@@ -0,0 +1,75 @@
+module Versionable
+  extend ActiveSupport::Concern
+
+  def get_modifiable(obj)
+    if obj.respond_to? 'template'
+      template = obj.template
+    elsif obj.is_a? Template
+      template = obj
+    else
+      return nil # Throw error as wrong obj added
```
jollopre: remember to raise error instead of returning nil

PR #1398, opened 2018-04-19
Fix for rake tasks

briri:
Was getting an unknown method 'add?' for nil class exception. Initializing the set seems to fix it.


lib/tasks/upgrade.rake
```diff
@@ -283,7 +283,7 @@ namespace :upgrade do
       .order(family_id: :asc, version: :asc, updated_at: :desc)
 
     current_family_id = nil
-    unique_versions = nil
+    unique_versions = Set.new
```
jollopre: The only edge case I can think for this is having templates with family_id set to nil but that should not happen...

PR #1399, opened 2018-04-19
Fix for contact-us when logged in - fixes #1394

### LINKED ISSUE #1394: Cannot submit contact us form when logged in

**Expected behaviour:**
User can no longer submit the contact us form when they are logged in. 

**Actual behaviour:**
User receives an invalid recaptcha message
screen shot 2018-04-18 at 9 51 45 am

**Steps to reproduce:**
- Log into the tool
- Go to the contact us page
- submit the form

There is logic that hides the recaptcha widget if the user is logged in the `views/contact_us/contacts/_new_left.html.erb` partial. We need to either disable the recaptcha check when the user is logged in or display it at all times. This is related to the work done on #501

vyruss:
Fixes #1394 .

Changes proposed in this PR:
- cater for logged in users

### DIFF
--- a/app/controllers/contacts_controller.rb
+++ b/app/controllers/contacts_controller.rb
@@ -3,13 +3,13 @@ class ContactUs::ContactsController < ApplicationController
   def create
     @contact = ContactUs::Contact.new(params[:contact_us_contact])
 
-    if verify_recaptcha(model: @contact) && @contact.save
-      redirect_to(ContactUs.success_redirect || '/', :notice => _('Contact email was successfully sent.'))
-    else
-      flash[:alert] = _('Captcha verification failed, please retry.')
-      redirect_to request.referrer
-      #render_new_page
+    unless user_signed_in?
+      unless verify_recaptcha(model: @contact) && @contact.save
+        flash[:alert] = _('Captcha verification failed, please retry.')
+        redirect_to request.referrer and return
+      end
     end
+    redirect_to(ContactUs.success_redirect || '/', :notice => _('Contact email was successfully sent.'))
   end
 
   def new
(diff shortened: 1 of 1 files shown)


app/controllers/contacts_controller.rb
```diff
@@ -3,13 +3,13 @@ class ContactUs::ContactsController < ApplicationController
...
     @contact = ContactUs::Contact.new(params[:contact_us_contact])
 
-    if verify_recaptcha(model: @contact) && @contact.save
-      redirect_to(ContactUs.success_redirect || '/', :notice => _('Contact email was successfully sent.'))
-    else
-      flash[:alert] = _('Captcha verification failed, please retry.')
-      redirect_to request.referrer
-      #render_new_page
+    unless user_signed_in?
+      unless verify_recaptcha(model: @contact) && @contact.save
```
briri: We should technically do the @contact.save check regardless of whether or not the user is signed in. Would also be good to do a render instead of redirect on failure so that the user doesn't lose the text they've entered in the fields.
Maybe something like (pseudocode .... you might need to tweak):
```ruby
@contact = ContactUs::Contact.new(params[:contact_us_contact])
flash[:alert] = nil

if !user_signed_in? && !verify_recaptcha(model: @contact)
  flash[:alert] = _('Captcha verification failed, please retry.')
end
if !flash[:alert].present? && @contact.save
  redirect_to(ContactUs.success_redirect || '/', :notice => _('Contact email was successfully sent.'))
else
  flash[:alert] = _('Unable to submit your request') unless flash[:alert].present?
  render 'new'
end
```

vyruss:
@briri done.

Issue #1064, opened 2018-01-19
Duplicate DOM ids for home page


jollopre:
Please complete the following fields as applicable:

**Expected behaviour:**
DOM ids must be unique across the entire page
**Actual behaviour:**
![dom](https://user-images.githubusercontent.com/7393044/35148536-791caad8-fd0a-11e7-8fbb-308e4f88ec48.png)

**Steps to reproduce:** (for Chrome)
1. Right click mouse
2. Click inspect
3. Go to home page for a non-login in user

stephaniesimms:
looks good upon inspection. closing issue.

PR #1400, opened 2018-04-20
Fixes #1397

### LINKED ISSUE #1397: Too many redirections in Sign Up
Hi, I found a error while working on ORCID authentication. I'm working on v1.0.3 but It occurss too in last release.

**Expected behaviour:** Fill fields in sign_up page and successfull login

**Actual behaviour:** Too many redirections after sign up (user is created in db but not redirecto to plans)

**Steps to reproduce:** Sign up in dmproadmap/users/sign_up

vyruss:
Fixes #1397 .

Changes proposed in this PR:
* after_sign_in|up : Check for paths not urls (so that it works on localhost/http)
* remove duplicate IDs from sign_up page by not including it in modal too (related to #1064)


jollopre:
@briri it should not be any duplicate id for the new_user_registration_path view since it has:
- the sign up form within the main view and
- the sign in alone in the popup

Note, the home page has loaded both forms too through the tab and there are no duplicated ids there since we covered that on one of the issues from the past. Does it make sense?

Issue #1373, opened 2018-04-11
Refactor Sections controller


briri:
Refactor the phases controller:
- Move the controller code to a new namespaced controller `org_admin/sections_controller.rb` 
- remove the `admin_` prefix from those actions
- update the routes.rb file and policies according to the above changes
- update all of the `old_route_path` references to the `new_route_path`
- update the tests

related to #1340

stephaniesimms:
closing issues that can't be tested in the UI

PR #1404, opened 2018-04-23
1) Section controller work for template versioning

briri:
For #1373 
- Moved sections controller, views and routes to org_admin namespace
- removed `admin_` prefix from action names
- updates views to work with new routing
- updated tests
- added 'edit', 'new', 'index' and 'show' actions that we can hopefully use when updating the UI


briri:
Update sections controller to use versionable methods (also added in StandardError handling for phases controller).
@jollopre The tests are failing due to an issue with the older version of Versionable.find_in_space(). The recursion is not happening properly.

briri:
synced with latest sprint2 branch

app/controllers/org_admin/phases_controller.rb
```diff
@@ -88,16 +88,21 @@ def create
...
-        flash[:notice] = success_message(_('phase'), _('created'))
-      else
-        flash[:alert] = failed_create_error(phase, _('phase'))
+        if phase.save!
+          flash[:notice] = success_message(_('phase'), _('created'))
+        else
+          flash[:alert] = failed_create_error(phase, _('phase'))
+        end
+      rescue StandardError => e
+        flash[:alert] = _('Unable to create a new version of this template.')
```
jollopre: We have specific messages for when an exception is raised. Do we want to display these directly to the user or we want to keep something more generic like this?

app/controllers/org_admin/sections_controller.rb
```diff
@@ -0,0 +1,148 @@
...
+    # POST /org_admin/templates/[:template_id]/phases/[:phase_id]/sections
+    def create
+      phase = Phase.includes(:template, :sections).find(params[:phase_id])
+      section = Section.new(section_params.merge({ phase_id: phase.id }))
+      authorize section
+      begin
+        section = get_new(section)
+      
+# TODO: update UI so that this comes in as part of the `section:` part of the params
+        section.description = params["section-desc"]
```
jollopre: now that we are refactoring why don't we use the description field instead of "section-desc" ?

app/policies/section_policy.rb
```diff
@@ -13,15 +13,31 @@ def initialize(user, section)
   #  - The template which they are modifying belongs to their org
   ##
 
-  def admin_create?
+  def index?
+    user.present?
+  end
+  
+  def show?
+    user.present?
```
jollopre: this policy is too open, isn't? I know they are not going to be able to modify anything but not sure if we want to allow anyone to access to the details of a section from a template that perhaps is under editing still...

app/views/org_admin/sections/_index.html.erb
```diff
@@ -0,0 +1,62 @@
...
+                        section: section, 
+                        edit: true,
+                        current_tab: current_tab } %>
+              <% end %>
+             </div>
+          </div>
+        </div>
+      <% end %>
+    </div>
+    <% if edit || template.customization_of.present? %>
```
jollopre: seeing this reminds me that we should add some sort of method in the template such as:

```
template.customised?
```

Issue #1376, opened 2018-04-11
Add a new Versionable concern


briri:
Add a new Versionable concern that can be used by the Phases, Sections, Questions and Annotations controllers. This concern should use the newly refactored template model's `new_version` method when the user actually updates, creates or destroys a phase, section, question or annotation.

Versioning should occur regardless of whether or not the template is a customization.

Also add tests!

related to #1340

stephaniesimms:
closing issues that can't be tested in the UI

PR #1411, opened 2018-04-24
Issue 1376

jollopre:
- find_in_space and get_new with unit tests associated.
- get_modifiable with tests associated. #1376

### DIFF
--- a/app/controllers/concerns/versionable.rb
+++ b/app/controllers/concerns/versionable.rb
@@ -1,109 +1,106 @@
 module Versionable
   extend ActiveSupport::Concern
 
-  ##
-  # Takes in a Template, phase, Section, Question, or Annotaion
-  # IF the template is published, generates a new template
-  # finds the passed object in the new template
-  # @param obj - Template, Phase, Section, Question, Annotation
-  # @return type_of(obj)
-  def get_modifiable(obj)
-    if obj.respond_to? 'template'
-      template = obj.template
-    elsif obj.is_a? Template
-      template = obj
-    else
-      raise ArgumentError, _('obj should be a Template, Phase, Section, Question, or Annotation') # Throw error as wrong obj added
-    end
-    template, copy = find_or_generate(template)
-    # copy not generated - return obj
-    # copy generated and obj.is_a? template - return template
-    if copy
-      if obj.is_a? Template
-        obj = template
+  included do
+    ##
+    # Takes in a Template, phase, Section, Question, or Annotaion
+    # IF the template is published, generates a new template
+    # finds the passed object in the new template
+    # @param obj - Template, Phase, Section, Question, Annotation
+    # @return type_of(obj)
+    def get_modifiable(obj)
+      if obj.respond_to?(:template)
+        template = obj.template
+      elsif obj.is_a?(Template)
+        template = obj
       else
-        obj = find_in_space(obj,template.phases)
+        raise ArgumentError, _('obj should be a Template, Phase, Section, Question, or Annotation')
       end
-    end
-    return obj
-  end
+      
+      new_template = Template.find_or_generate_version!(template) # raises RuntimeError if template is not latest
 
-  ##
-  # Takes in a phase, Section, Question, or Annotation which is newly
-  # generated and returns a modifiable verson of that object
-  # NOTE: the passed obj is still not saved, but it's parent_id will be updated
-  def get_new(obj)
-    if obj.respond_to? 'template'
+      if new_template != template
+        if obj.is_a?(Template)
+          obj = new_template
+        else
+          obj = find_in_space(obj,new_template.phases)
+        end
+      end
+      return obj
+    end
+    ##
+    # Takes in a phase, Section, Question, or Annotation which is newly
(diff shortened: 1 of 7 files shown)


test/unit/concerns/versionable_test.rb
```diff
@@ -0,0 +1,163 @@
...
+      get_modifiable({})
+    end
+    assert_equal(_('obj should be a Template, Phase, Section, Question, or Annotation'), exception.message)
+  end
+
+  test "#get_modifiable returns same object when template is not published" do
+    # Looking for phase
+    phase = @template.phases.first
+    new_phase = get_modifiable(phase)
+    assert_equal(phase.template_id, new_phase.template_id, 'returns the phase without generating a new template hierarchy')
```
briri: would be nice to test all levels here `[phase, section, question, annotation, question_option].each` which will help catch any issues introduced by changes to those models or their controllers in the future. Or create separate tests like you did for get_new above.

Would also probably be a good idea to compare the object ids as well as the base template's ids

test/unit/concerns/versionable_test.rb
```diff
@@ -0,0 +1,163 @@
...
+  end
+
+  test "#get_modifiable returns new phase when template is published" do
+    @template.published = true
+    @template.save!
+    phase = @template.phases.first
+    new_phase = get_new(phase)
+    assert_not_equal(phase.template_id, new_phase.template_id)
```
briri: see above comment

test/unit/concerns/versionable_test.rb
```diff
@@ -0,0 +1,163 @@
...
+    new_question = get_new(question)
+    assert_not_equal(question.section_id, new_question.section_id)
+  end
+
+  test "#get_new returns new annotation when template is published" do
+    # @template.published = true
+    # @template.save!
+    # annotation = Annotation.new(question: @template.phases.first.sections.first.questions.first)
+    # new_annotation = get_new(annotation)
+    # assert_not_equal(annotation.question_id, new_annotation.question_id)
```
briri: think maybe you forgot to uncomment this test

test/unit/concerns/versionable_test.rb
```diff
@@ -0,0 +1,163 @@
...
+    section = Section.new(phase: @template.phases.first)
+    new_section = get_new(section)
+    assert_equal(section.phase_id, new_section.phase_id, 'returns the section without generating a new template hierarchy')
+    # Looking for question
+    question = Question.new(section: @template.phases.first.sections.first)
+    new_question = get_new(question)
+    assert_equal(question.section_id, new_question.section_id, 'returns the question without generating a new template hierarchy')
+    # Looking for annotation fails
+    # annotation = Annotation.new(question: @template.phases.first.sections.first.questions.first, org: @template.org)
+    # new_annotation = get_new(annotation)
```
briri: Why not fix this so that they are working?

app/controllers/concerns/versionable.rb
```diff
@@ -1,109 +1,114 @@
...
+        when Phase
+          number = obj.phase.number
+          relation = :sections
+        when Section
+          number = obj.section.number
+          relation = :questions
+        when Question
+          number = obj.question.number
+          relation = :annotations
+        else
```
briri: what about `:question_options`?

app/controllers/concerns/versionable.rb
```diff
@@ -1,109 +1,114 @@
...
+        case obj
+          when Phase
+            belongs = :template
+          when Section
+            belongs = :phase
+          when Question
+            belongs = :section
+          when Annotation
+            belongs = :question
+          else
```
briri: We should also version on `QuestionOption` changes too

test/unit/concerns/versionable_test.rb
jollopre: get_new requires separates tests instead of one with multiple assertions because otherwise raises unique key exception trying to generate_version from the same template instance we got from the setup, otherwise I'm inclined to condense all the similar cases in the same test descriptor

briri: If you have a recursive method you should ensure that its able to traverse the hierarchy. The current controller tests are failing because the method returns nil when going beyond the phase level. It would be better to catch that in a test specific to the Versionable concern than to provide phantom errors to controller tests.
Please add them in.

jollopre: I said that is needed to generate separate test case to test the need for creating the entire hierarchy anytime an object (phase, section, question, annotation) wants to be added into a published template. I never said I’m not going to test that the algorithm traverses...

jollopre: I said that is needed to generate separate test case to test the need for creating the entire hierarchy anytime an object (phase, section, question, annotation) wants to be added into a published template. I never said I’m not going to test that the algorithm traverses...

jollopre:
@briri, changes requested are in place now. Note I have moved find_or_generate_version! to template model because is where it really belongs. This method unlike the other approach returns either a template or raises an error if attempting to retrieve or generate a template that is historical for being modified.

PR #1413, opened 2018-04-24
Refactored questions_controller

briri:
For #1374 
- Uses Versionable concern
- Moved routes, controller, and views to the org_admin namespace
- Updated tests
- cleaned up Question model and unit tests

Also updated phases_controller and view to use 'new' instead of 'add' action and added error handling to catch versionable exceptions.

Tests are failing due to issue with current state of Versionable concern. @jollopre merge #1373 first if it looks good and then I will rebase this one. I had to base this one off of 1373 to take advantage of changes to views and routes


app/controllers/org_admin/questions_controller.rb
```diff
@@ -0,0 +1,213 @@
...
+  class QuestionsController < ApplicationController
+    include Versionable
+  
+    respond_to :html
+    after_action :verify_authorized
+
+    # GET /org_admin/templates/[:template_id]/phases/[:phase_id]/sections/[:section_id]/questions
+    def index
+      authorize Question.new
+      section = Section.includes(:questions, phase: :template).find(params[:section_id])
```
jollopre: do we need to pull the section, the phase and the template to display a bunch of questions?

app/controllers/org_admin/questions_controller.rb
```diff
@@ -0,0 +1,213 @@
...
+          current_tab: params[:r] || 'all-templates'
+        }
+    end
+    
+    # POST /org_admin/templates/[:template_id]/phases/[:phase_id]/sections/[:section_id]/questions
+    def create
+      section = Section.includes(:questions, phase: :template).find(params[:section_id])
+      question = Question.new(question_params.merge({ section_id: section.id }))
+      current_tab = params[:r] || 'all-templates'
+      authorize question
```
jollopre: Please, change the order of these instructions to be something like this:

```
question = Question.new(question_params)
authorize question
section = Section.includes(:questions, phase: :template).find(params[:section_id])
current_tab = params[:r] || 'all-templates'
```
so that we don't run three queries four queries to retrieve section, questions, phase and template until pundit returns true

jollopre:
Apparently there are conflicts, please solve them and add my suggestion

Issue #1405, opened 2018-04-23
Refactor Template index UI


briri:
Refactor the template index view
- Create a `container` view that contains the affiliation change portion and tabbed navigation
- Separate tabs into multiple views that use the `container` as a layout
- Remove need for `lib/template_filter.rb` if possible and stop passing around hashes
- Refactor JS to work with changes

stephaniesimms:
closing issues that can't be tested in the UI

PR #1432, opened 2018-04-30
Admin Templates page rework

briri:
for #1405
- Separated templates page tabs into separate endpoints
- Updated pages so that they no longer use hashes
- Added an after update callback to set other versions to unpublished when publishing a template
- Remove need to pass around the currently selected tab
- Updated paginable to allow for non-pagination query params (e.g. the template filters of Published, All, etc.)
- Switched wording to 'Draft' to match history page for templates who have a published version but have new changes that have not yet been published (up for debate/discussion)
- Fixed tests

Did a bunch of manual testing for this. We'll likely find some minor issues as it starts to get heavily tested.


app/controllers/org_admin/templates_controller.rb
```diff
@@ -2,48 +2,91 @@ module OrgAdmin
...
-      
-      render 'index', locals: {
-        all_templates: all_templates_hash[:templates],
-        customizable_templates: customizable_hash[:templates], 
-        own_templates: own_hash[:templates],
-        customized_templates: customizable_hash[:customizations],
-        published: published,
-        current_org: current_user.org, 
+      templates = Template.latest_version
+      published, draft = templates.partition{|t| t.published? }
```
jollopre: nice trick!

app/controllers/org_admin/templates_controller.rb
```diff
@@ -2,48 +2,91 @@ module OrgAdmin
...
     after_action :verify_authorized
 
+    # The root version of index which returns all templates
     # GET /org_admin/templates
     # -----------------------------------------------------
     def index
       authorize Template
-      
-# TODO: Refactor this method and its views in next sprint if time allows so that
-#       we are no longer passing around hashes and do not need the template_filter.rb
```
jollopre: relief! Thanks!

app/controllers/org_admin/templates_controller.rb
```diff
@@ -212,40 +251,32 @@ def copy
...
-      
-      # Only allow the current version to be updated
-      if template.latest?
-        # Unpublish the older published version if there is one
-        live = Template.live(template.family_id)
-        if !live.nil? and self != live
-          live.published = false
-          live.save!
-        end
+      # Allowing to publish any version here for future compatibility with publishing older versions from the history page
```
jollopre: this might be problematic. Remember, publishing a historic version means that attempting to generate a new version after will fail since family_id and version are already inserted in the templates table

app/models/scopes/template_scope.rb
```diff
@@ -63,6 +63,24 @@ module TemplateScope
         unarchived.where(customization_of: nil).distinct
       end 
     }
+    # Retrieves the latest version of each customizable template
+    scope :latest_customizable, -> { 
+      family_ids = unarchived.where('org_id IN (?) OR is_default = ?', Org.where(org_type: Org.org_type_values_for(:funder).min).collect(&:id), true).pluck(:family_id)
```
jollopre: you may want to add distinct(:family_id) before.pluck otherwise we can retrieve duplicate family_id when a template is versioned... 
Also, you could reuse the already defined latest_version_per_org:
```
Template.latest_version_per_org(
Org.where(org_type: Org.where(org_type: Org.org_type_values_for(:funder).min).collect(&:id))
```
2. add a scope method for the edge case (default_template):
```
scope :latest_default, -> {
latest_version(unarchived.where(is_default: true).pluck(:family_id))
}
```
3. Chain them together where is needed

app/models/scopes/template_scope.rb
```diff
@@ -63,6 +63,24 @@ module TemplateScope
...
+      latest_version(family_ids)
+    }
+    # Retrieves the latest templates, i.e. those with maximum version associated for a set of org_id passed
+    scope :latest_version_for_org, -> (org_id=nil) {
```
jollopre: Did you see the scope method latest_version_per_org defined above?

app/models/scopes/template_scope.rb
```diff
@@ -63,6 +63,24 @@ module TemplateScope
...
+      org_id = [org_id] unless org_id.respond_to?(:each)
+      family_ids = where(org_id: org_id).pluck(:family_id)
+      latest_version(family_ids).where(org_id: org_id)
+    }
+    # Retrieves the latest templates, i.e. those with maximum version associated for a set of org_id passed
+    scope :latest_customized_version_for_org, -> (org_id=nil) {
```
jollopre: did you see latest_customized_version defined above?

app/models/template.rb
```diff
@@ -5,6 +5,7 @@ class Template < ActiveRecord::Base
   validates_with TemplateLinksValidator
 
   before_validation :set_defaults 
+  after_update :reconcile_published, if: Proc.new { |template| template.published? && template.version.present? && template.version > 0 }
```
jollopre: why those additional checks after published? template.version should always be present and greater than zero

app/models/template.rb
```diff
@@ -213,49 +214,6 @@ def upgrade_customization!
     return customization
   end
 
-  ##
```
jollopre: glad this is gone!

app/models/template.rb
```diff
@@ -278,4 +236,12 @@ def set_defaults
       self.archived ||= false
       self.links ||= { funder: [], sample_plan: [] }
     end
+    
+    # Only one version of a template should be published at a time, so if this one was published make sure other versions are not
+    def reconcile_published
+      if self.published?
```
jollopre: if the method is intended to be called when the callback is fired and passes the condition, why adding again the if here?

app/policies/template_policy.rb
```diff
@@ -8,9 +8,17 @@ def initialize(user, template = Template.new)
   end
   
   def index?
-    user.can_super_admin? || user.can_modify_templates?
+    user.can_super_admin?
```
jollopre: org_admins should not see index page?

app/views/org_admin/templates/_show.html.erb
```diff
@@ -10,10 +10,10 @@
   <dd>
     <% if template.published? %>
       <%= _('Published') %>
-    <% elsif template.version == 0 %>
+    <% elsif template.version <= 0 %>
       <%= _('Unpublished')%>
     <% else %>
-      <%= _('You have unpublished changes') %>
+      <%= _('Draft') %>
```
jollopre: - unpublished is just published to false and
- draft is whenever a template version is greater than the published for the same family_id

am I right?

app/views/paginable/templates/_customisable.html.erb
```diff
@@ -0,0 +1,78 @@
...
+      <tr>
+        <td>
+          <%= template.title %>
+        </td>
+        <td>
+          <%= template.org.name %>
+        </td>
+        <td>
+          <% if customization.present? %>
+            <% if customization.created_at < template.created_at %>
```
jollopre: customization.upgrade_customization? does this logic already...

lib/template_filter.rb
```diff
@@ -1,90 +0,0 @@
-module TemplateFilter
```
jollopre: thanks for removing this

app/controllers/org_admin/templates_controller.rb
briri: ah true. ok. I will put the check in to make sure its the latest

app/models/template.rb
briri: just so that it doesn't unecessarily execute the callback for the first time the template is published. Although, I suppose the select statement in the callback would not find any versions to unpublish so it wouldn't be an issue.

PR #1435, opened 2018-05-01
tweaks to templates index pages

briri:
Changes recommended for PR #1432 .

Also updated status of templates in tables. A template is now either Published or Draft to simplify things. If its really important to indicate that there are no published versions for a given family we can re-implement but it would require running n-queries to get the `published` for each family


briri:
closing to make some adjustments to default template designation

Issue #1434, opened 2018-05-01
Update history page


briri:
Update history page to work with new index page and controller

stephaniesimms:
closing issues that can't be tested in the UI

PR #1438, opened 2018-05-02
Refactored History page 

briri:
For #1434 

Updated history page to use new template controller and model. Also made it so that you can view the history of a customization

### DIFF
--- a/app/views/org_admin/templates/history.html.erb
+++ b/app/views/org_admin/templates/history.html.erb
@@ -1,7 +1,7 @@
 <div class="row">
   <div class="col-md-12">
     <h1>
-      <%= _('Template History') %>
+      <%= templates.first.customization_of.present? ? _('Template Customisation History') : _('Template History') %>
       <div class="pull-right">
         <%= link_to _('View all templates'), referrer, class: "btn btn-primary" %>
       </div>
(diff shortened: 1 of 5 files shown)


app/views/org_admin/templates/history.html.erb
```diff
@@ -18,10 +18,9 @@
         partial: '/paginable/templates/history',
         controller: 'paginable/templates',
         action: 'history',
-        path_params: { id: template.id },
-        query_params: { sort_field: :version, sort_direction: :desc }, 
+        query_params: { sort_field: 'templates.version', sort_direction: 'desc' },
```
jollopre: I remember I placed path_params when I added pagination for this action otherwise how do you know how to filter the history of templates for a specific template id? I think the pagination is broken if that line is removed...

app/views/org_admin/templates/history.html.erb
```diff
@@ -1,7 +1,7 @@
 <div class="row">
   <div class="col-md-12">
     <h1>
-      <%= _('Template History') %>
+      <%= templates.first.customization_of.present? ? _('Template Customisation History') : _('Template History') %>
```
jollopre: This title change seems more accurate to me however not sure if this will confuse users.

PR #1458, opened 2018-05-03
Add overview of phases to templates page

briri:
Adding the phase/section/question overview back to the Template details tab.

### DIFF
--- a/app/views/org_admin/phases/_overview.html.erb
+++ b/app/views/org_admin/phases/_overview.html.erb
@@ -1,0 +1,70 @@
+<br>
+<h2><%= _('Template Overview') %></h2>
+<% phases.each do |phase| %>
+  <hr>
+  <div class="row">
+    <div class="col-md-12">
+      <div>
+        <h3>
+          <%= phase.title %>
+          <div class="pull-right">
+            <% if template.customization_of.nil? %>
+              <%= link_to _('Edit phase'), edit_org_admin_template_phase_path(template.id, phase.id), { class: "btn btn-default", role: 'button' } %>
+            <% else %>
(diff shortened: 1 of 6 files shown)


app/views/org_admin/phases/_overview.html.erb
```diff
@@ -0,0 +1,70 @@
...
+  <div class="row">
+    <div class="col-md-12">
+      <div>
+        <h3>
+          <%= phase.title %>
+          <div class="pull-right">
+            <% if template.customization_of.nil? %>
+              <%= link_to _('Edit phase'), edit_org_admin_template_phase_path(template.id, phase.id), { class: "btn btn-default", role: 'button' } %>
+            <% else %>
+              <%= link_to _('Customize phase'), org_admin_template_phase_path(template.id, phase.id), { class: "btn btn-default", role: 'button' } %>
```
jollopre: don't think we need this distinction here. The wording 'customize' I would use only for when you want to copy a funder template to add things not when you already has a template that might or might not be a copy from a funder. What do you think if we leave it like it was?

app/views/org_admin/phases/_overview.html.erb
```diff
@@ -0,0 +1,70 @@
...
+        <div class="table-responsive">
+          <table class="table table-hover">
+            <thead>
+              <tr>
+                <th><%= _('Sections')%></th>
+                <th><%= _('Questions')%></th>
+              </tr>
+            </thead>
+            <tbody>
+              <% phase.sections.each do |section| %>
```
jollopre: splitting this into multiple partials will increase readability and perhaps maintenance too.

PR #1461, opened 2018-05-07
Issue 1377

jollopre:
annotations moved to org_admin namespace. DMPRoadmap/roadmap#1377


config/routes.rb
```diff
@@ -284,12 +277,13 @@
           
           resources :sections, only: [:index, :show, :edit, :new, :update, :create, :destroy] do
             resources :questions, only: [:index, :show, :edit, :new, :update, :create, :destroy] do
-              
             end
           end
         end
       end
       
+      resources :annotations, only: [:create, :destroy, :update] do ; end
```
briri: You don't need the `do ; end` here.
Also, why not make this a child of the templates->phases->sections hierarchy above?

jollopre: because an annotation not only requires a question but also an org in order to be valid and since both cases cannot be handled as a nested route, I consider is best to place it alone and require them with the body params.

briri:
ah true

Issue #1420, opened 2018-04-25
Clean unused Gems from the gemfile


bhavi:
In the process of Ruby Upgrade v2.4.1 its advisable to first clean up the unused Gems from the Gemfile.
Related to #826. 

This would include but not limited to:
#rolify for roles
gem 'rolify', '~> 5.1.0'

#API
gem 'swagger-docs', '>= 0.2.9 '

CODE DOCUMENTATION
gem 'yard', '>= 0.9.5'
gem 'redcarpet', '>= 3.3.4'

briri:
You should also be able to get rid of the following:
- friendly_id
- yaml_db
- capybara (also try dropping the others in that section (not sure we're using them all)
- jbuilder (@jollopre and @xsrust is this one really needed for the API?)

bhavi:
Skipping 'jbuilder' and 'administrate' gems for now as they are still used in the API.

stephaniesimms:
closing issues that can't be tested in the UI

PR #1464, opened 2018-05-08
Ruby Upgrade v2.4.1 (#826, #1420)

### LINKED ISSUE #826: Ruby 2.2 -> 2.4 upgrade
Ruby 2.2 is at EOL

We will need to upgrade Rails at the same time as some of the framework's gems will have conflicts.
Discovered the following json dependency issue when working with Ruby 2.4 and Rails 4.2.7 on a server here:

https://stackoverflow.com/questions/46042900/ruby-json-gem-install-error-ruby-2-4-1-json-1-8-6-windows-10

bhavi:
cleaning unused gems and upgrade to Ruby v2.4.1
fix node modules issues when running the upgrade
update travis.yml to Ruby v2.4.1

Fixes #826 and #1420 .

Changes proposed in this PR:
-


vyruss:
@bhavi appears to work fine on dmponline-dev. Is there any reason to use 2.4.1 and not 2.4.4 which is the latest stable 2.4 ruby?

PR #1480, opened 2018-05-10
Finished up questions/annotations

briri:
Finished up refactor of questions and annotations

### DIFF
--- a/app/controllers/org_admin/phases_controller.rb
+++ b/app/controllers/org_admin/phases_controller.rb
@@ -11,16 +11,14 @@ module OrgAdmin
       if !phase.template.latest?
         flash[:notice] = _('You are viewing a historical version of this template. You will not be able to make changes.')
       end
-      section_id = params.fetch(:section_id, nil)
-      question_id = params.fetch(:question_id, nil)
+      section = params.fetch(:section, nil)
       render('container',
         locals: { 
           partial_path: 'show',
           template: phase.template,
           phase: phase,
           sections: phase.sections.order(:number).select(:id, :title),
-          current_section: section_id.present? ? Section.find_by(id: section_id, phase_id: phase.id) : nil,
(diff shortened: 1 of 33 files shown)


app/views/org_admin/phases/container.html.erb
```diff
@@ -8,7 +9,7 @@
 <div class="row">
   <div class="col-md-12">
     <!-- render navigation tabs for the template-->
-    <%= render partial: "/org_admin/templates/navigation", locals: local_assigns %>
+    <%= render partial: "/org_admin/templates/navigation", locals: local_assigns.merge({ modifiable: modifiable }) %>
```
xsrust: since we do this .merge each time we render a partial, why not just save the change to local_assigns?

lib/assets/javascripts/views/org_admin/questions/new_edit.js
```diff
@@ -2,17 +2,23 @@ import { Tinymce } from '../../../utils/tinymce';
...
 import onChangeQuestionFormat from './sharedEventHandlers';
+import initAnnotations from '../../annotations/form';
 
 export default (context) => {
   const target = $(context);
   if (isObject(target)) {
-    Tinymce.init({ selector: `${target.attr('id')} .question` });
-    ariatiseForm({ selector: `${target.attr('id')} .question_form` });
+    // Wire up the Question and then any annotations
+    Tinymce.init({ selector: `#${context} .question` });
```
jollopre: why do want target variable then?

app/controllers/org_admin/phases_controller.rb
```diff
@@ -93,22 +89,21 @@ def create
 
 # TODO: update this so that the description comes through as part of the normal form attributes [:phase][:description]
         phase.description = params["phase-desc"]
-        current_tab = params[:r] || 'all-templates'
-
+        phase.modifiable = true
```
jollopre: I would be more inclined to have phase defaults set at phase model...

PR #1482, opened 2018-05-10
commented out version check on template update

### LINKED ISSUE #1476: Editing template details creates new version
I can open a template or customization without triggering a new version, but if I edit any template details and click Save it creates a new version (status change to "Draft" - screenshot below). Desired behavior is to automatically update the published template details with the saved changes but not create a new version. Only structural changes to phase-section-questions should trigger a new version.

template-details-versions

briri:
Fixes #1476 

Commented out the call to `Template.find_or_version` for `templates_controller.update`. Left the line there in case we decide that some fields should trigger a new version (e.g. title)


xsrust:
looks good, thanks

PR #1488, opened 2018-05-14
cleaned up remaining issues with template editing and customization

briri:
Lots of minor changes to views, JS and controller code to address remaining issues found when creating editing templates and customizations.


app/controllers/org_admin/questions_controller.rb
```diff
@@ -79,15 +73,39 @@ def destroy
...
       end
+      
+      # When a template gets versioned by changes to one of its questions we need to loop 
+      # through the incoming params and ensure that the annotations and question_options
+      # get attached to the new question
+      def transfer_associations(question)
+        attrs = question_params
+        if attrs[:annotations_attributes].present?
+          attrs[:annotations_attributes].each_key do |key|
+            old_annotation = question.annotations.select do |a|
```
jollopre: Use find instead of select since it will return the annotation object or nil instead of an array with all the matches. It also works as a search instead of iterating all the annotation for every key at annotations_attributes....

app/controllers/org_admin/sections_controller.rb
```diff
@@ -46,23 +46,20 @@ def edit
 
     # POST /org_admin/templates/[:template_id]/phases/[:phase_id]/sections
     def create
-      phase = Phase.includes(:template, :sections).find(params[:phase_id])
-      section = Section.new(section_params.merge({ phase_id: phase.id }))
+      section = Section.new(section_params.merge({ phase_id: params[:phase_id] }))
```
jollopre: this could create a data integrity issue since you are removing the phase find which automatically raises an ActiveRecord::RecordNotFound when an invalid phase id is passed...

app/controllers/org_admin/templates_controller.rb
```diff
@@ -133,47 +135,36 @@ def create
...
     # DELETE /org_admin/templates/:id
     # -----------------------------------------------------
     def destroy
       template = Template.find(params[:id])
       authorize template
-      if template.plans.length <= 0
-        # Only allow the current version to be destroyed
-        if template.latest?
-          if template.destroy
+      versions = Template.includes(:plans).where(family_id: template.family_id)
```
jollopre: the action should destroy a template given an id therefore you only have to check whether or not that template has plans associated. If it has plans, you can't destroy it... I don't understand why do you want to destroy all its versions too when they have no plans...
If we really want that case, I'm more inclined to create another action like 'destroy_all_versions' to accommodate what you are trying to do here

app/controllers/org_admin/templates_controller.rb
```diff
@@ -269,13 +262,13 @@ def publish
...
       template = Template.find(params[:id])
       authorize template
-      if template.latest?
-        template.published = false
-        template.save!
-        flash[:notice] = _("Your #{template_type(template)} is no longer published. Users will not be able to create new DMPs for this #{template_type(template)} until you re-publish it")
-      else
-        flash[:alert] = _("That #{template_type(template)} is not currently published.")
+      versions = Template.where(family_id: template.family_id)
+      versions.each do |version|
```
jollopre: why do we want this https://github.com/DMPRoadmap/roadmap/blob/sprint2/app/models/template.rb#L8 then?

briri: Because the user is always presented with the latest version in the templates table. The desire is to allow an admin to unpublish a template whether or not the published version is the latest.

e.g. I am an admin and start editing my published template which creates a draft version. I then realize that I want to unpublish the 'live' one. We currently offer no way to do that.

app/controllers/org_admin/questions_controller.rb
briri: switiched it over but it started failing. Going to leave as-is for now. We can adjust in a separate PR afterward since I'd like to get this out for UAT

app/controllers/org_admin/sections_controller.rb
briri: Yes good catch. Added a check in for the phase existence

briri:
@jollopre happy to reassess the publish/unpublish changes if they prove to be an issue. We can wait for UAt feedback and submit in a separate PR if changes are required

app/controllers/org_admin/templates_controller.rb
jollopre: Where is that conversation tracked?

Issue #1483, opened 2018-05-11
I cannot download my plan after submitting for feedback


stephaniesimms:
This is related to #1430 and can be addressed at the same time

As a user when I click the 'Request feedback' button on the Share tab while writing my plan, then Download tab disappears (screenshot below) and I can no longer download my plan until an admin finishes providing feedback. The desired behavior is for plan authors to always have the ability to download plans.

#1430 referenced above should enable admins to download the plan if feedback has been requested

![no-download](https://user-images.githubusercontent.com/12703638/39939441-2792892c-550b-11e8-9cd5-1683b9dd61d6.png)

sjDCC:
Can still download. Closing out

Issue #1430, opened 2018-04-30
Org Admin cannot download plans when they are being reviewed


rosiehigman:
**Expected behaviour:**
When an Org Admin looks at a DMP they see a download tab where they can export the DMP content as pdf, docx etc.

**Actual behaviour:**
If the DMP creator has requested feedback the download option is no longer visible for the Org Admin. 
The download option is available for DMPs where the creator has not requested a review.

**Steps to reproduce:**
Request expert feedback
As an Org Admin go to 'Plans' under the 'Admin' menu, select the DMP which requires feedback and the Download tab is not visible.
![nodownloadoption](https://user-images.githubusercontent.com/15738325/39436550-31995ee6-4c96-11e8-9f83-18abc12c6b7c.JPG)

stephaniesimms:
this is definitely not working as expected/desired - the opposite in fact. the idea was to provide more access to plans that users invited admins to comment on but minimize access from the main plans table in the admin area. we should just provide the ability to download plans.

briri:
Fixed in #1483

PR #1493, opened 2018-05-16
Issue1483

jollopre:
any user with readable rights should see download tab for a plan. DMPRoadmap/roadmap#1483


jollopre:
This PR should address #1483 and #1430. The root cause of the problem was the visibility rule to display the download link. This rule was presenting the link unless the user had a reviewer role on a plan. Since every org_admin gets assigned that role any time a feedback is requested, they were not able to see the link (the access to download page was still available if link was known anyway).

Issue #1473, opened 2018-05-09
Create plan wizard provides all funder templates, no default template


stephaniesimms:
This bug seems related to issue #1470 
When I select a funder for question 3 in the create plan wizard I get a dropdown containing all funder templates regardless of my selection (screenshot 1). there is no association between funder and template.

if i check the boxes for no Org and no Funder for questions 2 and 3 I get the error in screenshot 2. it can't find the default template

if my Org has templates then I do get those when I make the appropriate selection. However if I try to Write plan I get the error in screenshot 3 (same missing template error from #1470)

![screen shot 2018-05-09 at 1 12 18 pm](https://user-images.githubusercontent.com/12703638/39837636-e5e75a5a-538b-11e8-9278-c3aa350f57e6.png)

![default](https://user-images.githubusercontent.com/12703638/39837442-646da3bc-538b-11e8-9da3-440f6bc43a59.png)

![screen shot 2018-05-09 at 1 20 14 pm](https://user-images.githubusercontent.com/12703638/39837572-bdd753da-538b-11e8-9ee7-69c2e1d8a94b.png)

stephaniesimms:
I can get the default template now, but I still see the full list of funder template in a dropdown regardless of which funder I select @briri

sjDCC:
This all seems to be working fine now. Closing out

PR #1500, opened 2018-05-16
fixed issues with create plan template selection and changing plan vi…

briri:
For #1473 
- Fixed issues with what templates are presented to the user on create plan page
- Fixed issue with organizationally visible appearing in the customizable templates tab
- Fixed issue with changing the template visibility
- Added extensive tests to cover all of the create plan scenarios (e.g. default template, no funder, etc.)

### DIFF
--- a/app/models/scopes/template_scope.rb
+++ b/app/models/scopes/template_scope.rb
@@ -74,7 +74,7 @@ module TemplateScope
     # Retrieves the latest version of each customizable funder template (and the default template)
     scope :latest_customizable, -> {
       family_ids = families(Org.funder.collect(&:id)).distinct.pluck(:family_id) << default.family_id
-      published(family_ids.flatten)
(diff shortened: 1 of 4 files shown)


app/models/scopes/template_scope.rb
```diff
@@ -74,7 +74,7 @@ module TemplateScope
     # Retrieves the latest version of each customizable funder template (and the default template)
     scope :latest_customizable, -> {
       family_ids = families(Org.funder.collect(&:id)).distinct.pluck(:family_id) << default.family_id
-      published(family_ids.flatten)
+      published(family_ids.flatten).where('visibility = ? OR is_default = ?', visibilities[:publicly_visible], true)
```
jollopre: flatten is not needed at all since family_ids is already a unidimensional array. The where clause for visibility seems fine however don't see the need for adding again is_default condition...

PR #1506, opened 2018-05-16
Fix 'back' button on preview so that it goes to appropriate edit/show

### LINKED ISSUE #1477: Cannot preview template or customization
When I click the Preview button for a template or customization I get the error below

preview-customiz-error

briri:
Fixes #1477 .
- Added checks for `template.latest?` and `customization_of` when determining which back button to present to user
- Added redirect on phases_controller.edit to phases_controller.show if edit page is accessed but the template is a customization

### DIFF
--- a/app/controllers/org_admin/phases_controller.rb
+++ b/app/controllers/org_admin/phases_controller.rb
@@ -29,15 +29,20 @@ module OrgAdmin
       if !phase.template.latest?
         flash[:notice] = _('You are viewing a historical version of this template. You will not be able to make changes.')
       end
-      section = params.fetch(:section, nil)
-      render('container',
-        locals: { 
-          partial_path: 'edit',
-          template: phase.template,
-          phase: phase,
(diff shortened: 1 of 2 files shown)


app/controllers/org_admin/phases_controller.rb
```diff
@@ -29,15 +29,20 @@ def edit
...
-      render('container',
-        locals: { 
-          partial_path: 'edit',
-          template: phase.template,
-          phase: phase,
-          sections: phase.sections.order(:number).select(:id, :title, :modifiable),
-          current_section: section.present? ? Section.find_by(id: section, phase_id: phase.id) : nil
-        })
+      # User cannot edit a phase if its a customization so redirect to show
+      if phase.template.customization_of.present?
```
jollopre: I think this is too generic. A customizable template might have phases that do not belong to the customization_of template and therefore we should offer the possibility to edit them, am I right?

Issue #1495, opened 2018-05-16
Paginable actions not working on templates pages


briri:
The search and sort functions are not working on the templates pages: owned and customizable (when there are more than 10 templates in the list). 

Clicking on sorts or search results in which seems to be an issue with the chaining of scopes:
``ActiveRecord::StatementInvalid - Mysql2::Error: Unknown column 'orgs.name' in 'order clause': SELECT  `templates`.* FROM `templates` WHERE `templates`.`archived` = 0 AND `templates`.`published` = 1 AND `templates`.`family_id` IN (1874, 1875, 1876, 1877, 1878, 1879, 1880, 1881, 1882, 1883, 1884, 1885, 1886, 1887, 1888, 1889, 1890, 1891, 1892, 1894, 1895, 1896, 1903, 1907, 1911, 1912, 1932, 1936, 1940, 1941, 1947, 1949, 1950, 1951, 1952, 1953, 1962, 1964, 1965, 1971, 1977, 1979, 1984, 1986, 1992, 1993, 2000, 2001, 2005, 2008, 2009, 2016, 2019, 538382351, 756435778, 1936) AND `templates`.`visibility` = 1  ORDER BY orgs.name ASC LIMIT 10 OFFSET 0:```

jollopre:
@sjDCC I sent a PR #1511 before that solves the problem when attempting to sort by 'funder' or org however it is still pending to be reviewed/deployed...

stephaniesimms:
I can sort by all columns except Status. The Status column does sort properly when I test as a Super Admin only for the "All templates" tab but not the other two tabs. It doesn't sort for either tab when I test as an Org admin.

And now the search function is no longer working for a list of more than 10 items in the same pattern described for Status column sorting above. I can create a new issue for this bug if it should be handled separately?

briri:
No, its likely related. All of those links fire Javascript. An unfortunate side-effect of many javascript failures is that all remaining javascript on the page stops working. I think its probably what's happening.

What steps did you take to create the issue? Click sort on the status column and then search?

stephaniesimms:
i was toggling across tabs so no particular order to produce the issue with Status sort and Search. Every time I test them in whatever order neither works.

jollopre:
[...]

PR #1510, opened 2018-05-17
Issue1495

jollopre:
adding org to chained scope in order to have searchable and sortable templates. DMPRoadmap/roadmap#1495


jollopre:
@briri, note that only index and customisable actions have to 'include' orgs table for sorting/searching. 
The index action was not failing because the scoped query 'latest_version' joins orgs whereas 'latest_customizable' was missing that bit. I decided to chain with includes org at the end so that we avoid n+1 queries when displaying the org's name.

briri:
looks good @jollopre thanks for fixing it

PR #1515, opened 2018-05-18
Reintroduce edit/cancel buttons for questions

### LINKED ISSUE #1502: Creating and editing questions refinements
As an admin creating new questions for templates, a couple of behaviors are different now and should be revised to function as they did previously.

1. the default selection for Answer format should be 'Text area' (currently it defaults to Check box; screenshot 1)
2. after I enter text and click Save, the system should close the editors and just display my entries in read only. current behavior keeps all editors open and active, even when I click to create a new question below the first. this is confusing and produces a huge cascade of editors for a full template.
3. I noticed the same behavior when I click to edit an existing template question - all editors remain open. Desired behavior is to display existing content as read only and only open editors if a user clicks to Edit a question.

default-answer-type-textarea

briri:
Fixes #1502 
- Updated questions 'index' partial to always display questions as 'show' on section load
- Added a 'Edit' button to question 'show' partial
- Added a 'Cancel' button to question 'form' partial
- Updated JS to load 'edit', 'new' and 'show' partials via AJAX when the user clicks the above buttons. For some reason Rails' `remote: true` does not work here (maybe a name/selector collision with other remote forms/links elsewhere on the page?)
- Added 'Loading...' and failure messages for instances where AJAX is loading the form or is unable to load the form


briri:
@jollopre I'll need you to take a look at this one. I'm not seeing what the issue is. Scenario (you can test on roadmap-stg):
1. Open a template phase for editing
2. Expand a section (all questions should now display as 'show' partials with a new 'Edit' button)
3. Click 'Edit' on one of the questions ('show' partial should be replaced with 'edit' partial)
4. Scroll to the bottom of the form and click 'Cancel' ('show' partial should get reloaded)
5. Click 'Edit' again on the same question ('edit' partial reloads but Tinymce editors are not loaded)

The Tinymce editors get wired up properly the first time the questions 'form' partial is loaded when you click 'Edit' (same is true for 'Add question') but not on subsequent loads. All other JS is working properly. 
I tried calling the `Tinymce.destroy...` methods when the user clicks 'Cancel' on the form before replacing the contents of the container but it didn't help.

app/views/org_admin/questions/_index.html.erb
```diff
@@ -2,25 +2,21 @@
...
-        </div>
+    <div class="col-md-12">
+      <h4 style="display: none"><%= _('New Question') %></h4>
+      <div class="pull-right">
+        <%= link_to(_('Add Question'), new_org_admin_template_phase_section_question_path(template_id: template.id, phase_id: section.phase.id, section_id: section.id), { class: 'btn btn-default question_new_link', role: "button" }) %>
       </div>
     </div>
-<% end %>
\ No newline at end of file
+    <div class="col-md-12 question_new"></div>
```
jollopre: I will load the partial directly here when this partial is rendered instead of loading through AJAX. First because it is a small view and secondly because always will be the same object, remember we reload the entire page anytime a change happens so AJAXify here is a bit of over-engineering...

app/views/org_admin/sections/_index.html.erb
```diff
@@ -38,7 +33,7 @@
               <%# This is AJAX loaded on demand unless section_id is specified in the URL %>
               <% if current_section.present? && section.id == current_section.id %>
                 <% partial = "org_admin/sections/#{section.modifiable? ? 'edit' : 'show'}" %>
-                <%= render partial: partial, 
+                <%= render partial: "#{partial}",
```
jollopre: If we are now always displaying 'show' do we need the variable partial here? If I remember well I added this logic to display the active section (if any passed through URL)...

PR #1529, opened 2018-05-22
Fixed issue with editing new sections added to a customization

### LINKED ISSUE #1521: Can't go back and edit questions when customizing template
While customizing a template, once I create a question I can't go back and edit it. The only content that's available for editing are the "Annotations" (Example answer and Guidance). It's also not apparent what I should do to edit questions since the only button available to me is "Save." When I click Save below a specific question it opens the Section details for editing but I can never return to edit individual questions. We need to include an Edit button (same as template creation button options) that permits users to open the full question editing page (screenshot 2) and edit their questions, reorder them, etc.

screen shot 2018-05-21 at 11 06 40 am

Customize template options for creating and editing Questions should be the same as template edit options (below) with an Edit button that allows users to go back and edit questions. Save - Cancel - Delete buttons should also be available in both areas while a user is editing a question. Currently these buttons are only available when editing a template (not a customization).
edit-question

briri:
Fixes #1521 
- Add modifiable? check to allow user to edit customized sections

Also fixed the following:
- Update controller to auto expand the specified section
- Added confirmation message for deleting a question

### DIFF
--- a/app/views/org_admin/sections/_index.html.erb
+++ b/app/views/org_admin/sections/_index.html.erb
@@ -4,8 +4,9 @@
     <div class="panel-group" id="sections_accordion" role="tablist">
       <% sections.each do |section| %>
         <div class="panel panel-default">
+          <% href = (section.modifiable? ? edit_org_admin_template_phase_section_path(template_id: template.id, phase_id: phase.id, id: section.id) : org_admin_template_phase_section_path(template_id: template.id, phase_id: phase.id, id: section.id)) %>
           <a
-            href="<%= org_admin_template_phase_section_path(template_id: template.id, phase_id: phase.id, id: section.id) %>"
+            href="<%= href %>"
             data-remote="true"
             class="heading-button ajaxified-section"
             data-toggle="collapse"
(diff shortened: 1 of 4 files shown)


app/views/org_admin/sections/_index.html.erb
```diff
@@ -4,8 +4,9 @@
     <div class="panel-group" id="sections_accordion" role="tablist">
       <% sections.each do |section| %>
         <div class="panel panel-default">
+          <% href = (section.modifiable? ? edit_org_admin_template_phase_section_path(template_id: template.id, phase_id: phase.id, id: section.id) : org_admin_template_phase_section_path(template_id: template.id, phase_id: phase.id, id: section.id)) %>
           <a
-            href="<%= org_admin_template_phase_section_path(template_id: template.id, phase_id: phase.id, id: section.id) %>"
+            href="<%= href %>"
```
jollopre: The link now behaves differently when a section is modifiable which is fine but I though we want by default to open the section details in show mode and add edit button if the section is modifiable. I understood that from #1502

PR #1531, opened 2018-05-22
Fixed paginable query for customized templates table

### LINKED ISSUE #1519: Can't search 'Customizable templates' list
When the table has more than 10 items the search function isn't working at all in the Customizable templates list. As a Super Admin, search does work for All templates but not Customizable templates. And as an Org Admin it doesn't work for Customizable templates, no matter what I input.

briri:
Fixes #1519

### DIFF
--- a/app/controllers/paginable/templates_controller.rb
+++ b/app/controllers/paginable/templates_controller.rb
@@ -49,7 +49,7 @@ class Paginable::TemplatesController < ApplicationController
       when 'not-customised'
         templates = Template.latest_customizable.where.not(family_id: customizations.collect(&:customization_of))
     end
-    paginable_renderise partial: 'customisable', scope: templates.includes(:org), locals: { action: 'customisable', customizations: customizations }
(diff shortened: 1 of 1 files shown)


app/controllers/paginable/templates_controller.rb
```diff
@@ -49,7 +49,7 @@ def customisable
       when 'not-customised'
         templates = Template.latest_customizable.where.not(family_id: customizations.collect(&:customization_of))
     end
-    paginable_renderise partial: 'customisable', scope: templates.includes(:org), locals: { action: 'customisable', customizations: customizations }
+    paginable_renderise partial: 'customisable', scope: templates.joins(:org), locals: { action: 'customisable', customizations: customizations }
```
jollopre: why not 
```
templates.joins(:org).includes(:org)
```
will permit the search to work but also will pre-fetch the orgs for the view since it uses the org.name for every template...

PR #1545, opened 2018-05-23
Fixed issue with Search on public Templates page

### LINKED ISSUE #1541: Search not working on DMP Templates table
I can't search the DMP Templates table - nothing I input produces any results
https://roadmap-stg.cdlib.org/public_templates

briri:
Fixes #1541 
Added `.joins(:org)` to query in templates_controller

Also realized that a table removal from ruby-upgrade was not reflected in schema.rb , so updated schema.rb to remove `friendly_id_slugs` (gem was removed) and add in the foreign key indices


briri:
Very minor bug fix and update of schema.rb to fix a lost change from ruby-upgrade (I hadn't run db:migrate locally before finalizing the ruby_upgrade branch PR)

Issue #1185, opened 2018-02-21
Fix link formatting in DOCX files


stephaniesimms:
The hyperlinks in DOCX downloads of funder templates are all wonky with extra long underlining (see screenshot "Datacite organisation"). @briri suggested swapping the current gem for the one we use in DMPTool.

![screen shot 2018-02-13 at 8 19 14 am](https://user-images.githubusercontent.com/12703638/36160728-bcc4ef58-1096-11e8-95fe-f7f477acba63.png)

vyruss:
Duplicate of #592 , closing older issue

vyruss:
@briri PR submitted upstream as karnov/htmltoword#70 . Until they merge it in, we can put this in our gemfile to fix the situation:
`gem 'htmltoword', :git => 'https://github.com/vyruss/htmltoword'`

stephaniesimms:
links still appear with white space in DOCX templates and plans

briri:
@stephaniesimms I updated the gemfile as @vyruss mentions above on roadmap-stg. Can you see if the issue is resolved? We will need to wait until the owner of the htmltoword gem accepts the fix before releasing this, but we can at least verify that its working as expected.

stephaniesimms:
now things are looking good. links are formatted correctly and are functional. should i leave this issue open until the gem owner accepts the fix?

sjDCC:
Looks good. No funny underscores around sides, but doesn't format as a URL either (i.e. blue)

@vyruss can you confirm re: the gem?

stephaniesimms:
ready to test on roadmap stage when the PR comes through @vyruss @briri

vyruss:
@stephaniesimms @briri I can confirm that this has gone through, I just tested it and it works. The current `Gemfile` does not reference my fork therefore a ```bundle update htmltoword``` will just bring in `htmltoword 1.0.0` which includes my fix.

stephaniesimms:
links look lovely now. closing issue.

PR #1559, opened 2018-05-30
Add checks to controller to redirect requests for edit to show if not the latest template

### LINKED ISSUE #1554: Should not be able to open/edit historical version of template
When I click the View link for an historical version of a template or customization I'm taken to a page where it looks like I can edit the template or phase details. If I do try to make changes and save them I get an error message that I can't create a new version (screenshot below). All fields on historical versions should be read only to prevent confusion.

screen shot 2018-05-25 at 1 17 30 pm

briri:
Fixes #1554 

Update template/phase/section controllers to redirect requests to the edit action over to the show action if they do not belong to the latest version of the template. We control the flow via what actions are displayed on the views but we were not properly checking in the controller.

Also updated htmltoword to pull in @vyruss patch as mentioned in #1185

### DIFF
--- a/app/controllers/org_admin/phases_controller.rb
+++ b/app/controllers/org_admin/phases_controller.rb
@@ -26,12 +26,9 @@ module OrgAdmin
     def edit
       phase = Phase.includes(:template).find(params[:id])
       authorize phase
-      if !phase.template.latest?
-        flash[:notice] = _('You are viewing a historical version of this template. You will not be able to make changes.')
-      end
       section = params.fetch(:section, nil)
       # User cannot edit a phase if its a customization so redirect to show
-      if phase.template.customization_of.present?
+      if phase.template.customization_of.present? || !phase.template.latest?
         redirect_to org_admin_template_phase_path(template_id: phase.template, id: phase.id, section: section)
       else
         render('container',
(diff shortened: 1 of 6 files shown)


app/controllers/org_admin/phases_controller.rb
```diff
@@ -26,12 +26,9 @@ def show
     def edit
       phase = Phase.includes(:template).find(params[:id])
       authorize phase
-      if !phase.template.latest?
-        flash[:notice] = _('You are viewing a historical version of this template. You will not be able to make changes.')
-      end
       section = params.fetch(:section, nil)
       # User cannot edit a phase if its a customization so redirect to show
-      if phase.template.customization_of.present?
+      if phase.template.customization_of.present? || !phase.template.latest?
```
jollopre: where are we handling the case of editing a 'own' phase created in a template that is customised?

app/controllers/org_admin/sections_controller.rb
```diff
@@ -36,7 +36,8 @@ def show
     def edit
       section = Section.includes({phase: :template}, questions: [:question_options, { annotations: :org }]).find(params[:id])
       authorize section
-      render partial: 'edit', 
+      # User cannot edit a section if its not modifiable or the template is not the latest redirect to show
+      render partial: (section.modifiable? && section.phase.template.latest? ? 'edit' : 'show'),
```
jollopre: we should add a test to inspect the DOM and check whether or not is displayed the edit/show partial depending on the case.

PR #1561, opened 2018-05-30
fixed issue with policy check for email notifications

### LINKED ISSUE #1514: Notification Preferences can't be updated
Hello, 

It seems that a **non super_admin** user can't update his notification preferences on his profile page. 
The following message is displayed : 
image

briri:
Fixes #1514 .

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -2,7 +2,7 @@ class UsersController < ApplicationController
   helper PaginableHelper
   helper PermsHelper
   include ConditionalUserMailer
-  after_action :verify_authorized
+  after_action :verify_authorized, except: ['update_email_preferences']
   respond_to :html
 
   ##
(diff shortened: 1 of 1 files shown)


app/controllers/users_controller.rb
```diff
@@ -91,7 +91,8 @@ def admin_update_permissions
 
   def update_email_preferences
     prefs = params[:prefs]
-    authorize current_user, :update?
+    # Only allow the user to update their own preferences unless it is a super admin
+    raise Pundit::NotAuthorizedError unless current_user.can_super_admin? || current_user.id != params[:id]
```
jollopre: This line of code does not really fix the problem since unlike the above comment states, it does the following:
raise Pundit::NotAuthorizedError:
- when the current_user is super_admin or 
- when current_user id is different. 

I don't think we want to prevent a super_admin to update her email preferences neither it is needed to rely on the id passed at all since at the end we are using current_user to update the preferences.

jollopre:
@briri, I'm closing this PR since #1563 should fix the issue #1514

Issue #1339, opened 2018-03-29
rake gettext:find not collecting everything


briri:
The `rake gettext:find` task doesn't seem to be picking up everything.

The following 'Private: ...' section is not found in the app.pot file but the 'Organisation:' and 'Private:' ones are.

```ruby
    <div class="form-group col-xs-8">
      <div class="radio">
        <%= f.label :visibility_privately_visible, raw("#{f.radio_button :visibility, :privately_visible}\
        #{_('Private: visible to me, specified collaborators and administrators at my organisation')}") %>
      </div>
      <div class="radio">
        <%= f.label :visibility_organisationally_visible, raw("#{f.radio_button :visibility, :organisationally_visible} #{_('Organisation: anyone at my organisation can view')}") %>
      </div>
      <div class="radio">
        <%= f.label :visibility_publicly_visible, raw("#{f.radio_button :visibility, :publicly_visible} #{_('Public: anyone can view')}") %>
      </div>
    </div>
    <div class="col-xs-8">
      <%= f.submit(_('Update'), style: 'display:none') %>
    </div>
```

Interestingly though, some older versions are still in the app.pot:
```yaml
msgid "Private: restricted to me and my collaborators"
msgstr ""

msgid "Private: restricted to me and people I invite."
msgstr ""
```

briri:
Thanks to @stephaniesimms for pulling together this list of strings that are being missed.

**Emails**
There is code in place that forces the system to use the default language when creating email messages. We should switch this to use the recipient's preferred language (or the default if none is specified)

**Notification messages**
Notification messages all partial and strange (some screenshots below)
![screen shot 2018-04-05 at 10 53 14 am](https://user-images.githubusercontent.com/12703638/38382757-a185bbd6-38bf-11e8-83b7-7bd5d878c7e9.png)
![screen shot 2018-04-05 at 10 59 26 am](https://user-images.githubusercontent.com/12703638/38383113-779a142e-38c0-11e8-83c3-e0459f3772ee.png)
![screen shot 2018-04-05 at 11 12 13 am](https://user-images.githubusercontent.com/12703638/38383771-398820c0-38c2-11e8-8d6c-d783e6bee437.png)

**Checkboxes and Radio buttons**
We've been using string interpolation as a practice to create Bootstrapped checkboxes and radio buttons that still make use of Rails' ActionView form helpers to create these elements. We will need to adjust this by either using a '+' to combine the strings or set the label text as a variable 
[...]

PR #1578, opened 2018-06-05
Added new translatable:find rake task and pt_BR translations

briri:
For #1339 
- Created a new rake task called `translatable:find`
  - Scans all files in 'app/', 'lib/' (except 'node_modules/'), 'config/' for translatable text (meaning anything using the gettext markup: `_('some text')` or `n_('item', 'items', variable)`. (This works with Ruby string interpolation)
  - Makes a backup of the POT/PO files (git ignores these backups) 
  - Adds any new text (translatable text that does not exist in the current POT/PO file) with the gettext `#, fuzzy` notation
  - Marks any text that appears in the current PO/POT file but is no longer found in the translatable text list as obsolete using the gettext notation of commenting the lines out: `#msgid "some text"` and `#msgstr "some text translated`
- Made a few minor modifications to files to work with the `translatable:find` script
- Ran `rake translatable:find` to update the app.pot and app.po files
- Added in the Português (Brasil) translations

Caveats with the `translatable:find` text (we can fix later if it becomes an issue:
- If you use '+' as string concatenation you need to have a space between them and the gettext markup (e.g. `"This is " + _('translatable') + "text"` instead of `"This is "+_('translatable')+"text"`
- Do not use Ruby line continuations, place your entire text on one line. If you use line continuations the results are unpredictable.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,10 @@ public/stylesheets/*
 public/system/*
 public/videos/*
 
+# Ignore PO/POT backups
+*.po.bak
+*.pot.bak
+
 # Ignore branded content
 app/views/branded/*
 app/assets
--- a/app/controllers/answers_controller.rb
+++ b/app/controllers/answers_controller.rb
@@ -10,8 +10,7 @@ class AnswersController < ApplicationController
       p = Plan.find(p_params[:plan_id])
       if !p.question_exists?(p_params[:question_id])
         render(status: :not_found, json:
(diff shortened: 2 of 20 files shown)


briri:
@xsrust @jollopre feel free to review this one. Don't merge though as I need to address the test failure which is due to some leftover git merge stuff in the `en_US/app.po` file. I'll clean that up tomorrow morning

PR #1584, opened 2018-06-06
Added new translatable:find rake task and pt_BR translations

briri:
For #1339 
- Created a new rake task called `translatable:find`
  - Scans all files in 'app/', 'lib/' (except 'node_modules/'), 'config/' for translatable text (meaning anything using the gettext markup: `_('some text')` or `n_('item', 'items', variable)`. (This works with Ruby string interpolation)
  - Makes a backup of the POT/PO files (git ignores these backups) 
  - Adds any new text (translatable text that does not exist in the current POT/PO file) with the gettext `#, fuzzy` notation
  - Marks any text that appears in the current PO/POT file but is no longer found in the translatable text list as obsolete using the gettext notation of commenting the lines out: `#msgid "some text"` and `#msgstr "some text translated`
- Also added tasks to the `translatable.rake` file to assist with adding/removing new languages from the DB and to compare an app.po file to the app.pot
- Made a few minor modifications to files to work with the `translatable:find` script
- Ran `rake translatable:find` to update the app.pot and app.po files
- Added in the Português (Brasil) translations
- Cleaned up obsolete and fuzzy entries in pt_BR, en_US, en_GB and app.pot

Caveats with the `translatable:find` text (we can fix later if it becomes an issue:
- If you use '+' as string concatenation you need to have a space between them and the gettext markup (e.g. `"This is " + _('translatable') + "text"` instead of `"This is "+_('translatable')+"text"`
- Do not use Ruby line continuations, place your entire text on one line. If you use line continuations the results are unpredictable.

Example of PO/POT after running `translatable:find` ('fuzzy' denotes a new entry, '#msgid & #msgstr' denote an obsolete entry):
```yaml
 ...

msgid "Change affiliation"
msgstr ""

#, fuzzy
msgid "Changing your organisation will result in the loss of your administrative privileges."
msgstr ""

msgid "DMP Templates"
msgstr ""

#msgid "DMPRoadmap ('the tool', 'the system') is a tool developed by the DCC and CDL as a shared resource for the research community. It is hosted at CDL by the University of California Curation Center."
#msgstr ""

...
```


xsrust:
Happy for this to be merged but i dont have time today to run pull it locally to resolve the conflicts

briri:
Fixed merge conflicts.

Issue #1326, opened 2018-03-26
Create/reinstate mandatory fields


stephaniesimms:
This needs to be addressed together with #1217 to ensure someone receives the feedback request emails. 

Make the following fields mandatory (marked with red asterisk):
- Organization details: Admin contact email, Contact link text (if link was entered), Org Config Info/Org Type
- Templates: Template title, Phase title, Section title, Question text, Question format
- Guidance (related #1212 to require one theme) - Guidance group name, Guidance text

briri:
Drop formtastic gem and ensure no forms are using it

stephaniesimms:
Org details page has a few issues:
- the Org URL and link text should NOT be mandatory fields
- is there some way to ensure a user inputs a valid contact email? currently i can enter anything and save successfully (see screenshot)

![screen shot 2018-06-13 at 10 13 52 am](https://user-images.githubusercontent.com/12703638/41367063-d3a6af12-6ef2-11e8-97d8-2120ff7ff287.png)

briri:
The URLs are a bit tricky I guess. IF the user has entered a value into either of the boxes it displays the red asterisks. The validation though is just enforcing that if you enter a URL that you also add text and vice versa. Try adding a new link (line appears without asterisks), add some text to either box (asterisks appear), delete the text from both boxes (asterisks disappear)

stephaniesimms:
Ah ok @briri - as long as the red asterisks don't appear for a first-time user navigating to this page to enter Org details (will test to confirm). I understand now that I'm seeing asterisks because these fields already have content. No need to devise fancy solutions.

The other issue with validating the contact email, if possible, is more important anyway.

stephaniesimms:
now i see what's happening with section asterisks. they disappear after I click to Save the Section title. it would be better if they don't disappear after clicking Save since the Phase and Question asterisks above and below the Section are always visible. this is a small point but better to be consistent so as not to confuse users.

Asterisks appear when creating Section
![screen shot 2018-06-18 at 1 44 33 pm](https://user-images.githubusercontent.com/12703638/41561096-d31cb1d6-72fd-11e8-9f09-8e01d98bb37e.png)

Asterisks for Section disappear after I click Save
![screen shot 2018-06-18 at 1 44 57 pm](https://user-images.githubusercontent.com/12703638/41561101-d603b980-72fd-11e8-971d-e3386dd0f9f5.png)

stephaniesimms:
[...]

Issue #1212, opened 2018-02-23
Require admins to apply exactly one theme to Guidance


stephaniesimms:
as part of bug issue #1183 we accidentally enabled admins to remove all themes from Guidance. reinstate the requirement to apply at least one theme for each piece of Guidance - it doesn't make sense to create un-themed Guidance because this will never display on templates.

vyruss:
While we're fixing this, I think we should change this to be "exactly one Theme to Guidance". After discussion with @sjDCC we came to the conclusion that it makes no sense to apply multiple Themes to one piece of Guidance.

This requires:
- Changing the data model Themes <-> Guidance relationship from HABTM to FK
- Changing the guidance tickboxes so that only one can be selected 
- Changing the controller so that it only accepts exactly one theme per guidance
    - Updating the controller with the appropriate warning messages

briri:
fixed as part of #1326. It made sense to address it at the same time as adding in the validations

stephaniesimms:
after testing I am able to create and publish guidance with no themes (screenshot 1). once i apply a theme and save it, then uncheck the box and save, my changes are not applied so it appears to be requiring at least one theme at that point. i can still apply multiple themes.

- desired behavior is to mark the field "Themes" with a red asterisk and require users to apply one theme only before allowing them to save guidance.
- also need to revise the tooltip (screenshot 2) with the following proposed text (@sjDCC can confirm): "Select one theme that is relevant to this guidance. This will display your generic organization-level guidance, or any Schools/Departments for which you create guidance groups, across all templates that have questions with the corresponding theme tags."

![screen shot 2018-06-13 at 3 55 11 pm](https://user-images.githubusercontent.com/12703638/41382819-b1038b8a-6f22-11e8-921f-97b1e8b29a7d.png)

![tooltip](https://user-images.githubusercontent.com/12703638/41382823-b52e23e6-6f22-11e8-9321-3ce39106b967.png)

stephaniesimms:
tooltip looks good. but when I create a new piece of guidance the Text area doesn't recognize that I've input text and won't allow me to save. I get the field validation message (screenshot below) the first time I click Save. I have to click Save a second time and then it allows me to proceed. Please amend so it saves after the first click.

![screen shot 2018-06-18 at 1 37 04 pm](https://user-images.githubusercontent.com/12703638/41560693-c3aa3a08-72fc-11e8-8b53-71f1ed78dad6.png)

[...]

PR #1588, opened 2018-06-07
Add missing form validations across site

briri:
For #1326 and #1212
- removed formtastic
- added aria-required to fields that were missing them
- added namespaces to edit profile, sign in and create account forms to fix duplicate id issue
- fixed JS behind links and when their validations appear
- switched themes from checkboxes to radio buttons on Guidances page #1212 
- updated ariatiseForm to display an aria-required validation error message above the radio button set

Also: 
- added rake task to add the helpdesk email (defined in branding.yml) to any orgs that do not have a contact_email. That field is now required. The rake task outputs which records are changed so you can contact admins.
- added rake task that identifies org/guidance_groups that have guidance assigned to multiple themes. We now only allow one theme per guidance. The rake task will identify the org and guidance group in its output. You will need to manually consolidate those down to a single theme.

Be sure to run `rake upgrade:v1_1_2` to fix up missing org contact_emails and identify multi-themed guidance!


app/views/layouts/application.html.erb
```diff
@@ -105,7 +107,9 @@
 
       AJAX_LOADING: _('Loading ...'),
       AJAX_UNABLE_TO_LOAD_TEMPLATE_SECTION: _('Unable to load the section\'s content at this time.'),
-      AJAX_UNABLE_TO_LOAD_TEMPLATE_SECTION_QUESTION: _('Unable to load the question\'s content at this time.')
+      AJAX_UNABLE_TO_LOAD_TEMPLATE_SECTION_QUESTION: _('Unable to load the question\'s content at this time.'),
+
+      OTHER_ORG_ID: other_org.present? ? other_org.id : 0
```
xsrust: Does this need to be a constant for the JS or could it just be included in the page for the views it's needed on?

lib/tasks/upgrade.rake
```diff
@@ -1,6 +1,12 @@
 require 'set'
 namespace :upgrade do
 
+  desc "Upgrade to v1.1.2"
+  task v1_1_2: :environment do
```
xsrust: Thanks for adding these for the release!

lib/assets/stylesheets/overrides.scss
```diff
@@ -544,3 +541,8 @@ $tooltip-background: $grey;
   border: none;
   background: none;
 }
+
```
xsrust: we should flag changes to overrides.scss in release notes since it means people will need to update them if they keep the customized files in a different repository

app/views/layouts/application.html.erb
briri: I could have, but it seemed reasonable to load the variable only once. The shared `my_org.js` is referenced from several different pages. Each of those pages would need to have that variable stored in a hidden input field. We also have situations where the my_org control is loaded multiple times on one render which could cause collisions/discrepancies if there were multiple sources for the value.

jollopre: As far as I'm aware, the typeahead orgs is this partial (e.g. https://github.com/DMPRoadmap/roadmap/blob/master/app/views/shared/_my_org.html.erb) which gets rendered in four views. Why don't we create a hidden input for when allow_other_orgs is true to be set to the other org id instead of mixing concerns between Ruby and JS? Going for this approach, will reduce the need of an extra query to determine the other org id when the page rendered does not really need it. Does it make sense?

lib/assets/javascripts/utils/links.js
```diff
@@ -1,15 +1,29 @@
...
+
 const getLinks = elem =>
   $(elem).find('.link').map((i, el) => {
     const link = $(el);
+    // toggleValidations(link);
     const linkVal = link.find('input[name="link_link"]').val();
     const textVal = link.find('input[name="link_text"]').val();
-    if (linkVal || textVal) { // Validations are enabled if non-empty value is found
-      enableValidations(link);
+    if (linkVal.length > 0 || textVal.length > 0) {
```
jollopre: there is no need to add .length to this conditional since in the case that linkVal and textVal are empty strings, they are considered falsy in JS and therefore the body of that if will not be reached.

lib/assets/javascripts/utils/links.js
```diff
@@ -98,6 +112,14 @@ $(() => {
     const max = target.closest('.links').attr('data-max-number-links');
     target.text(convertToText(max).toLowerCase());
   });
+
+  $('.links').on('blur', 'input', (e) => {
+    toggleValidations($(e.target).closest('.link'));
+  });
+
+  $('.links').each((idx, el) => {
```
jollopre: I don't think this is needed. If you look at places that use the link functionality, (e.g. https://github.com/DMPRoadmap/roadmap/blob/master/lib/assets/javascripts/views/org_admin/templates/new.js#L7) triggers the enableValidation for that isolated context when the page is load instead of this generic handler that you have added and can be problematic for future tweaks.

briri:
made recommended changes

xsrust:
Changes look good, Thanks

Issue #1454, opened 2018-05-03
Edit project details accessibility changes


briri:
Wrap the Principal Investigator and Data Contact sections in `<fieldset>` tags. Move the heading/labels to the `<legend>` portion of the fieldset.

This feedback was provided from the UCOP electronic accessibility review. It was defined as having 'Medium' priority.

stephaniesimms:
changes look good. closing issue.

PR #1590, opened 2018-06-08
Edit project details accessibility changes #1454

bhavi:
Fixes # .

Changes proposed in this PR:
-
Wrap the Principal Investigator and Data Contact Person sections on Project details page within Fieldset tag. 
Adding custom CSS in overrides.css


lib/assets/stylesheets/overrides.scss
```diff
@@ -544,3 +544,26 @@ $tooltip-background: $grey;
   border: none;
   background: none;
 }
+
+/* Fieldset styling */
+fieldset.project-details {
```
xsrust: Is there a need to override fieldset and legend just for project-details?
If not it would be nice to see this genericly in dmproadmap.scss as changes to overrides.scss require merging with custom branding overrides.

bhavi: Well the fieldsets with borders are requested only for the project details page currently. The other places where we are using fieldsets are for grouping radio buttons or checkboxes which have no border styling. That is the reason why this is applicable only to this  page. I guess I could move this code to dmproadmap.scss but it would still pally only to project details page.

briri: @bhavi did all of her Bootstrap work in the overrides.scss file. I think the stuff in the dmproadmap.scss was added by @vyruss at one point for the header and footer changes. I think reorganizing the CSS files is beyond the scope of this ticket.

xsrust:
Adding CSS strategy as a topic for Friday's Dev call

vyruss:
IIRC `dmproadmap.scss` is just leftovers and everything important is actually in `overrides.scss`

Issue #1217, opened 2018-02-23
Only alert the main helpdesk when a request for feedback on a DMP is received


sjDCC:
Make helpdesk contact email a mandatory field on the organisation details page. We should only be alerting this central address, not every admin user for the organisation. I spammed Glasgow users when testing and doing demos last week!

Note: this issue needs to be addressed with related issue #1197 and confirm that contact email field is required on Org details page.
![capture](https://user-images.githubusercontent.com/6169940/36608379-92a7ce74-18c1-11e8-977b-f7d7ed568f43.PNG)

We have an opt out email notification setting that I guess needs to be removed if we add logic to only email the central helpdesk. Individual admin users wouldn't be alerted so it doesn't make sense to let them untick this box.
![picture1](https://user-images.githubusercontent.com/6169940/36608392-97c8bc92-18c1-11e8-86bc-0530a4959c33.png)

briri:
just a note that we should still add the `:reviewer` role for all of the org admins even though we're no longer sending them an email.

stephaniesimms:
#1326 this issue needs to be addressed at the same time. we need to make the Admin contact email a mandatory field so that someone receives the Request feedback emails. noting here that we should move these two issues to the next sprint.

stephaniesimms:
this appears to be working as expected now. i tested as a couple of different orgs and emails only went to the main org contact. @sjDCC should double check to confirm but i'll go ahead and close it.

PR #1597, opened 2018-06-12
Only email admin helpdesk when plan review requested (not all admins) #1217

xsrust:
- Only Email `org.contact_email` when a plan is submitted for review
- Removed the opt-out preference to not be notified from branding_example.yml
    -- Need to add to release notes that production branding.yml will need updated with release to remove the admin: preference
- Added rake task to remove the admin preference from the db
    -- Need to add to release notes and meta task for release

Apologies for editor ripping whitespace


lib/tasks/upgrade.rake
```diff
@@ -369,4 +369,18 @@ namespace :upgrade do
     Rake::Task['upgrade:remove_duplicated_non_customised_template_versions'].execute
     Rake::Task['upgrade:remove_duplicated_customised_template_versions'].execute
   end
+
+  desc "Remove admin preferences"
+  task remove_admin_preferences: :environment do
```
briri: You should call this from the `v1_1_2` task so that users only need to run the one scripts

xsrust:
Added the rake taks to v1.1.2 upgrade task

PR #1605, opened 2018-06-14
French (France) Translation

benjaminfaure:
@stephaniesimms asked me to send you a PR of our French translation so here it is !

It seems that the `contributions `branch is not even with the `development `branch so there is a lot of commits :)


briri:
Hi @benjaminfaure. I just pulled in all of the latest changes from the development branch and then merged your `fr_FR/app.po` file in. We have a new rake task called `translatable:find` that replaces the use of `gettext:find` (which was often missing a lot of our translatable strings). I ran this and it updated your original `fr_FR/app.po` with all of the new strings.

If you pull down the latest contributions branch, you can update your `app.po` to provide any missing strings. 
- `#, fuzzy` indicates a translatable string that was found in the code but not in your app.po. Update the msgstr and delete the `#, fuzzy` comment
- `#msgid` and `#msgstr` indicate that the string was in your app.po but no longer exists in the codebase, so you can delete them or leave for reference if you'd like

benjaminfaure:
Thank you @briri 
Lot of translations to do ! 

I suppose your goal is to have clean .po files before updating the Pootle

briri:
Yes! clean po files is the goal :)
We have had some recent transitions within the members of the core project team so we're a bit delayed with updating things like Pootle. We will try to make updating it a part of the release process in the future.

PR #1608, opened 2018-06-14
switched guidance themes from checkboxes to radio buttons

briri:
For #1212 

- Switch theme checkbox display to radio buttons and added logic for to display the validation message for radio button sets above the section. 
- Discovered and patched bug with removing theme selections from a question. When no themes were selected on the form it did not pass back the 'theme_ids' attr, so they were not being updated
- Added a 'No themes selected' message to show question


lib/assets/javascripts/utils/validation.js
```diff
@@ -154,21 +154,30 @@ const getValidationMessage = (el) => {
...
+  helpBlock.show();
 };
 const addValidationMessage = (el, excludeAsterisks) => {
   const target = $(el);
   if (isString(target.attr('id'))) {
     const id = target.attr('id');
+    const typ = getValidationTypeForElement(target);
     if (!isString(target.attr('aria-describedby'))) {
-      target.after(blockHelp(`help-${id}`, getValidationMessage(el)));
+      if (typ === 'radio' || typ === 'cehckbox') {
```
jollopre: spelling typo at 'checkbox'. You have made changes in both validation logic JS which are used across every form in the site. Any change in this file should be cross-tested.

app/views/org_admin/shared/_theme_selector.html.erb
```diff
@@ -1,5 +1,10 @@
-<%# locals: all_themes & popover_message %>
+<%# locals: all_themes, asRadio & popover_message %>
+<% asRadio = asRadio ||= false %>
```
jollopre: the memoized operator is not needed here and in any case it is used incorrectly apart from using camelCase notation which is not a convention in ruby. If you really want to use it, you should do the following:
```
asRadio ||= false
```
which does set asRadio if is falsy, i.e. nil or false.

app/views/org_admin/questions/_show.html.erb
```diff
@@ -54,7 +54,7 @@
           <% themes_q = question.themes %>
           <% if !themes_q.nil? %>
               <dt><%= _('Themes')%></dt>
-              <dd><%= themes_q.join(', ') %></dd>
+              <dd><%= themes_q.length <= 0 ? _('No themes selected') : themes_q.join(', ') %></dd>
```
jollopre: No need for '<' but anyway is harmless

app/controllers/org_admin/questions_controller.rb
```diff
@@ -70,6 +70,8 @@ def update
         # modifiable (versioned) question
         attrs = question_params
         attrs = transfer_associations(question) if question.id != params[:id]
+        # If the user unchecked all of the themes set the association to an empty array
+        attrs[:theme_ids] = [] unless attrs[:theme_ids].present?
```
jollopre: nice check here. Better to add it at the server level that relying on setting it up through JS .

lib/assets/javascripts/utils/validation.js
briri: thanks for catching this!

app/views/org_admin/shared/_theme_selector.html.erb
briri: ayiyi, the problems of switching back and forth between JS and Ruby. ;) Will switch asRadio to as_radio

PR #1611, opened 2018-06-14
fixed issue with Tinymce toolbar

### LINKED ISSUE #1594: Standardize tinyMCE editors for template editing
While trying to edit a template I noticed that the tinyMCE editors for sections and questions do not have the full formatting functionality. 

Replace the editors for Section Details & Question text, Example answer, etc. in the screenshot below with the version of the editor that includes lists and tables. All editors should include Underline formatting as well.

Roadmap template editors
screen shot 2018-06-11 at 3 02 53 pm

DMPTool template editors - for mysterious reasons there are different versions here for Section details
screen shot 2018-06-11 at 1 14 06 pm

briri:
Fixes #1594 

The `views/notifications/edit.js` passes in a new 'toolbar' option that removes tables and bullets. Subsequent `tinymce.init()` calls end up using that 'toolbar' instead of the one defined in the defaultOptions constant on `utils/tinymce.js`. 

We're using the jquery `$.extend()` method to combine the options passed in by the caller with the defaultOptions. I tried switching it to `Object.assigns()` in case it was somehow updating the defaultOptions constant. It made no difference though. I think that Tinymce itself must be doing something weird.

Using an interim solution here to pass in the full 'toolbar' options that we want on the ajax based init calls. We may need to add this in elsewhere if we find any other situations like this. Another alternate option would be to stop passing the pared down 'toolbar' options in `views/notifications/edit.js` and just rely on super admins to not use bullets and tables in their notifications.

### DIFF
--- a/app/views/guidances/new_edit.html.erb
+++ b/app/views/guidances/new_edit.html.erb
@@ -13,7 +13,7 @@
     <%= form_for(guidance, url: options[:url], html: { method: options[:method] , id: 'new_edit_guidance'}) do |f| %>
       <div class="form-group" data-toggle="tooltip" title="<%= _('Enter your guidance here. You can include links where needed.') %>">
         <%= f.label :text, class: 'control-label' %>
-        <%= text_area_tag("guidance-text", guidance.text, class: "tinymce form-control", 'aria-required': true, rows: 10) %>
+        <%= text_area_tag("guidance-text", guidance.text, class: "guidance-text form-control", 'aria-required': true, rows: 10) %>
       </div>
       <%= render partial: 'org_admin/shared/theme_selector', 
                  locals: { f: f, all_themes: themes,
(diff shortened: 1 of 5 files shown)


jollopre:
@briri, have you noticed that the selector #notification_body at https://github.com/DMPRoadmap/roadmap/blob/master/lib/assets/javascripts/views/super_admin/notifications/edit.js#L6 does not exist in any partial or view?  

I suspect an invalid selector is causing this 'weird' change of defaults. Can you try to create a specific selector for the tinymces at 
https://github.com/DMPRoadmap/roadmap/blob/master/app/views/super_admin/notifications/_form.html.erb and change the selector init at lib/assets/javascripts/views/super_admin/notifications/edit.js

briri:
ah good catch. I'll give that a try. Prefer if that fixes it rather than passing the toolbar options around

briri:
nope. doesn't seem to matter. going to leave the toolbar options in for now

xsrust:
Thanks for checking into that.  Your fix works in the interim.

Issue #1589, opened 2018-06-08
"My organization isn't listed." workflow is inconsistent


Yu-Chieh-Henry-Yang:
**Expected behaviour:**
I should not be able to edit the super admin in http://localhost:3000/super_admin/users/1/edit
OR
The "My organization isn't listed." button should work

**Actual behaviour:**
In http://localhost:3000/org/admin/users/admin_index, I can see the list of users, and there is no edit link under the privileges column. Then if I click that user who don't have the edit link button, I will arrive at http://localhost:3000/super_admin/users/1/edit.

But I shouldn't be able to edit it.

If I am able to edit it, the "My organization isn't listed." button doesn't work. (while the same button at http://localhost:3000/users/edit works)

**Steps to reproduce:**
1) Setup the roadmap application from scratch according to [installation wiki](https://github.com/DMPRoadmap/roadmap/wiki/Installation)
2) Open http://localhost:3000/org/admin/users/admin_index and click the first user (should be super user)
3) see that clicking the "My organization isn't listed."button doesn't work

**Note**
I tried to debug the issue by doing a system search on the key word "My organization isn't listed." but the js are all ulgified, do you have a normal version somewhere?

xsrust:
hi @Yu-Chieh-Henry-Yang,  thanks for the bug report. I can confirm that this is also an issue on our running services.

The view partial which renders that is shared with the edit profile page, and as is the js.

All of our js lives under `lib/assets/javascripts`
for whoever picks this ticket up, partial is `app/views/shared/_my_org.html.erb` and js at `lib/assets/javascripts/views/shared/my_org.js`

briri:
this may be fixed in #1588. If not we can add it since the PR makes modifications to those files for form validation updates.

stephaniesimms:
moving this to the current sprint so i can test. the link works now but the behavior should be the same as when a user clicks the "My Org isn't listed" link to edit their own profile or creates an account.

currently, as a Super Admin I get the following validation error if I don't enter a valid Org and then i can't save the changes (screenshot 1). I should be permitted to save this entry and/or provide a default "No affiliation" entry. 

this raises the question about how we're storing/identifying these free text user affiliations in the database. my understanding is that we want to present a "No affiliation" option and use this value in the db even if a user inputs something into this field. can someone confirm that this is the current db situation? our workflow for dmptool is slightly different so i'm confused about the correct behavior for roadmap.

[...]

PR #1614, opened 2018-06-14
fixed issue with saving other org on sup_admin edit user

briri:
For #1589 

Fixes issue with saving 'other org' on the edit user page.

Also fixed issue with super admin not having access to the admin menu after updating their org to 'other org'

### DIFF
--- a/app/views/layouts/_header.html.erb
+++ b/app/views/layouts/_header.html.erb
@@ -3,7 +3,7 @@
   <%= render partial: "layouts/navigation" %>
 </div>
 <div class="org-branding">
-  <% if user_signed_in? && !current_user.org.nil? && !current_user.org.is_other %>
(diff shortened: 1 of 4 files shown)


app/views/layouts/_header.html.erb
```diff
@@ -3,7 +3,7 @@
   <%= render partial: "layouts/navigation" %>
 </div>
 <div class="org-branding">
-  <% if user_signed_in? && !current_user.org.nil? && !current_user.org.is_other %>
+  <% if user_signed_in? && ((!current_user.org.nil? && !current_user.org.is_other) || current_user.can_super_admin?) %>
```
xsrust: What is the desired behavior here?
I think if the super admin did not have an org, this would still render the page which would break.
Should this be:
` user_signed_in? && !current_user.org.nil? && (current_user.can_super_admin? || !current_user.org.is_other`

briri:
yep. Update that line of logic

xsrust:
Thanks

PR #1616, opened 2018-06-15
Issue 1409

jollopre:
Introduces a service to delegate the behaviour of selecting annotations and guidances by theme for a question from a plan. This service introduces some good patterns:
- lazy initialisation of instance variables
- memoization of hash structures representing guidances or annotation for each org
- reduction and improvement of server burden in terms of db queries
- dump partials to display the structures according to the service
- preventing polluting models and controllers with methods specific to a specific data structure

### DIFF
--- a/app/controllers/plans_controller.rb
+++ b/app/controllers/plans_controller.rb
@@ -1,6 +1,6 @@
+require 'pp'
 class PlansController < ApplicationController
   include ConditionalUserMailer
-  require 'pp'
   helper PaginableHelper
   helper SettingsTemplateHelper
   after_action :verify_authorized, except: [:overview]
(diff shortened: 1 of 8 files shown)


jollopre:
TODO:
- test for guidance service public methods
- trimming accordion titles (e.g. guidance group X on theme Y) when their size is greater than the space allocated for the guidance container

briri:
@xsrust this one wasn't a part of the current sprint. I'd like to see a few modifications and spend some more time reviewing, so lets hold off on merging this one until we start the next sprint later this week.

briri:
still think we could do some cleanup on this. Will merge and deploy for UAT though since we're having bandwidth issues. ;) We can clean up in further iterations

Issue #1333, opened 2018-03-28
Implement default sort for DMP Templates, Public DMPs, Guidance


stephaniesimms:
Implement the following default sorting behavior and retain it when users click the link to 'View All' for: 

DMP Templates (public) and Templates tables in Admin area (all tabs): alphabetized by template title
Public DMPs: most recent date created (or last edited) at top of table
Guidance list: alphabetized by first word of guidance Text (first column)

stephaniesimms:
- DMP Templates table does not retain default (alphabetical) sort when I click View All
- if possible implement alphabetical default sort for the Guidance List (second table below Guidance Group List - screenshot), by first word of Guidance text

![screen shot 2018-06-15 at 3 34 01 pm](https://user-images.githubusercontent.com/12703638/41492082-948927c2-70b1-11e8-94eb-b9a64e32074f.png)

briri:
The 'Give a summary' guidance appears first because it starts with a blank line, or technically `<p></p>`. We cannot easily select the first piece of text to sort on. 

If its a major concern/issue we should update the logic to prevent empty lines (paragraph tags) from being saved at the beginning of the entry. This would require a bit of refactoring though and a script to go through and clean up all of the existing guidance. I think though thats beyond the scope of this ticket and more along the lines of preventing bad user input.

stephaniesimms:
all of the tables retain a lovely default sort now - hooray! and thank you for the explanation re the Guidance list. I deleted the empty line and it looks fine now. You're correct that preventing bad user input is out of scope here and unlikely to become a priority. closing issue.

PR #1621, opened 2018-06-18
Fixed issues with Public Template page sorting and form validations 

briri:
For #1333 
Fixed issue with sort for public templates

For #1326:
We were using 'aria-describedby' to store the relationship between the input element and its help text block. Bootstrap however uses the 'aria-describedby' attribute when displaying an input's tooltip. This was causing the relationship between the input and its help block to be lost.

Switched from 'aria-describedby' to 'validation-help-block'. Also updated the id of the help block to match the id of the input field instead of a counter number. This keeps the help block ids unique across the entire page.


xsrust:
Changes make sense.  Thanks for the explanation for reference.

PR #1624, opened 2018-06-19
 removed default links from schema

xsrust:
Only changed the schema.rb defaults here

The migration adding this default is currently commented out so shouldnt add it going forward.


db/schema.rb
```diff
@@ -183,7 +183,7 @@
     t.string   "logo_name"
     t.string   "contact_email"
     t.integer  "org_type",               default: 0,              null: false
-    t.text     "links",                  default: "{\"org\":[]}"
+    t.text     "links"
```
briri: Should remove the default templates as well.

xsrust: good catch, thanks

PR #1627, opened 2018-06-20
fix issue with duplicate asterisks on checkboxes

briri:
Discovered this issue while merging v1.1.2 with DMPTool customizations.


xsrust:
Thanks for the bugfix

PR #1633, opened 2018-06-20
Minor translation fixes

briri:
- Added in the missing gettext markup to a string in the roles_controller
- Reran translatable:find
- Updated the gem based localization files in config/locales so that they are properly named for fr_FR and pt_BR


xsrust:
Thanks.  I'll get pootle updated with all the translations for the release today.

Issue #1443, opened 2018-05-03
Some buttons do not provide a visual cue indicating that they have focus


briri:
Some of the buttons do not clearly indicate that they have focus. This is confusing for users that navigate the page via keyboard.

The following pages were identified during the testing of the researcher facing pages, an audit of all admin pages should be done as well.
- Contact us - Submit
- My Dashboard - Create Plan
- Project Details - Submit

It would also be a good idea to go through and use the proper Bootstrap classes for our buttons (btn-primary, btn-default, etc.) We're mostly using btn-default everywhere.

This feedback was provided from the UCOP electronic accessibility review. It was defined as having 'High' priority.

stephaniesimms:
i tabbed through all the pages and this appears to be working now. closing pending thorough retest with a screenreader.

briri:
Having a link inside a button is not valid HTML5 and causes Firefox and possibly other browsers to fail. We need to find a different solution.

bhavi:
Adding tabindex: 0, role: "button" to the link_to helper seems to get the tab focus to work In Chrome. But Firefox is still not very happy about it.

This url has some interesting suggestions: [Reinventing Hyperlinks](http://www.heydonworks.com/article/reinventing-the-hyperlink)

magdalenadrafiova:
not sure whether this is till an issue? can we close @briri ?

briri:
yes. this was fixed by some of accessibility work that @johnpinto1 did for us

PR #1636, opened 2018-06-21
Reverted accessibility changes for buttons

briri:
It turn out that have a link nested within a button is invalid HTML5. The buttons are no longer functional in Firefox and possibly other browsers

Reverting the work done on PR #1600 (issue #1443)

I will reopen the ticket an make notes about this issue.


xsrust:
thanks @briri

Issue #1632, opened 2018-06-20
standardize tinyMCE editors - some lose formatting lists/tables


stephaniesimms:
testing on dmptool stage revealed persistent issues with losing some formatting functionality (lists/tables) on tinyMCE editors. in this case the Comments area on the write plan page (screenshot). 

@briri noted: this is related to DMPRoadmap/roadmap#1594.

We weren't happy with the resolution for that one. My suggestion is to update the notifications page so that its Tinymce editor allows tables and lists. We can just rely on Super Admins not to use that functionality since it will not display properly. That should become evident pretty quickly if they do use it.

![tinymce](https://user-images.githubusercontent.com/12703638/41681251-e22194c6-7488-11e8-929f-fb8afda789f2.png)

stephaniesimms:
all editors appear to be the same now. closing issue.

briri:
Reopening for verification after git cherry-pick

sjDCC:
comments box has same functions now

Issue #1562, opened 2018-05-31
Tooltips for clarity across the site


xsrust:
The purpose of this ticket is to highlight areas of the site where we can add tooltips to help guide users without their needing to consult the manual.

xsrust:
One item which was flagged in our meeting was the confusing area of Guidance/Themes.  There have been repeated queries by our users who are confused about how their different guidance will display on questions.  Tooltips using a similar language of 'tagging' on both the themed guidance and the theme portion of questions were suggested as a potential solution.

stephaniesimms:
happy to add tooltips @xsrust - but trying to keep the backlog focused on actionable issues. i'm going to close this one but please go ahead and create a new one with specific instructions for tooltips to add.

Yu-Chieh-Henry-Yang:
I'm coming from #1596, found theme confusing when I first start learning this project.
Also is there a page that introduce new people to this project about how this project work? (https://github.com/DMPRoadmap/roadmap/wiki/Data-model is very useful) Like what does theme, template do? And some tips and subtle behaviours related to them.

stephaniesimms:
@Yu-Chieh-Henry-Yang we also have a community Slack channel that is great for these discussions. i encourage you to join here: https://join.slack.com/t/dmproadmap/shared_invite/enQtMzI0NTIyODA4NjU4LWJiMDQxM2IyMDhkNmRmM2FhZWZmOGFlMThiNWI3MTJkZmNkYWRjOTc3M2M5MGI4N2Q4YjM3YzgzZTUwODg0MWY

the dmptool wiki contains a user guide for administrators, including info on themes https://github.com/CDLUC3/dmptool/wiki/Help-for-Administrators
and the dmptool app has a quick start guide for researchers https://dmptool.org/help

gjacob24:
Closing this issue because the application uses more tooltips now. Any new or specific tooltips required in the future should be described in a new issue.

PR #1641, opened 2018-07-02
Fixed issues with Tinymce toolbar options

briri:
For #1632 

Removed all occurrences of toolbar item specification from JS for views. It seems that Tinymce engine retains the last used options despite us reverting to the default options when instantiating a new editor. 

The only editor that does not use the defaults was the notifications page which removed the lists and tables. Since this page is limited to super admin use we have decided to remove that JS as an easy solution to the problem. It should become quickly apparent to an admin that their system message looks bad if they opt to use lists of tables. They have the ability to update the message afterward to correct any visual issues.

We will encounter this again at some point if we need to introduce an editor with additional or limited toolbar options.


xsrust:
That's wierd behavior from TinyMCE.
I agree that the admins should be able to quickly see any issues with display so that solution should work.  We should possibly add some tooltip text as well just to cover all bases.
#1562 is setup to collect things like this

Issue #1631, opened 2018-06-20
Create toggle on/off for email signature


stephaniesimms:
I noticed this while testing on dmptool stage. Now that we changed the functionality for the request feedback emails we need to be able to disable the signature currently applied to all emails that tells users they can change their notification preferences. In this case they can't opt not to receive this email since it goes to the main org contact no matter what.

![screen shot 2018-06-20 at 1 33 53 pm](https://user-images.githubusercontent.com/12703638/41683149-b066bcc6-748e-11e8-804c-8afee9f3827f.png)

sjDCC:
I just tested this on roadmap-staging but the email to admins of a plan request still has the configuration options on.
![capture](https://user-images.githubusercontent.com/6169940/42316104-e0d6823e-8040-11e8-9468-420f3f79f787.PNG)

The email out to users that their plan has been submitted was unusually brief, but maybe the text was intentionally adjusted as this is a staging / test area?
![text](https://user-images.githubusercontent.com/6169940/42316115-e60d71ae-8040-11e8-88c0-51f22bad2992.PNG)

stephaniesimms:
It looks like you just need to switch off the signature for the email to admins of a request for feedback  @briri . that's the only email message where we don't want the signature currently.

@sjDCC the second email to users in your screenshot above contains whatever text is entered for your Org on the Org details page. so likely just brief testy text, yes.

stephaniesimms:
functionality and email all working as expected now. closing issue.

briri:
Reopening for verification after git cherry-pick

stephaniesimms:
still good. closing again.

PR #1642, opened 2018-07-02
Make 'Change notification prefs' optional in email signatures

briri:
For #1631 

Added a flag to the email signature to hide the 'Change you notification prefs' portion of the signature for automated emails that cannot be disabled (e.g. feedback request notification to the org contact email)


xsrust:
Looks good.  I like that you've kept the partial specific behavior out of the controller

PR #1643, opened 2018-07-02
Pull in the reversion of nested links within buttons

briri:
Reverted the accessibility work that placed links inside of button tags. This is invalid HTML5 and causes some browsers to be unable to use the buttons.


briri:
Just pulling in the bug fix from master to dev

PR #1647, opened 2018-07-03
update to 1409

xsrust:
updated guidance tab not to render if no guidances are available
   - Not a big fan of the iterative approach I've taken here within the guidance service.  It would be nice to have better early stopping

Also updated the label to `Question-Level Guidance`

### DIFF
--- a/app/services/guidance_service.rb
+++ b/app/services/guidance_service.rb
@@ -20,9 +20,22 @@ class GuidanceService
     end
     return @orgs
   end
+
   def any?(org:nil, question:nil)
-    if org.nil? && question.nil?
-      return hashified_annotations? || hashified_guidance_groups?
+    if org.nil?
+      if question.nil?
+        return hashified_annotations? || hashified_guidance_groups?
+      end
+      if question.present?
+        # check each annotation/guidance group for a response to this question
+        # Would be nice not to have to crawl the entire list each time we want to know this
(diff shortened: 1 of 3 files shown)


xsrust:
Updated the PR with the suggestion from the team meeting

app/services/guidance_service.rb
```diff
@@ -20,9 +20,22 @@ def orgs
...
-    if org.nil? && question.nil?
-      return hashified_annotations? || hashified_guidance_groups?
+    if org.nil?
+      if question.nil?
+        return hashified_annotations? || hashified_guidance_groups?
+      end
+      if question.present?
+        # check each annotation/guidance group for a response to this question
+        # Would be nice not to have to crawl the entire list each time we want to know this
+        anno = orgs.reduce(false) {|found, o| found || guidance_annotations?(org: o, question: question)}
```
briri: agreed, there's likely a more efficient way to check this. Maybe a query in the other direction passing in the list of org_ids and using `joins` which uses an `inner join` so would return an empty set for us
We can tackle this later as an enhancement or if you have time during this sprint feel free to send a PR. I'm going to push this one through now so the visual stuff can be UATed.

briri:
looks good @xsrust I agree that w could optimize that one query but I think this will work for now. If you have time and want to send a PR for the optimization separately that would be fine. Going to merge this one so UAT can continue 👍

PR #1652, opened 2018-07-04
Update to 1409

xsrust:
added prefetching to question to get rid of n+1 issue on themes


briri:
looks good Sam. nice improvement

PR #1657, opened 2018-07-04
Update feeback requests to provide more info

### LINKED ISSUE #1163: Display message on Share tab instead of sending email for Request feedback
USABILITY TESTING FEEDBACK
Request expert feedback… has no idea who this would go to. He would like to know who this is going to. Wants to know what to expect when he clicks it. Clicks it… He barely notices the notification message. He has no idea who receives the message and how long it will take for him to hear back. Would like Request feedback to provide more context.

- suggestion to display the email message created by org admins below the heading for "Request feedback" on Share tab page so users have more context for what happens here. revise the request feedback tab as follows (screenshot below). 
    - revise heading above editor to "Request Expert Feedback - Message Displayed on Share Plan Tab"   
    - remove "Subject" field   
    - make Message a mandatory field (related issue #1197 )
    - Revise tooltip text: "Someone will respond to your request within 48 hours. If you have questions pertaining to this action please contact us at [insert email address]."
- display the message instead of sending the email to the user (but admins should still receive an email when a user clicks the Request Feedback button)
    - display the Message created by admins on the Share plan tab (screenshot 2 below) between the two lines of text (after "Click below to give..." and before "You can continue to edit...")

screen shot 2018-05-21 at 2 25 26 pm

screen shot 2018-05-21 at 2 29 39 pm

Bodacious:
Fixes #1163 .

Changes proposed in this PR:
- Revise heading above editor to "Request Expert Feedback - Message Displayed on Share Plan Tab"
- Remove "Subject" field from organisation feedback form
- Make Message a mandatory field (related issue #1197 ) in organisation feedback form
- Revise tooltip text: "Someone will respond to your request within 48 hours. If you have questions pertaining to this action please contact us at [insert email address]."
- Display the feedback message on the page instead of sending the email to the owner(s)
- Display the Message created by admins on the Share plan tab

### DIFF
--- a/app/helpers/orgs_helper.rb
+++ b/app/helpers/orgs_helper.rb
@@ -1,19 +1,17 @@
 module OrgsHelper
+  # frozen_string_literal: true
 
-  DEFAULT_EMAIL = '%{organisation_email}'.freeze
+  DEFAULT_EMAIL = '%{organisation_email}'
 
   # Tooltip string for Org feedback form.
   #
   # @param org [Org] The current Org we're updating feedback form for.
   # @return [String] The tooltip message
   def tooltip_for_org_feedback_form(org)
-    output = <<~TEXT
-      Someone will respond to your request within 48 hours. If you have
-      questions pertaining to this action please contact us at
-      %{email}.
-    TEXT
-    email = org.contact_email.present? ? org.contact_email : DEFAULT_EMAIL
-    _(output % { email: email })
+    email = org.contact_email.presence || DEFAULT_EMAIL
+    _("Someone will respond to your request within 48 hours. If you have \
+    questions pertaining to this action please contact us at %{email}") % {
+      email: email
+    }
   end
-
 end
--- a/app/helpers/orgs_helper.rb
+++ b/app/helpers/orgs_helper.rb
@@ -1,0 +1,19 @@
+module OrgsHelper
+
+  DEFAULT_EMAIL = '%{organisation_email}'.freeze
+
+  # Tooltip string for Org feedback form.
+  #
+  # @param org [Org] The current Org we're updating feedback form for.
+  # @return [String] The tooltip message
+  def tooltip_for_org_feedback_form(org)
+    output = <<~TEXT
+      Someone will respond to your request within 48 hours. If you have
+      questions pertaining to this action please contact us at
+      %{email}.
+    TEXT
+    email = org.contact_email.present? ? org.contact_email : DEFAULT_EMAIL
+    _(output % { email: email })
+  end
+
+end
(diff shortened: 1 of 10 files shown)


briri:
Overall this one looks good @Bodacious. You just need to fix the broken test which may have been introduced I think when I merged your other PR that added the `inviter` to some of the email messages. I would also recommend a tweak to the JS bit (see my comments) and wrapping the tooltip message so its available to the translation engine.

Bodacious:
- @briri Broken tests fixed now 👍 😄 
- I'm not sure which comments you're referring to on the JS, could you point me in the right direction?
- The tooltip text is wrapped in the `_()` method, but I moved this to the `FeedbacksHelper` for better readability. Is that likely to cause problems?

briri:
@Bodacious thanks for fixing the tests :)

I was referring to the comments in the PR here. It looks like `enableValidations` is called twice and that the second one contradicts the first (based on the CSS selectors).

Also, the tooltip is pulled in from the OrgsHelper. I like the abstraction of placing it in a helper but it needs to have the gettext markup `_('my text')` in order for it to be available for translations

Bodacious:
Sorry, you're right it's the OrgsHelper.

On line 16, I've used the `_()` method: https://github.com/DMPRoadmap/roadmap/pull/1657/files#diff-bd4eb2d84d6507a078639f775e789775R16 Is there something else I've missed?

`enableValidations` is called twice but the jQuery selectors are different (calling for two separate forms) `#edit_org_profile_form ` vs `#edit_org_feedback_form`. I can move those both into the same function call, but IMO it reads better as two separate calls.

briri:
Ah right. I missed that the form ids were different in the JS. That is fine then, no need to change it.

We need the `_()` markup around the actual text. Our new `translatable:find` rake task scrapes the site for text translations prior to compilation. The original `gettext:find` is dodgy (hence our new rake task) and would likely also have issues with it. 
Gettex has 2 scenarios, the first is the rake task that updates the app.pot/.po files which are then used to create translations; the second is the runtime portion which retrieves the text translation from the appropriate app.po file

Bodacious:
@briri Sorry, I goofed up on the `_()`. I've played around with the rake tasks and have a better idea of how it works.

The latest commit should fix that issue in the OrgsHelper?

app/helpers/orgs_helper.rb
```diff
@@ -1,19 +1,17 @@
 module OrgsHelper
+  # frozen_string_literal: true
```
Bodacious: @briri @xsrust just want to make sure you're both familiar with this macro. [It's quite useful to use in most cases](https://freelancing-gods.com/2017/07/27/friendly-frozen-string-literals.html) :)

xsrust: Cool, I was not familiar with that.

xsrust:
Gettext markup looks good to me now

app/helpers/orgs_helper.rb
briri: nor was I

PR #1662, opened 2018-07-05
Refactor Plan#readable_by? to check system settings

### LINKED ISSUE #1661: Configure Plan readability through branding.yml
Currently, plans are readable by a super_admin, an admin of any of the organizations to which co-owners belong, or to any user with a role in the plan.  

For some services, it is not wanted for org-admins to be able to view the full text of any plan within their organization, and this should be restricted to super-admins, or org-admins with requested reviews.
Note that a user requesting a review adds a role for the associated org-admin users on the plan.

We should add a configuration option for whether or not org-admins can view plans from users of their org. This will affect whether the plans at /org_admin/plans have active links to the full DMP content or not. Where all plans can't be viewed by org admins, they will only have access to plans where a specific review has been requested (see notification panel above plans table)

plans-view

Bodacious:
Fixes #1661 .

Changes proposed in this PR:
- Add helper class for accessing system configs from `branding.yml`
- Refactor `Plan#readable_by?` to check system settings

### DIFF
--- a/app/models/plan.rb
+++ b/app/models/plan.rb
@@ -346,14 +346,23 @@ class Plan < ActiveRecord::Base
   # @param user_id [Integer] the id for a user
   # @return [Boolean] true if the user can read the plan
   def readable_by?(user_id)
-    user = user_id.is_a?(User) ? user_id : User.find(user_id)
-    owner_orgs = self.owner_and_coowners.collect(&:org)
+    user           = user_id.is_a?(User) ? user_id : User.find(user_id)
+    owner_orgs     = owner_and_coowners.collect(&:org)
+    sys_permission = Branding.fetch(:service_configuration, :plans,
+                                    :org_admins_read_all)
 
-    # Super Admins can view plans read-only, Org Admins can view their Org's plans
-    # otherwise the user must have the commenter role
-    (user.can_super_admin? ||
-     user.can_org_admin? && owner_orgs.include?(user.org) ||
-     has_role(user.id, :commenter))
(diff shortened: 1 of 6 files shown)


app/models/plan.rb
```diff
@@ -346,14 +346,23 @@ def editable_by?(user_id)
...
+    user           = user_id.is_a?(User) ? user_id : User.find(user_id)
+    owner_orgs     = owner_and_coowners.collect(&:org)
+    sys_permission = Branding.fetch(:service_configuration, :plans,
+                                    :org_admins_read_all)
+
+    # Super Admins can view plans read-only
+    if user.can_super_admin?
+      return true
+    # Org Admins can view their Org's plans if system permission allows
+    elsif user.can_org_admin? && owner_orgs.include?(user.org) && sys_permission
```
briri: Yes. This is more readable. :)

lib/branding.rb
```diff
@@ -0,0 +1,19 @@
...
+  # Loads branding config from YAML file.
+  #
+  # @param keys [Array<Object>] A list of the keys to return configs for.
+  #
+  # @example Return a value
+  #   Branding.fetch(:settings, :should_work) # => true
+  #   Branding.fetch(:settings, :email) # => 'user@example.com'
+  #   Branding.fetch(:settings, :missing) # => nil
+  # @return [Object] The value of the config
+  def fetch(*keys)
```
briri: Thank you this is so much nicer than the old `Rails.configuration.branding['blah']['blah']['blah']` :)

PR #1685, opened 2018-07-06
Issue 1684

xsrust:
updated logic around theme deletion
Action is used by annotations and by questions form

Ensure question.number is present before clearing out themes

### DIFF
--- a/app/services/guidance_service.rb
+++ b/app/services/guidance_service.rb
@@ -23,9 +23,6 @@ class GuidanceService
 
   def any?(org:nil, question:nil)
     if org.nil?
-      if question.nil?
-        return hashified_annotations? || hashified_guidance_groups?
-      end
       if question.present?
         # check each annotation/guidance group for a response to this question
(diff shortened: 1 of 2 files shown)


app/services/guidance_service.rb
```diff
@@ -20,9 +20,22 @@ def orgs
     end
     return @orgs
   end
+
   def any?(org:nil, question:nil)
-    if org.nil? && question.nil?
-      return hashified_annotations? || hashified_guidance_groups?
+    if org.nil?
+      if question.nil?
+        return hashified_annotations? || hashified_guidance_groups?
```
briri: This could have just been the `else` to the `if question.present?` below?

xsrust: yes it could.  I'll push an updated version before I leave work today.  Good catch

xsrust:
Updated the PR to go to development(and learned you can edit this after creation).  Still need to push that update to the control logic you pointed out.

Issue #1680, opened 2018-07-06
Unable to create a new user account 


Bodacious:
**Expected behaviour:**
When I fill in the form and click the "Create account" button, it should create an account, sign me in, and redirect me to the root path

**Actual behaviour:**
When I click the button, nothing happens. Nothing changes on the page, and no requests are sent to the server.

**Steps to reproduce:**
- Visit the sign up page http://localhost:3000/#create-account-form and fill in the form with valid data
- Click the "Create account" button

stephaniesimms:
i can successfully create a new account. working as expected now. closing.

briri:
Reopening for verification after git cherry-pick

stephaniesimms:
this is working as expected. closing again.

PR #1686, opened 2018-07-06
Fixed issues with create account form

briri:
For #1680 
- removed old irrelevant JS associated with the create account form
- updated the create account form's namespace designation so that the param names are correct when they reach the controller
- fixed an issue with the display of the 'accept terms' checkbox validation error message


xsrust:
Looks good,
Thanks @briri

PR #1692, opened 2018-07-10
Refactor queries to avoid interpolating strings

Bodacious:
Changes proposed in this PR:
- Refactored a few queries to avoid string interpolation

### DIFF
--- a/app/models/role.rb
+++ b/app/models/role.rb
@@ -4,7 +4,7 @@ class Role < ActiveRecord::Base
 
   ##
   # Associationsrequire "role"
-  
+
   belongs_to :user
   belongs_to :plan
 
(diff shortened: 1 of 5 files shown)


app/models/role.rb
```diff
@@ -21,6 +21,21 @@ class Role < ActiveRecord::Base
   validates :user, :plan, :access, presence: {message: _("can't be blank")}
   validates :access, numericality: {greater_than: 0, message: _("can't be less than zero")}
 
+  ##
+  # Roles with given FlagShihTzu access flags
+  #
+  # @param flags [Array] One or more symbols that represent access flags
+  #
+  # @return [ActiveRecord::Relation]
+  scope :with_access_flags, -> (*flags) {
```
briri: Much better!

PR #1705, opened 2018-07-11
Fix Rails load issue re: TemplateScope

### LINKED ISSUE #1658: uninitialized constant Template::TemplateScope
**Expected behaviour:**
When starting up the rails server with `rails s -d`, and access it via browser, it should work

**Actual behaviour:**
It show me this page instead
screen shot 2018-07-05 at 10 05 43 am

**Steps to reproduce:**
install the app normally and use rails s -d to start the application.

Bodacious:
Fixes #1658 .

Changes proposed in this PR:
- Fix bug caused by non-railsy module naming.


briri:
agreed this shouldn't be in a separate `/scope` directory. Do we even need it in a separate file? I suppose one could argue that it makes things a bit more readable. That code though isn't reusable/relevant beyond the Template model which makes it an unlikely candidate for a concern. We also don't follow that pattern with our other models and its never been intuitive to me that I need to go to a separate file to see the model's scopes. 

What are your thoughts @Bodacious? What have you seen done in other projects?  

Also, a line was added to the [config/application.rb](https://github.com/DMPRoadmap/roadmap/blob/c62704337e7dc66fb8a4b519327e75d04ec9871b/config/application.rb#L43) to load this file (and a service as well). We should clean those up if they're no longer necessary or update them @Bodacious if you know of a better Railsy way to properly load them

Bodacious:
@briri You're right! (I didn't realise this module was only loaded into one model).

I think it's best to remove this module (it's technically not a _concern_), and move it back to the `Template` model.

xsrust:
I agree with the above that the scopes belong in the template model.
I fixed the merge conflicts so as long as @briri is happy with the changes, this can merge

briri:
@Bodacious it looks like the Template model has a lingering `attr_accessible` statement (according to Travis)

Bodacious:
@briri argh this must be from merging `development` in for merge conflicts 😣

PR #1706, opened 2018-07-11
Fix new bug in Paginatable module sort columns

Bodacious:
Fixes bug introduced in #1692 .

Changes proposed in this PR:
- Fixes new bug introduced by me 😞 in 1692. 
- Removes `included` block from that module (not required, and not used correctly)


briri:
looks alright to me. Thanks @Bodacious!

PR #1711, opened 2018-07-11
Fix for funder being unable to create a new template

### LINKED ISSUE #1710: Funders are unable to create new templates
Funders cannot create a new template.

Login as super admin and go to the Admin -> Templates page. Change your Org affiliation to a funder org. Click the 'Create template' button. You will receive the following error:

screen shot 2018-07-11 at 12 21 41 pm

briri:
Fixes #1710

Added nil check to the Template model's links getter. This way the hash will always be initialized correctly.

Might be a more elegant way by moving this into the model's getters/setters but this works for now

### DIFF
--- a/app/views/org_admin/templates/_form.html.erb
+++ b/app/views/org_admin/templates/_form.html.erb
@@ -56,7 +56,7 @@
     locals: { 
       context: 'funder',
       title: _('Funder Links'),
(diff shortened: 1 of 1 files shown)


briri:
Ah nice. I like that approach better. Will update it

briri:
made suggested change.

PR #1715, opened 2018-07-12
Add ability to sort phase sections by dragging

### LINKED ISSUE #1371: Organisation customised templates - section display ordering incorrect
**Expected behaviour:**

**Actual behaviour:**

data-summary-order-of-display

data-summary-order-of-display

user-plan-section-order

**Steps to reproduce:**

### LINKED ISSUE #1681: Add validations to Sections Model
On the sections model, we need to validate that for a given phase, it's sections numbers are unique.
Essentially, the combination of section.phase_id and section.number should be unique.

This could be added to the database as a migration adding an index on these fields.
Validation should also be added on the rails side.

Finally, a rake task should be added to ensure that the data matches the validation.
This will need to be run before the index can be added.

As an example, the DMPonline dataset has 528 cases of matching section numbers within the same phase.

Some discussion/further thought may be needed on how the rake task decides the new ordering

Related to #1574 (requirements capturing)
Related to #1371

Bodacious:
Fixes #1371 
Fixes #1681 

Changes proposed in this PR:
- Add ability to drag and drop phase sections. Order is updated server-side via AJAX
- Add validations to Section model to enforce unique numbers
- Add rake task to update existing db records with unique number


Bodacious:
@xsrust @briri  I've run the rake task against the DB I have from last week. It didn't raise any errors, and there aren't any invalid Sections on the db after running it.

Please have a look at the task and run it yourself just to be sure though.

xsrust:
@Bodacious 
Travis is complaining about:
```
/home/travis/build/DMPRoadmap/roadmap/lib/assets/javascripts/views/org_admin/phases/show.js
  12:76  error  Missing semicolon 
```

app/models/phase.rb
```diff
@@ -11,43 +11,62 @@ class Phase < ActiveRecord::Base
   ##
   # Associations
   belongs_to :template
-  has_many :sections, -> { order(:number => :asc) }, dependent: :destroy
+
+  has_one :prefix_section, -> (phase) {
```
xsrust: How do the prefix and suffix sections work when there are no sections that arent modifiable?

app/models/section.rb
```diff
@@ -3,20 +3,60 @@ class Section < ActiveRecord::Base
...
+
+  def self.update_numbers!(*ids, phase:)
+    # Ensure only section ids belonging to this Phase are included.
+    ids    = ids.map(&:to_i) & phase.section_ids
+    return if ids.empty?
+    # Build an Array with each ID and its relative position in the Array
+    values = ids.each_with_index.map { |id, i| "(#{id}, #{i + 1})" }.join(", ")
+    # Run a single UPDATE query for all records.
+    query  = <<~SQL
+      UPDATE #{table_name} \
```
xsrust: does table_name get infered from the model here?

lib/tasks/upgrade.rake
```diff
@@ -1,6 +1,14 @@
 require 'set'
 namespace :upgrade do
 
+  desc "Ensure all section numbers are unique per Phase."
+  task clean_section_numbers: :environment do
+    Phase.all.each do |phase|
+      ids_in_order = phase.sections.order("number, created_at").pluck(:id)
```
xsrust: created_at seems like as sensible a way as any to determine the order if multiple have the same number.

app/models/phase.rb
Bodacious: `prefix_section` will return nil, since it's a *has one* association.
`suffix_sections` will return an *ActiveRecord::Relation* object in any case

app/models/section.rb
Bodacious: Yes. Table name is defined as part of ActiveRecord. This is preferred, since you might change the table name in future.

Bodacious:
Rebased on current *development* branch to fix merge conflicts.

app/models/section.rb
briri: I like the single query update. Much more efficient.

app/views/org_admin/sections/_index.html.erb
```diff
@@ -1,57 +1,42 @@
...
-                      locals: { 
-                        template: template, 
-                        phase: phase, 
-                        section: current_section } 
-                %>
-              <% end %>
-             </div>
-          </div>
-        </div>
+    <div class="draggable-sections" data-url="<%=
```
briri: This is significant change to the DOM structure (e.g. `#sections_accordion` to `.draggable-sections`). We'll need to do some thorough UAT to make sure none of the underlying JS has been impacted.

I'm going to hold off merging this one into the development branch for now. I will include it in the deploy to roadmap-stg so that @stephaniesimms can do UAT. If UAT passes muster I'll merge it otherwise will leave it here so it can become part of the next sprint.

app/views/org_admin/sections/_section_group.html.erb
```diff
@@ -0,0 +1,14 @@
+<!-- TODO: Move this away from using ID attr -->
+<div class="panel-group section-group"
```
briri: We're now going to have 3 `#sections_accordion` will need to see how JS behaves. We should probably switch to a class selector instead

config/routes.rb
```diff
@@ -269,25 +269,26 @@
...
+
         # Used for the organisational and customizable views of index
         collection do
           get 'organisational'
           get 'customisable'
         end
-        
-        resources :phases, only: [:show, :edit, :new, :create, :edit, :update, :destroy] do
+
+        resources :phases, except: [:index] do
```
briri: 👍

app/models/section.rb
briri: Just tried running this on roadmap-stg and it looks like MySQL doesn't like the structure of the statement:
```bash
ActiveRecord::StatementInvalid: Mysql2::Error: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'FROM (VALUES (3599, 1), (3600, 2), (3601, 3), (3602, 4), (3603, 5), (3604, 6), (' at line 1: UPDATE sections         SET number = svals.number         FROM (VALUES (3599, 1), (3600, 2), (3601, 3), (3602, 4), (3603, 5), (3604, 6), (3605, 7)) AS svals(id, number)         WHERE svals.id = sections.id;
```

Bodacious:
@briri I've updated those changes to provide support for MySQL and SQLite

briri:
thanks @Bodacious

Issue #1664, opened 2018-07-05
Update test suite to use RSpec


Bodacious:
It has been suggested that RSpec is the preferred testing framework moving forwards.

RSpec is a more feature-rich testing framework than MiniTest, and can offer better readability for developers (e.g. examples can be nested within contexts).

#### Pros

- Encourages atomic testing—fewer bugs
- Most readable testing syntax, will help new contributors with adoption
- Well supported (4th most popular Ruby gem in 2018)
- Good performance (but not best)
- Can be configured locally to suite individual developer preferences without affecting repo

#### Cons

- Requires developer to learn DSL
- Slower than MiniTest

Issue #1651, opened 2018-07-04
Replace fixtures with Factories for testing.


Bodacious:
Factories (e.g. [FactoryBot](https://github.com/thoughtbot/factory_bot)) are a lot more flexbile than fixtures, and should be used for most testing scenarios. By implementing factories within the current codebase we would be able to remove quite a lot of code from the existing tests and simplify them ([example](https://github.com/DMPRoadmap/roadmap/blob/master/test/unit/user_test.rb#L8-L26) )

#### Pros

- Makes it faster to write new tests
- Makes tests easier to read
- Makes it easier to maintain test suite over time as models change
- Ensures test objects are valid records
- Well maintained

#### Cons

- Slightly slower than YAML Fixtures
- Requires developers to learn a new DSL

PR #1721, opened 2018-07-12
Add RSpec and support gems for testing

Bodacious:
Related to #1664 
Related to #1651

Changes proposed in this PR:
- Add the RSpec and RSpec-rails gems
- Update the application generator configs to prefer RSpec and factories
- Add FactoryBot gem for better testing objects
- Add DatabaseCleaner for a clean DB on each test run


config/environments/test.rb
```diff
@@ -1,6 +1,10 @@
 Rails.application.configure do
   # Settings specified here will take precedence over those in config/application.rb.
 
+  # This is required for the Faker gem. See this issue here:
+  # https://github.com/stympy/faker/issues/266
+  config.i18n.available_locales = %w[en en_GB]
```
briri: most tests won't require anything other than the en_GB default. If we do move towards allowing some of DB data becoming translatable though and add UI tests we may want others.
Would a `Dir["config/locale/"].each{ ... }` approach make sense to pull in whatever languages are currently available in the codebase? If not we could just hard-code 'en_US', 'es', etc. as needed I guess

spec/factories/annotations.rb
```diff
@@ -0,0 +1,20 @@
...
+#  type        :integer          default(0), not null
+#  created_at  :datetime
+#  updated_at  :datetime
+#
+
+FactoryBot.define do
+  factory :annotation do
+    question
+    org
+    text { Faker::Lorem.paragraph }
```
briri: I like this already :)

spec/factories/exported_plans.rb
```diff
@@ -0,0 +1,21 @@
+# == Schema Information
```
briri: This is perhaps for a larger discussion topic, but I've never understood the purpose of this model. It get created when a user exports their plan to PDF,DOCX, etc. but is never used anywhere else. Not sure why we're recording it if we don't even use the info for stats

spec/factories/annotations.rb
Bodacious: :)

spec/factories/exported_plans.rb
Bodacious: I think @xsrust feels the same way.

config/environments/test.rb
Bodacious: We could do. It's really just to get the tests passing—since we won't actually run tests in different locales. So it should be fine to just have one or two values there.

briri:
Already cherry-picked bug fixes to a separate branch so merging this into development so we can keep moving forward with the tests

PR #1731, opened 2018-07-13
Fix bug when feedback message is nil

Bodacious:
Changes proposed in this PR:
- Fixes bug spotted while working on #1722 that breaks this screen if feedback message is nil


briri:
Thanks

Issue #1300, opened 2018-03-15
Bottom bar of page overlaps long lists


dsisu:
Please complete the following fields as applicable:

**Expected behaviour:**

When selecting primary research organisation for a project (for example), bottom bar of the page should not overalp the list

**Actual behaviour:**

Bottom bar overlaps list, making it difficult for user to seclect desired organisation, see screenshot

![image](https://user-images.githubusercontent.com/2138956/37479647-0db33d7a-2875-11e8-94b5-ee42c3dbe39b.png)

**Steps to reproduce:**

Create new plan -> Select primary research organisation.

dsisu:
I've seen it somewhere else but can't remember now and another user reported it too.

vyruss:
confirmed - we either have to:
1. Redraw the footer at the bottom of the page when the page becomes longer to accommodate the dropdown
2. Make the dropdown smaller and scrollable
3. Both 1. and 2.

briri:
We can probably get this to work by setting the `z-index` css attribute for the selection list

sjDCC:
Looks good thanks @johnpinto1 Closing out

stephaniesimms:
the behavior on the create plan page is weird. the dropdown menu pushes everything down below it on the page rather than covering over items that should be fixed (screenshot below - Q2 dropdown pushes Q3). the desired behavior is to push the footer down and not create overlap there, but keep other elements on the page fixed in place.

![screen shot 2018-07-13 at 10 09 53 am](https://user-images.githubusercontent.com/12703638/42704592-50e102bc-8685-11e8-9ead-4c905ce3189e.png)

johnpinto1:
This is now fixed.

stephaniesimms:
this appears to have been deployed to roadmap stage. i tested and things look good. @johnpinto1 - we can walk you through the Current Sprint project board workflow. Sarah and I check anything that needs to be tested in the UI (issues in the UAT column) and then close out the issue when everything is working as expected.

sjDCC:
Yes, agree this looks good. Thanks @johnpinto1 

closing out

PR #1733, opened 2018-07-16
The previous fix for the  long suggestion lists for the combobox woul…

johnpinto1:
Issue #1300

The previous fix for the  long suggestion lists for the combobox would be overlapped by the footer had the following issue:
- long lists pushed everything down below it on the page rather than covering over items.

Changes proposed in this PR:

  This fix makes the following changes:
      - the css selector ".combobox-suggestions" has position reverted back to "position:absolute"
        (as the change "position:static" caused the issue)
        and z-index increased to 1100 (as z-index for footer is 1030) to appear above footer;
      - replaced css selector "#footer-navbar" with "footer" for sticky footer to work correctly with above changes.

Fixes # 1300.

### DIFF
--- a/lib/assets/stylesheets/overrides.scss
+++ b/lib/assets/stylesheets/overrides.scss
@@ -318,12 +318,12 @@ body{
 }
 
 /* footer */
-#footer-navbar {
+footer {
+  position: absolute;
   bottom: 0;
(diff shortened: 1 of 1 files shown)


briri:
Looks good to me

PR #1737, opened 2018-07-16
Fixed issues with Template

briri:
Fixed a few issues with viewing and saving the template stack (template, phase, section, question).

Commented out the draggable icon for now until we discuss implementation in the team meeting tomorrow. You cannot currently reorder sections for a template that you own. Its working ok when the user is customizing a template. The UX is a bit weird. The section shakes when you drag it to an invalid position but it will not likely be clear to a user why the position is invalid (e.g. only one prefix section)

### DIFF
--- a/app/models/phase.rb
+++ b/app/models/phase.rb
@@ -50,8 +50,8 @@ class Phase < ActiveRecord::Base
   has_many :suffix_sections, -> (phase) {
     modifiable.where(<<~SQL, phase_id: phase.id, modifiable: false)
       sections.number > (SELECT MAX(number) FROM sections
-                           WHERE sections.modifiable = :modifiable)
-                           AND sections.phase_id = :phase_id
+                           WHERE sections.modifiable = :modifiable
+                           AND sections.phase_id = :phase_id)
     SQL
   }, class_name: "Section"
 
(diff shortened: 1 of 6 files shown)


app/models/phase.rb
```diff
@@ -50,8 +50,8 @@ class Phase < ActiveRecord::Base
   has_many :suffix_sections, -> (phase) {
     modifiable.where(<<~SQL, phase_id: phase.id, modifiable: false)
       sections.number > (SELECT MAX(number) FROM sections
-                           WHERE sections.modifiable = :modifiable)
-                           AND sections.phase_id = :phase_id
+                           WHERE sections.modifiable = :modifiable
+                           AND sections.phase_id = :phase_id)
```
Bodacious: @briri Thanks! I just caught this too updating my feature branch

Bodacious:
Hi @briri thanks for this.

I had reached a similar solution here: https://github.com/DMPRoadmap/roadmap/blob/issue-1371/app/views/org_admin/sections/_index.html.erb#L9-L40, but if you think that `phase.modifiable?` is a suitable check then I much prefer that solution.

briri:
Yeah, the two main scenarios for modifiable have to do with whether or not the template is an original or a customization of an original. The `customize!` method is called on the Template model which performs a `deep_copy` of the template but sets it and all of its nested objects to `modifiable=false`. It generates a new family_id for the customization and sets the customization_of field equal to the original template id.

briri:
@Bodacious if this looks ok go ahead and merge (assuming you're able to)

PR #1744, opened 2018-07-18
Add Brakeman to Travis CI script

Bodacious:
Adds Brakeman to Travis CI config to scan for known security vulnerabilities.

Some points to note:
- Moved Ruby version in Travis to 2.4.1. This is already installed on Travis, so should be faster than installing a 2.4.4 build. (In theory, 2.4.4 shouldn't have any breaking changes since 2.4.1).
- This PR includes the updates I made to the DB schema to ensure cross-compatability for MySQL and PostgreSQL
- Rails comes with a bin file in `bin/setup` which should be used to set up the application when a dev is getting started. I've updated this to handle the cloning of example files.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,6 +1,6 @@
 source 'https://rubygems.org'
 
-ruby '>= 2.4.4'
+ruby '>= 2.4.0'
 
 # ------------------------------------------------
 # RAILS
@@ -19,10 +19,13 @@ gem 'responders', '~> 2.0'
 
 # ------------------------------------------------
 #    DATABASE/SERVER
-# A simple, fast Mysql library for Ruby, binding to libmysql (http://github.com/brianmario/mysql2)
+
+# A simple, fast Mysql library for Ruby, binding to libmysql
+# (http://github.com/brianmario/mysql2)
 gem 'mysql2', '~> 0.4.10'
 
-# Pg is the Ruby interface to the {PostgreSQL RDBMS}[http://www.postgresql.org/] (https://bitbucket.org/ged/ruby-pg)
+# Pg is the Ruby interface to the {PostgreSQL
+# RDBMS}[http://www.postgresql.org/](https://bitbucket.org/ged/ruby-pg)
 gem 'pg', '~> 0.19.0'
 
 # Bit fields for ActiveRecord (https://github.com/pboling/flag_shih_tzu)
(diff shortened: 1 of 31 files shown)


Bodacious:
Please note that I've edited the description and pushed out new code since this PR was approved.

Gemfile
```diff
@@ -1,6 +1,6 @@
 source 'https://rubygems.org'
 
-ruby '>= 2.4.4'
+ruby '>= 2.4.0'
```
briri: why the downgrade here?

db/schema.rb
```diff
@@ -389,21 +390,21 @@
...
   end
 
-  add_index "user_identifiers", ["user_id"], name: "index_user_identifiers_on_user_id"
+  add_index "user_identifiers", ["user_id"], name: "index_user_identifiers_on_user_id", using: :btree
 
   create_table "users", force: :cascade do |t|
     t.string   "firstname"
     t.string   "surname"
-    t.string   "email",                  default: "",   null: false
+    t.string   "email",                  limit: 80, default: "",   null: false
```
briri: not sure where this size limit is coming from. do we want it?

Bodacious: It's generally good practice to constrain all string columns where you can. I thought this one was low hanging fruit. I don't think we'll ever see an email address longer than 80 characters?

Gemfile
Bodacious: In `.travis.yml` I've set travis to use Ruby 2.4.1 because this is pre-installed on Travis, and should be faster.

The downgrade here will accommodate that, but still allow us to run 2.4.4

db/schema.rb
Bodacious: @briri This shouldn't be merged in with these changes. I've updated this branch to revert the change.

Bodacious: Correction: Since the email column is indexed on the users table, this raises an error on MySQL dbs when not constrained.

```
ActiveRecord::StatementInvalid: Mysql2::Error: Specified key was too long; max key length is 767 bytes: CREATE UNIQUE INDEX `index_users_on_email` USING btree ON `users` (`email`) 
```

Reverting the previous commit to keep this in place

briri: ok. I have no objections to setting character limits in general. No point in allocating unnecessary space if we will never use it.

We had some issues with our templates.links and orgs.links columns in the past where it was changing the between TEXT and VARCHAR and then various size limits simply based on whataver machine had last run a migration. Wanted to make sure it wasn't something similar :)

PR #1748, opened 2018-07-18
Allow multiple themes per guidance

### LINKED ISSUE #1742: Reinstate ability to add multiple themes to Guidance
We changed the way admins apply themes to Guidance mainly to prevent having no themes - #1212 . But the restriction on applying only one theme per individual piece of Guidance turns out to be undesirable for users. @briri confirmed that it's not technically difficult to reinstate the ability to apply multiple themes. 

Many DMPtool users have requested this but if it poses problems for DCC services @sjDCC we can make this configurable. See screenshot below for an example of why users want to be able to apply multiple themes still. Previously UCSF could have one piece of Guidance with the themes "Data repository, Preservation, Data sharing" that ensured their suggested language about the IR would appear on any template with a question related to data preservation. Now they have to create 3 pieces of guidance to achieve the same.

screen shot 2018-07-17 at 11 53 32 am

briri:
Fixes #1742 
- Update flag to switch control from radio button to check boxes (DB already allows for 1-m relationship)
- Discovered that the 'accept terms' checkbox on create account form was not displaying an error message while testing. Fixed HTML structure 
- Added generic validation message for scenarios when user must select at least one of multiple checkboxes
- Updated ariatiseForm.js to distinguish between single checkboxes and mutiple checkboxes.

### DIFF
--- a/app/views/guidances/new_edit.html.erb
+++ b/app/views/guidances/new_edit.html.erb
@@ -14,8 +14,8 @@
         <%= text_area_tag("guidance-text", guidance.text, class: "form-control", 'aria-required': true, rows: 10) %>
       </div>
       <%= render partial: 'org_admin/shared/theme_selector',
-                 locals: { f: f, all_themes: themes, as_radio: true, required: true,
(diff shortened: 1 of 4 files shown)


briri:
closing. Included this in the #1741 PR that cherry picked commits from development

PR #1756, opened 2018-07-25
Add validations for all models

### LINKED ISSUE #1574: Data Validations Capturing Requirements
There are many assumptions our code makes about our data.  Currently, very few of these are enforced, either through the DB or activerecord validations.  

Explicitly enforcing validation not only ensures we have the expected data, but also makes those assumptions explicit for the developers working on the code, and the contributing community.

This ticket will collect validation requirements which will be split off into other tickets of the same milestone.

Validations should not be added without an accompanying rake task to ensure that the data on our and contributors services match the validation.

### LINKED ISSUE #1667: Use shoulda to test validations and associations in models
[Shoulda](https://github.com/thoughtbot/shoulda) provides macros that make it much simpler and faster to create tests for common rails behavaiour (associations, validations, etc.). This gem would be really helpful in the quest for most rhobust unit tests.

#### Example:

``` ruby
describe 'User' do
  
  # Test that user has a first name presence validation
  it { is_expected.to validate_presence_of(:first_name) }
  
end
```

#### Pros

- Makes it very easy to write new tests for common behaviour
- Very readable
- Well maintained

#### Cons

- Requires developer to learn a new API (simple)

Bodacious:
Fixes #1667
Closes #1574 

Changes proposed in this PR:
- Add validations to each of the model files
- Add specs for model validations (with Shoulda)
- Add Travis CI check for RSpec
- Add framework for catching invalid database records
- Add rules for fixing invalid database records
- Fix newly broken seeds

## Discussion

When running the rake tasks to find bad data on my local machine with a copy of our data, I still see the following issues (the rest are fixed by the rules):

### I think these can be ignored, since this model is to be deprecated

- ExportedPlan was invalid: Format pdf (with admin) is not a valid format
- ExportedPlan was invalid: Format pdf (without admin) is not a valid format
- ExportedPlan was invalid: Plan can't be blank
- ExportedPlan was invalid: Format json is not a valid format
- ExportedPlan was invalid: Format xml is not a valid format

### I've ignored these for now because I'm not 100% sure we want these validations? Can @briri or @xsrust  have a look at the validations in `app/models/identifier_scheme.rb` please?

- IdentifierScheme was invalid: Logo url can't be blank
- IdentifierScheme was invalid: User landing url can't be blank

### This was only one record on our DB. I think we can fix it manually.
- Plan was invalid: Title can't be blank

Please see the [README for the data cleaning rules](https://github.com/DMPRoadmap/roadmap/blob/6b23413259b3024a30b8b697aeb3f8f6b91d5cc7/lib/data_cleanup/README.md) for more info.

### DIFF
--- a/app/models/annotation.rb
+++ b/app/models/annotation.rb
@@ -21,9 +21,14 @@
 #
 
 class Annotation < ActiveRecord::Base
+  include ValidationMessages
+
   enum type: [ :example_answer, :guidance]
-  ##
-  # Associations
+
+  # ================
+  # = Associations =
+  # ================
+
   belongs_to :org
   belongs_to :question
   has_one :section, through: :question
@@ -34,16 +39,23 @@ class Annotation < ActiveRecord::Base
   # I liked type as the name for the enum so overriding inheritance column
   self.inheritance_column = nil
 
-  validates :question, :org,  presence: {message: _("can't be blank")}
+  # ===============
+  # = Validations =
+  # ===============
 
-  ##
-  # returns the text from the annotation
-  #
-  # @return [String] the text from the annotation
-  def to_s
-    "#{text}"
-  end
+  validates :text, presence: { message: PRESENCE_MESSAGE }
 
+  validates :org, presence: { message: PRESENCE_MESSAGE }
+
+  validates :question, presence: { message: PRESENCE_MESSAGE }
+
+  validates :type, presence: { message: PRESENCE_MESSAGE },
+                   uniqueness: { message: UNIQUENESS_MESSAGE,
+                                 scope: :question_id }
+
+  # =================
+  # = Class Methods =
+  # =================
 
   ##
   # deep copy the given annotation and all it's associations
@@ -54,6 +66,18 @@ class Annotation < ActiveRecord::Base
     annotation_copy = annotation.dup
     annotation_copy.save!
     return annotation_copy
+  end
+
+  # ===========================
+  # = Public instance methods =
+  # ===========================
+
+  ##
+  # returns the text from the annotation
+  #
+  # @return [String] the text from the annotation
+  def to_s
+    "#{text}"
   end
 
   def deep_copy(**options)
--- a/app/models/annotation.rb
+++ b/app/models/annotation.rb
@@ -51,7 +51,7 @@ class Annotation < ActiveRecord::Base
 
   validates :type, presence: { message: PRESENCE_MESSAGE },
                    uniqueness: { message: UNIQUENESS_MESSAGE,
-                                 scope: :question_id }
+                                 scope: [:question_id, :org_id] }
 
   # =================
   # = Class Methods =
--- a/app/models/guidance_group.rb
+++ b/app/models/guidance_group.rb
@@ -21,6 +21,9 @@
 
 class GuidanceGroup < ActiveRecord::Base
   include GlobalHelpers
+  include ValidationValues
+  include ValidationMessages
+
   ##
   # Associations
   belongs_to :org
@@ -30,8 +33,19 @@ class GuidanceGroup < ActiveRecord::Base
   # has_and_belongs_to_many :guidances, join_table: "guidance_in_group"
 
 
+  # ===============
+  # = Validations =
+  # ===============
 
-  validates :name, :org, presence: {message: _("can't be blank")}
+  validates :name, presence: { message:  PRESENCE_MESSAGE }
+
+  validates :org, presence: { message: PRESENCE_MESSAGE }
+
+  validates :optional_subset, inclusion: { in: BOOLEAN_VALUES,
+                                           message: INCLUSION_MESSAGE }
+
+  validates :published, inclusion: { in: BOOLEAN_VALUES,
+                                     message: INCLUSION_MESSAGE }
 
 
   # EVALUATE CLASS AND INSTANCE METHODS BELOW
--- a/app/models/guidance_group.rb
+++ b/app/models/guidance_group.rb
@@ -37,7 +37,8 @@ class GuidanceGroup < ActiveRecord::Base
   # = Validations =
   # ===============
 
-  validates :name, presence: { message:  PRESENCE_MESSAGE }
+  validates :name, presence: { message:  PRESENCE_MESSAGE },
+                   uniqueness: { message: UNIQUENESS_MESSAGE }
 
   validates :org, presence: { message: PRESENCE_MESSAGE }
 
--- a/app/models/guidance_group.rb
+++ b/app/models/guidance_group.rb
@@ -38,7 +38,7 @@ class GuidanceGroup < ActiveRecord::Base
   # ===============
 
   validates :name, presence: { message:  PRESENCE_MESSAGE },
-                   uniqueness: { message: UNIQUENESS_MESSAGE, scope: :org }
+                   uniqueness: { message: UNIQUENESS_MESSAGE, scope: :org_id }
 
   validates :org, presence: { message: PRESENCE_MESSAGE }
 
--- a/app/models/guidance_group.rb
+++ b/app/models/guidance_group.rb
@@ -38,7 +38,7 @@ class GuidanceGroup < ActiveRecord::Base
   # ===============
 
   validates :name, presence: { message:  PRESENCE_MESSAGE },
-                   uniqueness: { message: UNIQUENESS_MESSAGE }
+                   uniqueness: { message: UNIQUENESS_MESSAGE, scope: :org }
 
   validates :org, presence: { message: PRESENCE_MESSAGE }
 
--- a/app/models/identifier_scheme.rb
+++ b/app/models/identifier_scheme.rb
@@ -13,8 +13,29 @@
 #
 
 class IdentifierScheme < ActiveRecord::Base
+  include ValidationMessages
+  include ValidationValues
+
+  ##
+  # The maximum length for a name
+  NAME_MAXIMUM_LENGTH = 30
+
   has_many :user_identifiers
   has_many :users, through: :user_identifiers
-  
-  validates :name, uniqueness: {message: _("must be unique")}, presence: {message: _("can't be blank")}
+
+  # ===============
+  # = Validations =
+  # ===============
+
+  validates :name, uniqueness: { message: UNIQUENESS_MESSAGE },
+                   presence: { message: PRESENCE_MESSAGE },
+                   length: { maximum: NAME_MAXIMUM_LENGTH }
+
+  validates :logo_url, presence: { message: PRESENCE_MESSAGE }
+
+  validates :user_landing_url, presence: { message: PRESENCE_MESSAGE }
+
+  validates :active, inclusion: { message: INCLUSION_MESSAGE,
+                                  in: BOOLEAN_VALUES }
+
 end
--- a/app/models/identifier_scheme.rb
+++ b/app/models/identifier_scheme.rb
@@ -31,10 +31,6 @@ class IdentifierScheme < ActiveRecord::Base
                    presence: { message: PRESENCE_MESSAGE },
                    length: { maximum: NAME_MAXIMUM_LENGTH }
 
-  validates :logo_url, presence: { message: PRESENCE_MESSAGE }
-
-  validates :user_landing_url, presence: { message: PRESENCE_MESSAGE }
-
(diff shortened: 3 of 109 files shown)


briri:
Hey @Bodacious, 

@jollopre may find the exported_plans model/table useful for the work he will be doing with usage statistics (although I think a new streamlined data structure specific for that purpose may be better and could replace exported_plans). For now I would say go ahead and comment out the format validation on that model. 

For the identifier_schemes table you can remove the presence check for logo_url and landing_url. They are pulled in for UI display purposes. The idea/code behind that was originally for it to be somewhat generic so that we could easily add other oAuth options on the sign in form and the 'link your account' section of the profile page (e.g. Google). The UI code would then dynamically check for those 2 fields to determine what it displays to the user. I think its veered away from that though over the past year and become specific to the scheme.

app/models/guidance_group.rb
```diff
@@ -30,8 +33,19 @@ class GuidanceGroup < ActiveRecord::Base
...
+  validates :name, presence: { message:  PRESENCE_MESSAGE }
+
+  validates :org, presence: { message: PRESENCE_MESSAGE }
+
+  validates :optional_subset, inclusion: { in: BOOLEAN_VALUES,
+                                           message: INCLUSION_MESSAGE }
 
-  validates :name, :org, presence: {message: _("can't be blank")}
+  validates :published, inclusion: { in: BOOLEAN_VALUES,
+                                     message: INCLUSION_MESSAGE }
```
briri: Do we also want to check uniqueness for name for a given org? `validates :name, uniqueness: { message: UNIQUENESS_MESSAGE, scope: :org }` to keep names unique across an org.

app/models/identifier_scheme.rb
```diff
@@ -13,8 +13,29 @@
...
+
+  # ===============
+  # = Validations =
+  # ===============
+
+  validates :name, uniqueness: { message: UNIQUENESS_MESSAGE },
+                   presence: { message: PRESENCE_MESSAGE },
+                   length: { maximum: NAME_MAXIMUM_LENGTH }
+
+  validates :logo_url, presence: { message: PRESENCE_MESSAGE }
```
briri: ok for this to be blank

app/models/identifier_scheme.rb
```diff
@@ -13,8 +13,29 @@
...
+
+  validates :user_landing_url, presence: { message: PRESENCE_MESSAGE }
```
briri: ok for this to be blank

app/models/org.rb
```diff
@@ -33,38 +33,80 @@
...
-  validates :name, presence: {message: _("can't be blank")}, uniqueness: {message: _("must be unique")}
+
+  # ===============
+  # = Validations =
+  # ===============
+
+  validates :name, presence: { message: PRESENCE_MESSAGE },
+                   uniqueness: { message: UNIQUENESS_MESSAGE }
+
+  validates :abbreviation, presence: { message: PRESENCE_MESSAGE }
```
briri: This one should also be unique

app/models/org.rb
```diff
@@ -33,38 +33,80 @@
...
+
+  validates :is_other, inclusion: { in: BOOLEAN_VALUES,
+                                    message: INCLUSION_MESSAGE }
+
+  validates :language, presence: { message: PRESENCE_MESSAGE }
```
briri: can we default this to whatever the gettext default is?

app/models/org.rb
```diff
@@ -33,38 +33,80 @@
...
+
+  validates :contact_email, presence: { message: PRESENCE_MESSAGE,
```
briri: Why not also use our `validators/email_validator`?

app/models/org.rb
```diff
@@ -33,38 +33,80 @@
...
+                                        if: :feedback_enabled }
+
+  validates :org_type, presence: { message: PRESENCE_MESSAGE }
+
+  validates :feedback_enabled, inclusion: { in: BOOLEAN_VALUES,
+                                            message: INCLUSION_MESSAGE }
+
+  validates :feedback_email_subject, presence: { message: PRESENCE_MESSAGE,
```
briri: I think we can drop this now can't we? We removed it from the Org edit form. We would need to make sure that the user_mailer is no longer referencing the DB value though.

app/models/org.rb
```diff
@@ -178,20 +220,25 @@ def grant_api!(token_permission_type)
...
+      if logo.height != 100
+        self.logo = logo.thumb('x100')  # resize height and maintain aspect ratio
+      end
     end
+  end
 
+  # creates a dfefault Guidance Group on create on the Org
+  def create_guidance_group
+    GuidanceGroup.create!({
+      name: abbreviation? ? self.abbreviation : self.name ,
```
briri: @Bodacious what are your thoughts on models creating other models like this? Is it good practice? The other option I suppose would be to have this method on the `guidance_group` itself if not just a part of its `initialize` method.

We've gone back and forth on this project. It originally concerned me because it wasn't always apparent (due to method names) that a record was being created. For example we at one point had `copy = Template.deep_copy` which would make me expect that a copy of the model and its dependencies had been created but not yet persisted when in fact `create!` had been called for all of the objects. 

The name for this one is much clearer than deep_copy so maybe its ok. Should we add a '!' to the end of it though?

app/models/org_identifier.rb
```diff
@@ -17,14 +17,33 @@
...
+
   # Should only be able to have one identifier per scheme!
-  validates_uniqueness_of :identifier_scheme, scope: :org
-  
-  validates :identifier, :org, :identifier_scheme, presence: {message: _("can't be blank")}
-  
+  validates :identifier_scheme_id, uniqueness: { scope: :org_id,
+                                                 message: UNIQUENESS_MESSAGE }
+
+  validates :identifier, presence: { message: PRESENCE_MESSAGE }
```
briri: This one should be unique for the scheme. (e.g. one org's shibboleth id cannot be the same as any others)

app/models/question_format.rb
```diff
@@ -12,15 +12,25 @@
...
   has_many :questions
 
-  enum formattype: [ :textarea, :textfield, :radiobuttons, :checkbox, :dropdown, :multiselectbox, :date, :rda_metadata ]
+  enum formattype: FORMAT_TYPES
+
+  validates :title, presence: { message: PRESENCE_MESSAGE },
+                    uniqueness: { message: UNIQUENESS_MESSAGE }
 
-  validates :title, presence: {message: _("can't be blank")}, uniqueness: {message: _("must be unique")}
+  validates :description, presence: { message: PRESENCE_MESSAGE }
```
briri: description can be nil. Not sure if the field is even necessary.

app/models/template.rb
```diff
@@ -33,26 +33,57 @@
...
+  # = Validations =
+  # ===============
+
+  validates :title, presence: { message: PRESENCE_MESSAGE }
+
+  validates :org, presence: { message: PRESENCE_MESSAGE }
+
+  validates :locale, presence: { message: PRESENCE_MESSAGE }
+
+  validates :version, presence: { message: PRESENCE_MESSAGE }
```
briri: version should be unique across the family_id. We maintain that in the code but might be worth adding here

app/models/user.rb
```diff
@@ -88,14 +102,26 @@ def filter(query)
...
-  ##
-  # Scopes
+  # ===============
+  # = Validations =
+  # ===============
+
+  validates :active, inclusion: { in: BOOLEAN_VALUES,
+                                  message: INCLUSION_MESSAGE }
+
+
```
briri: email should be unique. Devise may manage that for us though ... not sure

db/seeds.rb
```diff
@@ -2,29 +2,69 @@
 # This file should contain all the record creation needed to seed the database with its default values.
 # The data can then be loaded with the rake db:seed (or created alongside the db with db:setup).
 
+require 'factory_bot'
+include FactoryBot::Syntax::Methods
+
```
briri: Yes! Thank you

app/models/identifier_scheme.rb
Bodacious: 👍

app/models/org.rb
Bodacious: I'm not sure what's required here. Should we wait for @xsrust to comment?

app/models/org.rb
Bodacious: It's not the worst approach. But I think a service object would create a better de-coupling, and make it clearer what is happening.

app/models/question_format.rb
Bodacious: @xsrust what do you think? can we remove the column?

app/models/user.rb
Bodacious: Yep, we include the `validatable` module in devise.

db/seeds.rb
Bodacious: My pleasure!

app/models/perm.rb
```diff
@@ -9,18 +9,38 @@
...
+
+  scope :modify_templates, -> { Perm.find_by(name: 'modify_templates') }
+
+  scope :modify_guidance, -> { Perm.find_by(name: 'modify_guidance') }
+
+  scope :use_api, -> { Perm.find_by(name: 'use_api') }
+
+  scope :change_org_details, -> { Perm.find_by(name: 'change_org_details') }
+
+  scope :grant_api, -> { Perm.find_by(name: 'grant_api_to_orgs') }
```
xsrust: Seeing as these records never change, would there be a downside to making these constants?
It would likely save quite a few queries, but these are so frequently accessed they likely get cached

app/models/question_format.rb
xsrust: Description can be nil, but in the case of the RDA question type, we need the field to hold the URL of the API.
We could change the name of the field on the db though, as 'description' isnt that descriptive of what we're using it for.

briri:
Some questions to consider:
- Duplicate Guidance groups names are probably ok if they belong to different orgs. I would suggest adding a 'scope' to that validation
- Our `is_other` Org (the one non-shibbolized users get attached to) has an id of zero in the DB. I believe this was introduced during our migration. This makes the `guidance_groups.org_id` blank check fail. I'm not sure whether its preferable to update the validation to allow zero or write a script to create a new row, reattach all dependencies to that new record and then drop record zero. I'm thinking the later. What are your thoughts?
- Should we allow the `user_identifiers` table to have more than one identifier per scheme? We have a handful of records like this. I suppose that its not inconceivable that a user could have shibbolized credentials at multiple institutions or if we add Google or other oauths later it might be that they have multiples.
- We have a lot of blank `annotations` and `answers` due to the way we were handling the create/update of those items in the past. Are we still creating empty records? If so we need to clean that up so that we're not perpetuating the issue

app/models/guidance_group.rb
Bodacious: Added

app/models/question_format.rb
Bodacious: Added new issue for that

app/models/annotation.rb
```diff
@@ -34,16 +39,23 @@ class Annotation < ActiveRecord::Base
...
-  end
+  validates :text, presence: { message: PRESENCE_MESSAGE }
+
+  validates :org, presence: { message: PRESENCE_MESSAGE }
 
+  validates :question, presence: { message: PRESENCE_MESSAGE }
+
+  validates :type, presence: { message: PRESENCE_MESSAGE },
+                   uniqueness: { message: UNIQUENESS_MESSAGE,
+                                 scope: :question_id }
```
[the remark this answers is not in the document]
Bodacious: Changed

app/models/perm.rb
Bodacious: They won't be cached across requests.

Instead of making them constants, it might be a better approach to explicitly cache them. E.g.

``` ruby
def self.grant_api
  cache.fetch(["perm", "grant_api"], expires_in: 1.hour) { Perm.find_by(name: 'grant_api_to_orgs') }
end
```

This would require some caching store to be set up. Most likely Redis or Memcached.

Bodacious:
- @briri yeesh! I think it would be better to update the Org to a new row. Otherwise you'll constantly be working around this in multiple places for a single edge case.
- Removed `user_identifiers` uniqueness validation for now. Since this is not essential.
- Can we create a new issue to clean up this empty Answer issue? It sounds like the original approach is wrong

PR #1757, opened 2018-07-25
UI updates for draggable sections

Bodacious:
@briri Here are the changes that were discussed last week.

Changes proposed in this PR:
- Show alert when dragging section to 1st place
- Extend notification code to make them floatable
- Remove wobble effect from failed sort
- Change sort handle to arrows instead of bars
- Add help icon to indicate sections are draggable

### DIFF
--- a/lib/assets/javascripts/utils/notificationHelper.js
+++ b/lib/assets/javascripts/utils/notificationHelper.js
@@ -1,32 +1,58 @@
 import { isString, isObject } from './isType';
+
 /*
   Helpers that will display the specified message in in the notification
   area at the top of the page
 */
-export const renderNotice = (msg) => {
+
+export function hideNotifications() {
+  $('#notification-area')
+    .addClass('hide')
+    .removeClass('notification-area--floating');
+}
+
+function renderMessage(options = {}) {
   const notificationArea = $('#notification-area');
 
-  if (isString(msg) && isObject(notificationArea)) {
-    notificationArea.removeClass('alert-warning').addClass('alert-info');
+  if (isString(options.message) && isObject(notificationArea)) {
+    notificationArea
+      .removeClass('alert-info', 'alert-warning')
+      .addClass(options.className);
+
+    if (options.floating) {
+      notificationArea.addClass('notification-area--floating');
+    }
+
     notificationArea.find('i, span').remove();
     notificationArea.append(`
-      <i class="fa fa-check-circle" aria-hidden="true"></i><span>${msg}</span>`);
+      <i class="fa fa-${options.icon}" aria-hidden="true"></i>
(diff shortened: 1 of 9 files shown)


Bodacious:
@briri please give this a go and let me know what you think? Happy to make updates before we merge if you think they'll make it more intuitive.

lib/assets/javascripts/utils/notificationHelper.js
```diff
@@ -1,32 +1,58 @@
...
-    notificationArea.removeClass('alert-info').addClass('alert-warning');
+  if (isString(options.message) && isObject(notificationArea)) {
+    notificationArea
+      .removeClass('alert-info', 'alert-warning')
+      .addClass(options.className);
+
+    if (options.floating) {
+      notificationArea.addClass('notification-area--floating');
+    }
+
```
briri: Reusing the `#notification-area` like this may cause issues with any global notifications that are active. If so it might be better to either pass in the `notificationArea` so that the caller can determine which DOM element they want to work with.

lib/assets/stylesheets/dmproadmap.scss
```diff
@@ -1,3 +1,5 @@
+@import "dmproadmap/notifications";
```
briri: I like this idea of breaking the css out into specific files

briri:
I think things look good @Bodacious. There's some weirdness going on with the fontawesome icons both becoming minus signs when you click to expand a section. There tinymce editor doesn't load either after adding a new section (I think this is unrelated to your changes though).

@stephaniesimms and @sjDCC can you test the changes made to the drag-and-drop template sections out on roadmap-stg. Let us know if there are any issues or desired changes (aside from the 2 points I've mentioned)

lib/assets/javascripts/utils/notificationHelper.js
Bodacious: This is how the original JS code was set to function. But I agree, it's not ideal. 

If you like, I can refactor this on a separate Issue?

Bodacious:
I've addressed the two issues described above in the latest two commits.

lib/assets/javascripts/utils/notificationHelper.js
briri: Ok. hold off on refactoring for now.

lib/assets/stylesheets/dmproadmap.scss
briri: Will create a separate ticket for that. The various class/id selectors should be reevaluated as well since we've completely refactored all of the org admin facing pages as well as Bhavi's accessibility changes. Its likely that quite a few are no longer necessary

PR #1764, opened 2018-08-01
updated translateable:find task to separate out app/views/branded domain

xsrust:
Updates `rake translateable:find[app]` task to separately generate pot/po files for the main app and for the branded files.
If `app` is passed to the task, it excludes `/app/views/branded` in it's search.
Otherwise, pass the name of your translation project.
Then only `app/views/branded` will be searched and `<domain>.pot` file will be created, along with the corresponding `<domain>.po` files

@briri I can also share with you the initilizer we're using for gettext translations using multiple domains.

### DIFF
--- a/lib/tasks/translatable.rake
+++ b/lib/tasks/translatable.rake
@@ -12,13 +12,13 @@ namespace :translatable do
       puts "You must provide a ISO-639 language code, name: `rake translatable:add_language[ja,日本語]`"
     end
   end
-  
+
   desc 'Remove the specified language from the database'
   task :remove_language_from_db, [:code] => [:environment] do |t, args|
     if args[:code].present?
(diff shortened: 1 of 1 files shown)


lib/tasks/translatable.rake
```diff
@@ -198,14 +201,19 @@ namespace :translatable do
     return hash
   end
 
-  def files_to_translate
-    Dir.glob("{app,lib,config,locale}/**/*.{rb,erb,md,haml,slim,rhtml}")
+  # TODO: exclude app/views/branded
```
briri: I had forgotten that we set the site up to look for branded versions of files there 😆 
Will need to move my files from my 'dmptool' subdirs to 'branded'
We should also consider doing the same in the JS dir

PR #1768, opened 2018-08-01
Consolidate validation.js and ariatiseForm.js

briri:
Hey @xsrust and @Bodacious. Here's the start to the new and hopefully improved JS for form validations. Its not finished completely yet. I need to go through the html pages and make sure things are wired up properly and working as expected. 

The new file is called `lib/assets/javascripts/utils/validator.js`. Please take a look at it now before I get much further (and before this PR gets too big) and let me know what you think. In summary the new file allows you to:
- add/remove validation handling for an entire form (like the old ariatiseForm.js did)
- add/remove validation for a specific form field (for dom elements created via ajax)
- Run a validation or all form validations on demand

The code works with some new 'validatable-x' attributes on the form fields. It then:
- Adds a required field asterisk if appropriate
- Adds an error message block for the field
- Adds an onblur handler so that the validation fires when the user tabs off the field (can be disabled)
- Adds a onsubmit handler for the form


Bodacious:
Hi @briri 

The JS code looks great so far! Very well written.

briri:
Discussed with the team and we will be removing the JS field validation logic and instead allow Rails to process validations and report back to the user. Now that we have proper model level validations in place this will simplify maintenance going forward.

Need to:
- Update controller logic to return field level validation errors so that they can be displayed in the normal notification area.
- Update JS to only provide support for the accessible autocomplete controls and to add the required field asterisks
- Ensure that all necessary aria tags are on the appropriate fields

PR #1769, opened 2018-08-02
Replace Minitest test suite  with RSpec test suite

### LINKED ISSUE #1664: Update test suite to use RSpec

Bodacious:
Closes #1664 .

Changes proposed in this PR:
- Added specs for all models
- Added feature specs for
  - Plan creation
  - Registration
  - Sessions
- Removed deprecated method calls
- Removed Minitest test suite


app/controllers/registrations_controller.rb
```diff
@@ -66,14 +66,14 @@ def create
...
-          params[:user][:org_id] = other_org.id 
+          params[:user][:org_id] = other_org.id
         end
         build_resource(sign_up_params)
         if resource.save
           if resource.active_for_authentication?
             set_flash_message :notice, :signed_up if is_navigational_format?
             sign_up(resource_name, resource)
-            UserMailer.welcome_notification(current_user).deliver
+            UserMailer.welcome_notification(current_user).deliver_now
```
briri: Thanks for cleaning up these deprecated calls

app/models/plan.rb
```diff
@@ -460,13 +448,8 @@ def owner_and_coowners
...
           .includes(:question_options, { question: :question_format })
-          .reduce(0) do |m, a|
-
-      if a.is_valid?
-        m += 1
-        m
-      end
-    end
+          .to_a
+          .sum { |answer| answer.is_valid? ? 1 : 0 }
```
briri: Nice. much cleaner

app/models/question_option.rb
```diff
@@ -52,6 +52,10 @@ class QuestionOption < ActiveRecord::Base
   scope :by_number, -> { order(:number) }
 
 
+  # ===========================
+  # = Public instance methods =
+  # ===========================
+
```
briri: Probably just need the one comment block ;)
No need to hold off on merging due to this though

PR #1775, opened 2018-08-03
Fix bug preventing admin from adding new Section to Template

### LINKED ISSUE #1371: Organisation customised templates - section display ordering incorrect

Bodacious:
Fixes #1371  .

Changes proposed in this PR:
- Remove published column from sections table
- Add specs for Template editing
- Add ignore file for Brakeman to pass TravisCI

New changes introduced to the validations has caused the section creation to stop working (issue #1371). Specifically, sections were not valid if published wasn't set, and this wasn't being set anywhere in the code.

This PR removes the offending validation code, as well as the DB column from the DB. There are additional schema comments that have been added by the `annotate` gem. These do not affect the function of the code.


.travis.yml
```diff
@@ -13,23 +13,23 @@ cache:
...
+
 addons:
   chrome: beta
   apt:
-    sources:
-      - google-chrome
     packages:
       - nodejs
       - google-chrome-stable
+      - google-chrome-stable
```
briri: duplicate package declaration

PR #1789, opened 2018-08-07
Add Brakeman ignore file to pass Travis

Bodacious:
Ignores certain issues that we think are OK

Fixes failing tests on TravisCI.

Changes proposed in this PR:
- Add a Brakeman ignore file to ignore the issues we think aren't a security risk.


briri:
Wow there are quite a few. I'd imagine we want to go back and clean these up properly at some point soon and remove the Brakeman exceptions?

PR #1801, opened 2018-08-13
Fix for "Plan download file not filtering on selection on Download pa…

johnpinto1:
Fixes # 1730.

Changes proposed in this PR:
- The previous code did not filter the output based on selection of
    (i) phases (if there was multiple phase)
    (ii) customised sections were not filtered based on checkbox "supplementary section(s) not requested by funding organisation"

- Also fixed rendering error in Plan Controller when saving or updating a plan (in certain cases) failed.

### DIFF
--- a/app/controllers/plans_controller.rb
+++ b/app/controllers/plans_controller.rb
@@ -162,19 +162,9 @@ class PlansController < ApplicationController
 
     plan, phase = Plan.load_for_phase(params[:id], params[:phase_id])
 
-    readonly = !plan.editable_by?(current_user.id)
-
     guidance_groups =  GuidanceGroup.where(published: true, id: plan.guidance_group_ids)
-    # Since the answers have been pre-fetched through plan (see Plan.load_for_phase)
-    # we create a hash whose keys are question id and value is the answer associated
-    answers = plan.answers.reduce({}){ |m, a| m[a.question_id] = a; m }
 
-    render('/phases/edit', locals: {
-      base_template_org: phase.template.base_org,
-      plan: plan, phase: phase, readonly: readonly,
-      guidance_groups: guidance_groups,
-      answers: answers,
-      guidance_service: GuidanceService.new(plan) })
+    render_phases_edit(plan, phase, guidance_groups)
   end
 
   # PUT /plans/1
@@ -196,13 +186,13 @@ class PlansController < ApplicationController
           format.json {render json: {code: 1, msg: success_message(_('plan'), _('saved'))}}
         else
           flash[:alert] = failed_update_error(@plan, _('plan'))
-          format.html { render action: "edit" }
+          format.html { render_phases_edit(@plan, @plan.phases.first, @plan.guidance_groups) }
           format.json {render json: {code: 0, msg: flash[:alert]}}
         end
 
       rescue Exception
         flash[:alert] = failed_update_error(@plan, _('plan'))
-        format.html { render action: "edit" }
+        format.html { render_phases_edit(@plan, @plan.phases.first, @plan.guidance_groups) }
         format.json {render json: {code: 0, msg: flash[:alert]}}
       end
     end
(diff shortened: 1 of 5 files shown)


app/helpers/plans_helper.rb
```diff
@@ -42,4 +42,9 @@ def visibility_tooltip(val)
       return _('Private: restricted to me and people I invite.')
     end
   end
+
+  def download_plan_page_title(plan, phase, hash)
+    # If there is more than one phase show the plan title and phase title
+    return hash[:phases].length > 1 ? "#{plan.title} - #{phase[:title]}" : plan.title
```
Bodacious: @johnpinto1 

Please remove the redundant `return` here, and replace `.length > 1` with `.many?`

That should read a bit nicer :)

app/views/shared/export/_plan.erb
```diff
@@ -32,55 +32,59 @@
...
-          <h3><%= section[:title] %></h3>
-        <% end %>
+      <%# Only render selected phase %>
+      <% if phase[:title] == @selected_phase.title %>
+        <div style="page-break-before:always;"></div> <!-- Page break before each phase -->
+        <h1><%= download_plan_page_title(@plan, phase, @hash) %></h1>
+        <hr />
+        <% phase[:sections].each do |section| %>
+          <%# Only render if @show_sections_questions is true and either section not customised or section is customised and @show_custom_sections is true %>
+          <% if @show_sections_questions && (!section[:modifiable] || (@show_custom_sections && section[:modifiable]))  %>
```
Bodacious: It would be good to extract this out to a method, in the same way you have `download_plan_page_title`

app/views/shared/export/_plan.erb
```diff
@@ -32,55 +32,59 @@
...
-                <% if ah['standards'].present? %>
+              <% answer = @plan.answer(question[:id], false) %>
+              <% blank = answer.present? ? answer.is_blank? : true %>
+              <% options = answer.present? ? answer.question_options : [] %>
+              <%# case where question has not been answered sufficiently to display%>
+              <% if @show_unanswered && (answer.blank? || (options.blank? && blank))%>
+                <p><%= _('Question not answered.') -%></p>
+              <% else %>
+                <%# case where Question has options %>
+                <% if options.present?%>
```
Bodacious: No need to check if `options.present?`. It should always be an Array, so we can remove this conditional .

Alternatively, if you don't want an empty `<ul>` tag, use `options.any?` to make it clearer that `options` is an Enumerable object.

app/models/concerns/exportable_plan.rb
```diff
@@ -17,23 +17,24 @@ def as_csv(headings = true, unanswered = true)
...
-            flds = (hash[:phases].length > 1 ? [phase[:title]] : [])
-            if headings
-              if question[:text].is_a? String
-                question_text = question[:text]
+        if  selected_phase.nil? || phase[:title] == selected_phase.title
+          phase[:sections].each do |section|
+            section[:questions].each do |question|
+              answer = self.answer(question[:id], false)
+              answer_text = answer.present? ? answer.text : (unanswered ? _('Not Answered') : '')
+              flds = (hash[:phases].length > 1 ? [phase[:title]] : [])
```
Bodacious: Prefer `.many?` over `.length > 1`

johnpinto1:
Updated code based on comments. 

Also fixed **as_csv** in **ExportPlanModule** to work correctly with selections made on Plan Download page.

app/controllers/plans_controller.rb
```diff
@@ -439,4 +433,23 @@ def request_feedback_flash_notice
...
     feedback_constant_to_text(text, current_user, @plan, current_user.org)
   end
+
+  private
+
+  # ============================
+  # = Private instance methods =
+  # ============================
+
+  def render_phases_edit(plan, phase, guidance_groups)
```
briri: Nice refactor :)

PR #1802, opened 2018-08-13
Remove redundant database tables and columns

### LINKED ISSUE #1382: DB Cleanup
this ticket is derived from #170 

These are some general cleanup tasks for the DB that we should perform:
- Remove the old friendly_id gem from the Gemfile and drop its DB elements
  - Drop table `friendly_id_slugs`
  - Drop field `plans.slug`
- Drop the following tables which as no longer used: `file_types`, `file_uploads`, `regions`, `splash_logs`
- Remove the `sections.published` field. Templates are published, not individual sections
- Remove the `orgs.banner_text` field since it is no longer used.
- Remove the `guidances.question_id` column (and update model if necessary). Guidances stored in this table are associated with themes not questions
- Drop the old `admin` and `org_admin` records from the `perms` table and also any references to them in `users_perms` these are the old permission types which have been replaced by our more granular permissions
- Drop the `orcid_id` and `shibboleth_id` fields from the `users` table. These identifiers have been moved to the `user_identifiers` table. If necessary create a rake task to port existing data to the new DB table

Copying more notes from a duplicate issue to drop the following unused tables. There may be overlap w/bullets above:
We should drop the following tables from the database and any controllers/models/routes associated with them since they are no longer used:
- file_types
- file_uploads,
- friendly_id_slugs
- question_format_labels (part of dmptool migration that is no longer used)
- regions
- splash_logs

### LINKED ISSUE #1719: Remove slug attributes from models
There are quite a few `slug` attributes on the models that don't seem to be in use. 

Can these be removed?

Bodacious:
Changes proposed in this PR:
- Remove redundant columns and tables listed in #1382 
- Update models and factories with new schema structure comments

Fixes #1382 
Fixes #1719

### DIFF
--- a/app/controllers/registrations_controller.rb
+++ b/app/controllers/registrations_controller.rb
@@ -5,8 +5,8 @@ class RegistrationsController < Devise::RegistrationsController
     @user = current_user
     @prefs = @user.get_preferences(:email)
     @languages = Language.sorted_by_abbreviation
-    @orgs = Org.where(parent_id: nil).order("name")
-    @other_organisations = Org.where(parent_id: nil, is_other: true).pluck(:id)
+    @orgs = Org.order("name")
+    @other_organisations = Org.where(is_other: true).pluck(:id)
     @identifier_schemes = IdentifierScheme.where(active: true).order(:name)
     @default_org = current_user.org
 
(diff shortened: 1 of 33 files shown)


briri:
Changes look good to me. Just need to address the failing tests. Feel free to merge once they're passing

Bodacious:
@briri Sorry, I meant to ask. Could you double-check that it's fine to remove `orgs.parent_id`?

xsrust:
we are not using the field in any of our databases, but wait for @briri to confirm before merging. 

Other than that, looks good to me.

briri:
nope. we have no need of the field. I think this was a legacy DMPOnline_v3 field that allowed for Org hierarchies at one point

PR #1807, opened 2018-08-15
Feature/validations

Bodacious:
Changes proposed in this PR:
- Adds a new rake task for much faster checking that known validations issues have been fixed.
- Update .gitignore to ignore SQL files
- Update Plan Rule for better performance


briri:
looks good

Issue #1804, opened 2018-08-14
JS linter errors introduced by PR 1778


briri:
PR #1778 introduced 2 new JS linter errors
```shell
/home/travis/build/DMPRoadmap/roadmap/lib/assets/javascripts/views/shared/create_account_form.js
  3:8   error  Using exported name 'validateOrgSelection' as identifier for default export  import/no-named-as-default
  4:10  error  'isValidText' is defined but never used                                      no-unused-vars
```

stephaniesimms:
can't test in UI. closing issue.

PR #1810, opened 2018-08-15
Fix bug preventing orgs loading on register page

### LINKED ISSUE #1809: Orgs won't load on registration page

**Expected behaviour:**
When I go to register a new account, I should see my Org in the Orgs dropdown field.

**Actual behaviour:**
There are no Orgs to choose from

Bodacious:
Recent updates to the DB have made the `is_other` field on the Orgs table mandatory.
Previously, it was loading null values. This PR updates this to load false values instead.

Fixes #1809

Changes proposed in this PR:
- Load orgs where `is_other` is false, instead of null

### DIFF
--- a/lib/tasks/assets.rake
+++ b/lib/tasks/assets.rake
@@ -15,7 +15,7 @@ namespace :assets do
       # Ensure all dependencies are installed
       system("npm install")
       # Run the webpack command via npm
-      system("npm run bundle -- #{webpack_options.join(" ")}")
+      system("npm run bundle -- #{webpack_options.join(" ")}") or exit()
     end
   end
 end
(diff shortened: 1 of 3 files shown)


lib/tasks/assets.rake
```diff
@@ -15,7 +15,7 @@ namespace :assets do
       # Ensure all dependencies are installed
       system("npm install")
       # Run the webpack command via npm
-      system("npm run bundle -- #{webpack_options.join(" ")}")
+      system("npm run bundle -- #{webpack_options.join(" ")}") or exit()
```
Bodacious: @briri This should ensure Travis fails early if the webpack script doesn't compile properly

Bodacious:
Should help with #1804

PR #1811, opened 2018-08-15
Change tooltip on container page if all modifiable

### LINKED ISSUE #1371: Organisation customised templates - section display ordering incorrect

Bodacious:
Fixes #1371  .

Changes proposed in this PR:
- If all of the sections are modifiable, we don't need the second part of the tooltip that reads: "You may place them before or after the main template sections."
Only show this if some of the Sections are unmodifiable.


briri:
looks good. tests are failing due to JS issues on dev branch that have since been resolved

PR #1812, opened 2018-08-15
Remove bad migration

briri:
Removes a bad index from `templates` that was forcing `org_id, customization_of and version` to be unique. This was preventing an org from having multiple templates at version zero.


briri:
@Bodacious or @xsrust can you see if you can figure out why its throwing a migration error. I'm having trouble running the tests on my new machine. I'm reluctant to run `db:migrate` because it adds limits to all the fields and changes up some of the Key names

Bodacious:
this should be fixed once #1832 is merged in.

PR #1814, opened 2018-08-16
Reset other_organisation when User org changes

### LINKED ISSUE #1736: Reset other_organisation name when a User changes org
When a User changes from one Org to another, the System should reset the `other_organization` column to prevent stale data from lingering.

Bodacious:
Fixes #1736.

Changes proposed in this PR:
- Add callback to user model
- Refactored how user model defines callbacks
- Add specs


app/models/user.rb
```diff
@@ -132,7 +132,11 @@ class User < ActiveRecord::Base
   # = Callbacks =
   # =============
 
-  after_update :when_org_changes
+  before_update :clear_other_organisation, if: :org_id_changed?
+
+  after_update :delete_perms!, if: :org_id_changed?, unless: :can_change_org?
+
+  after_update :remove_token!, if: :org_id_changed?, unless: :can_change_org?
```
xsrust: This is much cleaner. makes it easier to tell what happens on a save.

briri:
Looks good to me. I just restarted the failing Travis build. It was complaining about not being able to find web pack

PR #1823, opened 2018-08-17
Add tests for locale setting

Bodacious:
Changes proposed in this PR:
- Adds Faker::Language module for testing languages with real-ish data
- Refactors global constants into Language model (See Issue#1696)
- Add specs to test locale change works


config/initializers/constants.rb
```diff
@@ -2,11 +2,3 @@
 MAX_NUMBER_LINKS_FUNDER = 5
```
briri: These could probably be moved elsewhere too. Maybe the branding.yml would be a good candidate

PR #1824, opened 2018-08-19
Issue 1173

jollopre:
- History tables that are updated through rake tasks:
  - initially (e.g. stat:joined_user:generate:full_all_orgs)
  - monthly (e.g. stat:joined_user:generate:last_month_all_orgs)
- Models capturing monthly aggregates for created plans and joined
users:
  - Use of Single Table Inheritance from rails through Stat model
  - class method for generating CSV data according to a set of instances
  (e.g. StatCreatedPlan or StatJoinedUser)
  - class method to retrieve data for a monthly range through
  StatCreatedPlan.monthly_range or StatJoinedUser.monthly_range
- Introduction of actions to avoid polluting models
- Unit test for each newly model created

### DIFF
--- a/app/actions/stat_created_plan/generate.rb
+++ b/app/actions/stat_created_plan/generate.rb
@@ -1,0 +1,48 @@
+module Actions
+  module StatCreatedPlan
+    class Generate
+      class << self
+        def full(org)
+          OrgDateRangeable.split_months_from_creation(org) do |start_date, end_date|
+            create_count_for_date(start_date: start_date, end_date: end_date, org: org)  
+          end
+        end
+
+        def last_month(org)
+          months = OrgDateRangeable.split_months_from_creation(org)
+          last = months.last
+          if last.present?
+            create_count_for_date(start_date: last[:start_date], end_date: last[:end_date], org: org)
+          end
+        end
+        
+        def full_all_orgs
+          Org.all.each do |org|
+            full(org)
+          end
+        end
+
+        def last_month_all_orgs
+          Org.all.each do |org|
+            last_month(org)
+          end
+        end
+
+        private
+
+        def count_plans(start_date: , end_date: , org:)
+          users = User.where('users.org_id = ?', org.id)
+          plans = Plan.where('plans.created_at >= ? AND plans.created_at <= ?', start_date, end_date)
+          owner_coowner = Role.where(access: Role.access_values_for(:creator).concat(Role.access_values_for(:administrator)).uniq)
+          
+          Role.joins([:plan, :user]).merge(owner_coowner).merge(users).merge(plans).select(:plan_id).distinct.count
+        end
+
+        def create_count_for_date(start_date:, end_date:, org:)
+          count = count_plans(start_date: start_date, end_date: end_date, org: org)
+          ::StatCreatedPlan.create(date: end_date.to_date, count: count, org_id: org.id)
+        end
+      end
+    end
+  end
+end
--- a/app/models/stat.rb
+++ b/app/models/stat.rb
@@ -1,0 +1,13 @@
+class Stat < ActiveRecord::Base
+  belongs_to :org
+
+  class << self
+    def to_csv(stats)
+      data = stats.reduce([]) do |acc, stat|
+        acc << { date: stat.date, count: stat.count }
+        acc
+      end
+      Csvable.from_array_of_hashes(data)
+    end
+  end
+end
--- a/lib/csvable.rb
+++ b/lib/csvable.rb
@@ -1,0 +1,16 @@
+module Csvable
+  class << self
+    def from_array_of_hashes(data = [])
+      headers = data.first && data.first.keys
+      if headers
+        return CSV.generate do |csv|
+          csv << headers
+          data.each do |row|
+            csv << row.values
+          end
+        end
+      end
+      ''
+    end
+  end
+end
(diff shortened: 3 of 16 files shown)


app/models/stat.rb
```diff
@@ -0,0 +1,13 @@
+class Stat < ActiveRecord::Base
+  belongs_to :org
+
+  class << self
+    def to_csv(stats)
+      data = stats.reduce([]) do |acc, stat|
```
Bodacious: Any reason why you didn't use `Array#map` here?

``` ruby
data = stats.map  { |acc, stat| { date: stat.date, count: stat.count } }
```

lib/csvable.rb
```diff
@@ -0,0 +1,16 @@
+module Csvable
+  class << self
+    def from_array_of_hashes(data = [])
```
Bodacious: Suggested refactor:

``` ruby
def from_array_of_hashes(data = [])
  return '' unless data.first&.keys
  headers = data.first.keys
  CSV.generate do |csv|
    csv << headers
      data.each do |row|
        csv << row.values
      end
    end
  end
end
```

lib/org_date_rangeable.rb
```diff
@@ -0,0 +1,27 @@
+module OrgDateRangeable
+  def monthly_range(org:, start_date: Date.today.end_of_month, end_date: Date.today.end_of_month)
+    raise ArgumentError.new('missing org') unless org.present?
+    raise ArgumentError.new('missing start_date') unless start_date.respond_to?(:end_of_month)
```
Bodacious: Instead of raising an exception if the parameter doesn't respond to an expected method, you can force the type by explicitly calling the `to_[type]` method. 

By _soft testing_ the type of an object, the code should be more flexible and robust.

``` ruby 
# bad
def yell_name(name)
  raise ArgumentError, "Not a string" unless name.respond_to?(:uppercase)
  puts name.uppercase
end

# good
def yell_name(name)
  puts name.to_s.uppercase
end
```

Bodacious: ... in this case, just use `start_date.to_date`

lib/org_date_rangeable.rb
```diff
@@ -0,0 +1,27 @@
...
+    raise ArgumentError.new('missing end_date') unless end_date.respond_to?(:end_of_month)
+
+    where(["org_id = ? and date >= ? and date <= ?", org.id, start_date, end_date])
```
Bodacious: Prefer named variables in SQL strings:

``` ruby
    where("org_id = :org_id AND date >= :start_date AND :end_date <= ?", org_id: org.id, start_date: start_date, end_date: end_date) 
```

lib/org_date_rangeable.rb
```diff
@@ -0,0 +1,27 @@
...
+    raise ArgumentError.new('missing end_date') unless end_date.respond_to?(:end_of_month)
+
+    where(["org_id = ? and date >= ? and date <= ?", org.id, start_date, end_date]) 
+  end
+
+  class << self
+    def split_months_from_creation(org, &block)
+      starts_at = org.created_at
+      ends_at = starts_at.end_of_month
+      callable = block.nil? ? lambda{ |start_date, end_date| } : lambda{ | start_date, end_date| block.call(start_date, end_date) }
```
Bodacious: If you use `Proc.new {}` here, you won't need to define the unused variables in the first block. 
Procs don't complain if parameters are missing.

app/models/stat.rb
jollopre: just forgot, indeed I'm more inclined to create immutable data...

lib/csvable.rb
jollopre: nice usage of safe navigation operator

lib/org_date_rangeable.rb
jollopre: it is also good for methods declaration involving more than two arguments...

lib/org_date_rangeable.rb
jollopre: thanks for sharing these notes. Since the query will run with any argument passed (except for org that send ':id'), I think is best removing duck typing checks. What do you think?

jollopre:
@Bodacious, @briri thanks for the review. I made the changes suggested.

app/actions/stat_created_plan/generate.rb
```diff
@@ -0,0 +1,48 @@
...
+            last_month(org)
+          end
+        end
+
+        private
+
+        def count_plans(start_date: , end_date: , org:)
+          users = User.where('users.org_id = ?', org.id)
+          plans = Plan.where('plans.created_at >= ? AND plans.created_at <= ?', start_date, end_date)
+          owner_coowner = Role.where(access: Role.access_values_for(:creator).concat(Role.access_values_for(:administrator)).uniq)
```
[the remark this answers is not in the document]
Bodacious: It’s defined on the development branch, but not master.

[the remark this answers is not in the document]
jollopre: ok, should I assume that we want the this PR based on development then?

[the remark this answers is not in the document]
Bodacious: I'm not sure. @xsrust what do you think?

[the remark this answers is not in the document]
xsrust: @jollopre We had originally thought Master made more sense as it's more stable, but with Development diverging so far from it this sprint, it may make sense to rebase off development.

There's been work done to clean up the models so you'd be using the re-factored functions.
The downside is that there's still quite a few bugs on the dev branch at the moment, but we should be clearing those out soon, and as this feature is new, it shouldnt conflict with anything else.

Rebasing would also mean that Travis runs the up to date CI against the PRs.

Issue #1755, opened 2018-07-24
Guidance w/multiple themes displayed multiple times


stephaniesimms:
A single piece of guidance tagged w/multiple themes is displayed multiple times (once for each theme - see screenshot) per question tagged w/multiple themes. This guidance/theme display is problematic no matter what we do it seems. @briri thinks it should be possible to create a logic whereby org guidance that has multiple themes would only be displayed once, e.g. just display the first theme it's tagged with.

![screen shot 2018-07-23 at 9 30 18 am](https://user-images.githubusercontent.com/12703638/43090220-a617a6ea-8e5b-11e8-9065-6ee62a41d4e6.png)

PR #1829, opened 2018-08-21
updated guidance on only display once per question.  Added break betw…

xsrust:
#1755 

only output each guidance once per question.

Also adds separation between individual guidance when output within the same theme

### DIFF
--- a/app/views/guidance_groups/_index_by_theme.html.erb
+++ b/app/views/guidance_groups/_index_by_theme.html.erb
@@ -9,34 +9,47 @@
     </div>
   </div>
   <% guidance_groups_by_theme.each_pair do |guidance_group, theme_hash| %>
+    <% guidances_output = [] %>
     <% theme_hash.each_pair do |theme, guidances| %>
-      <div class="panel panel-default">
-        <div
-          class="heading-button"
-          role="button"
-          data-toggle="collapse"
-          href="<%= "##{guidances.object_id}" %>"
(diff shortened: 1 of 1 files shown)


app/views/guidance_groups/_index_by_theme.html.erb
```diff
@@ -9,34 +9,47 @@
...
-          data-toggle="collapse"
-          href="<%= "##{guidances.object_id}" %>"
-          aria-expanded="false"
-          aria-controls="<%= "##{guidances.object_id}" %>">
-          <div class="panel-heading" role="tab" id="<%= "panel-heading-#{guidances.object_id}" %>">
-            <h2 class="panel-title">
-              <%= theme.title %>
-              <i class="fa fa-plus pull-right" aria-hidden="true"></i>
-            </h2>
+      <% if (guidances.map{|g| g.id} - guidances_output).any? # any not output%>
```
Bodacious: - Can you move the comment to above the line please? (Trying to stay strict to the style guide)

- Suggested minor refactor: `guidances.map(&:id)`

xsrust: addded the comment above and updated the statement.  
I need to get in the habit of using the `&:`

Bodacious:
The broken tests should be fixed once #1832 is merged in.

This broken test bug was caused by a combination of refactoring how Languages are loaded and the cache suite being added.

PR #1830, opened 2018-08-21
Added new Rspect tests to org_spec.rb.

johnpinto1:
Updated Rspec tests in **spec/org_spec.rb**.

### DIFF
--- a/spec/models/org_spec.rb
+++ b/spec/models/org_spec.rb
@@ -346,15 +346,16 @@ RSpec.describe Org, type: :model do
   describe "#plans" do
 
     let!(:org) { create(:org) }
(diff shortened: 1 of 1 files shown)


spec/models/org_spec.rb
```diff
@@ -363,28 +364,83 @@
 
     context "when user belongs to Org and plan user with role :administrator" do
 
-      pending("TBD") do
+      before do
+        create(:role, :administrator, user: user)
+        plan.send(:add_user, user.id, false, true, false); # Calling private method #add_user using #send
```
Bodacious: Please move the trailing comment to above the line

Issue #1752, opened 2018-07-19
Consolidate JS validation code


briri:
We currently have 2 JS files that deal with form field validations. The original and most heavily used is `utils/ariatiseForm.js`. This file though only allows for validations to occur on form submit and does not allow for us to make fields 'validatable' based on certain scenarios (e.g. link text only becomes required when a user enters a link URL). 

To deal with that scenario we created a `utils/validations.js` file which allows you to 'enable/disable' validations on a given control. With the intent of updating all form validations so that they use this new file and trigger a validation check when the user tabs off the field as well as when the form is submitted.

Our JS validation alerts are also not accessible #1446. We should address that while reconciling these 2 files.

These files are beginning to become out of sync and I fear will become maintenance nightmares as time goes by and new devs enter the mix. We should refactor `utils/validations.js` so that it is properly handling all use cases and then update all of the individual form js files to use it.

briri:
Discussed with the team and we will be removing the JS field validation logic and instead allow Rails to process validations and report back to the user. Now that we have proper model level validations in place this will simplify maintenance going forward.

Need to:
- Update controller logic to return field level validation errors so that they can be displayed in the normal notification area.
- Update JS to only provide support for the accessible autocomplete controls and to add the required field asterisks
- Ensure that all necessary aria tags are on the appropriate fields

sjDCC:
Am unsure if we can UAT this. Please advise what we should test if so

stephaniesimms:
i dont think we can UAT exactly but would notice issues w/field form validations, etc if things were amiss. will close this issue and just be vigilant when testing all the things.

Issue #899, opened 2017-12-05
Theme Rails' default 404 and 500 error pages.


briri:
Please complete the following fields as applicable:

**Expected behaviour:**
Rails has built-in handling for any 404 and 500 errors that occur when running in production mode (dev and other modes should continue to display the detailed error information)
Refer to the ActionController documentation for handling generic 404 and 500 errors
http://guides.rubyonrails.org/action_controller_overview.html#the-default-500-and-404-templates

**Actual behaviour:**
Screenshot of a 404 error when running in production mode:

![screen shot 2017-12-05 at 10 48 29 am](https://user-images.githubusercontent.com/1204467/33625716-5a28d29e-d9ad-11e7-94df-feb7fb4b2b9c.png)

**Steps to reproduce:**
1. Update your database.yml to ensure that you will not be testing against any live data!
2. Make sure your production secrets key is set in secrets.yml
3. Start the application in production mode `bundle exec rails s -e production`
4. Login
5. Go to a fake plan page (e.g. http://localhost:3000/plans/i5ty495yth5)

magdalenadrafiova:
@briri is this still relevant?

briri:
yes. this is still something we should consider doing.

PR #1838, opened 2018-08-23
Form validations (DO NOT MERGE)

briri:
For #1752 
Started work on the new form validations logic. Please review the approach before I start going through and updating all the views/controllers accordingly:
- Removed the ariatiseForm.js and validations.js files and all references
- Removed validation error messages for JS from the `views/layouts/application.html.erb`
- Added a generic requiredField.js which continues to add the asterisk to fields marked `aria-required="true"`. This was originally part of the 2 JS files above.
- Consolidated the 3 error message helpers in ApplicationController to one `failure_message` method
- Added a `views/layouts/_error.html.erb`
- Added a named yield section to `views/layouts/application.html.erb` notification section. Individual views can render `content_for` that named section
- Simplified the Notifications and Theme controllers/views in SuperAdmin namespace and updated their save logic so that the user isn't driven back out to index when they create/update a record
- Added some CSS to take advantage of Rails `field_with_errors` class that it adds to form elements that have errors so that the boxes have a red border. 
- I also added a generic ActiveRecord:RecordNotFound handler to the ApplicationController that will redirect to a branded 404 page in production (related to #899)

Let me know what you think of the approach. Happy to adjust if there's a better/more conventional way to handle things.


xsrust:
+1 for the interactive rebase, but I like the approach taken here.

briri:
Closing as this one is no longer relevant

PR #1847, opened 2018-08-27
Issue 532

### LINKED ISSUE #532: Replace "raw" & "html_safe" with "sanitize" for security reasons
For security reasons (elimination of XSS attacks) we should replace `raw` & `html_safe` instances with `sanitize` which has very sane defaults out of the box:

https://apidock.com/rails/ActionView/Helpers/SanitizeHelper/sanitize

Bodacious:
Fixes #532  .

Changes proposed in this PR:
- Replace instances of unsafe `raw` method calls with safer alternatives
- Remove exceptions from brakeman ignores file.

### DIFF
--- a/app/controllers/concerns/paginable.rb
+++ b/app/controllers/concerns/paginable.rb
@@ -131,7 +131,16 @@ module Paginable
     if @paginable_params[:sort_field] == sort_field
       className = upcasing_sort_direction == 'ASC'? 'fa-sort-asc' : 'fa-sort-desc'
     end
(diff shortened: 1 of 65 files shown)


Bodacious:
Test are passing. Travis has failed because of this issue and bundle-audit https://github.com/rubyzip/rubyzip/issues/369

xsrust:
I kicked travis and it looks like there are actually 2 tests failing as well as the bundle audit.

Other than that the changes look good and the defaults for sanitize sensible

PR #1849, opened 2018-08-28
Fix bug breaking template selection when creating plan

### LINKED ISSUE #1843: Can't create plans w/funder or Org templates
I can only create a new plan with the default template: if I check the box for "No funder" and there are no Org templates. 

If I try to use an Org template I get the error message in screenshot 1. If I try to select a funder, the appropriate selections appear in the combobox but I get the error message in screenshot 2.

screen shot 2018-08-24 at 2 23 06 pm

screen shot 2018-08-24 at 2 20 26 pm

Bodacious:
Fixes #1843  .

Changes proposed in this PR:
- Fix bug breaking template selection when creating plan
- Refactor template options as own resource


app/controllers/org_admin/template_options_controller.rb
```diff
@@ -0,0 +1,67 @@
+# frozen_string_literal: true
+
+class OrgAdmin::TemplateOptionsController < ApplicationController
```
xsrust: I like the refactor of this out of the templates controller into it's own.

Semantically, I dont think this shouldn't be under the org-admin module.
The action is accessed when regular users are trying to create a plan.

app/controllers/plans_controller.rb
```diff
@@ -23,9 +23,9 @@ def new
     # Get all of the available funders and non-funder orgs
     @funders = Org.funder
                   .joins(:templates)
-                  .where(templates: { published: true }).uniq.sort(&:name)
+                  .where(templates: { published: true }).uniq.sort_by(&:name)
```
xsrust: How much difference is there between sort and sort_by ?

xsrust:
Kicked the rspec tests over and they are passing

app/controllers/plans_controller.rb
Bodacious: Depends how academic you want to get about it. There are performance differences, since `sort_by` uses more effective algorithm behind the scenes. But the main practical difference is that `sort` doesn't take a method name proc in the way I used it above. So this was a bug.

app/controllers/org_admin/template_options_controller.rb
Bodacious: Yes, I agree. 

I kept it there because that's where it was originally. But I've just pushed out a new commit to move it into the main namespace.

app/controllers/org_admin/templates_controller.rb
```diff
@@ -278,67 +278,8 @@ def unpublish
       redirect_to request.referrer.present? ? request.referrer : org_admin_templates_path
     end
 
-    # GET /org_admin/template_options  (AJAX)
```
briri: Good . Thanks for moving this to the appropriate controller

Bodacious:
@briri I think it was the introduction of `sort` instead of `sort_by`. This was my bad.

briri:
Ah ok. so it was returning an empty result set. Makes sense

briri:
Some conflicts in the routes file. Would feel more comfortable if you resolve them @Bodacious since you're closer to all of the recent changes. Feel free to merge this yourself once they're cleaned up

briri:
never mind. the conflicts were mostly due to some differences in tabbing/spacing of the routes.rb code.

Bodacious:
@briri Thanks!

PR #1851, opened 2018-08-29
Fix bug preventing super admin from changing orgs

### LINKED ISSUE #1835: Super-admin organisation switcher not working
The templates page on the admin interface has a place for Stephanie and I to change affiliation. This is currently throwing an error. You can successfully change organisation via edit profile though.

Please note that this should only display for super-admin users.

ord-switcher

Bodacious:
Fixes #1835 

Changes proposed in this PR:
- Fix bug preventing super admin from changing orgs
- Extract org swaps into own resource
- Add integration specs for org swaps


briri:
👍

PR #1863, opened 2018-09-06
[Dont Merge] Configurable installation of db and server gems

xsrust:
We are using the Puma server for production and development at DCC so I'm adding it to the gemfile for the main project.  

Adding it shouldnt have any detrimental effects, but I wanted to flag the addition to you @briri.


briri:
I have concerns about adding deployment strategies that are specific to individual installations. This is a shared project that goes beyond just our two products. 

We're already removing some gems like 'pg' at deploy time via our Capistrano script. This prevents us from having to install and support Postgres on a machine that doesn't use it. 

We use Passenger at the moment and manage that in our own fork. You should be able to do the same pretty easily. I'm happy to discuss and share what we do if you would find it helpful.

xsrust:
Further to your concerns, @briri the database specific gems, along with server-specific gems, have been added to their own gem-groups.  

In order to  not install or use a subset of these, [bundler offers](https://bundler.io/v1.16/guides/groups.html#how-to-manage-groups-of-gems) the `--without` flag.  In addition, bundler will cache your passed options, so the next time you run a `bundle install`, it will skip the options previously passed to without.

example usage for you use case should be:
`bundle install --without puma thin psql` for production, as you need none of the dev servers nor postgresql.

briri:
Thanks for the explanation and setting those gems up into their respective groups.

@Bodacious I remember you modifying those bin/* files a few months back to check for the `/spring` directory and to see if bundler had been invoked via a script (I don't recall if it was for Travis or new installations). If we no longer need those checks feel free to merge. 👍

Bodacious:
@briri Changes to `bin` directory look good to me.

These are the same changes I see when I run `rake rails:update`, so they're the default files for Rails 4.2

PR #1865, opened 2018-09-06
Remove JS Form validations

### LINKED ISSUE #1752: Consolidate JS validation code

briri:
Fixes #1752

- Remove old JS form validations
- Update message handlers in application_controller.rb
- Update all controllers to use new message handlers

Also went through and fixed up all of the controllers so that they pass Rubocop tests


briri:
rebasing to remove old commits will resubmit PR

PR #1868, opened 2018-09-07
added validation tests

briri:
Add in tests for new validations on Guidance and User models


briri:
added a `config/locales/fr.yml` to try and deal with some CI test failures that complain about it not existing

PR #1869, opened 2018-09-10
Fix naming format of settings controller

Bodacious:
Changes proposed in this PR:
- Fix naming convention for settings controller. This was breaking in production mode


briri:
Interesting. Change looks fine to me

Issue #1173, opened 2018-02-15
Create Usage reports that include all org data for users and plans


stephaniesimms:
Usage Dashboard:
Right now as we wanted something workable, the CSV files are calculated on demand, i.e. any time you press download in any topic, 12 queries to the db are send, one for each month. The ideal solution for this would be to have history tables that are updated automatically each month (e.g. cron job). That would mean that we would reduce from 12 to 1 query.

- History tables that are updated monthly   
- Each table should capture monthly aggregates per organisation for the following topics:   
    - Users joined   
    - Plans created   
    - Plans by template (will be done in #1679)   
- Note, the rows are sorted descending by date for each topic   
- The data (since the organisation was created in the service) must be CSV downloadable. The file could have the following format for no. users joined:

Date | No. users joined
-- | --
June 2018 | ...
May 2018 | ...
April 2018 | ...
... | ...

Date | No. plans created
-- | --
June 2018 | ...
May 2018 | ...
April 2018 | ...
... | ...

jollopre:
@sjDCC, @stephaniesimms can we mark this issue as completed? There is no much to test from your side right now, however there is plenty of test in place to satisfy the DoD.

sjDCC:
Yes, fine by me. This is incorporated into the tickets you're working on

PR #1871, opened 2018-09-11
Issue 1689

jollopre:
This PR addresses totals for stats, currently "Created Plans" and "Joined Users". There is a button enabled at the usage view that permits downloading a CSV with these totals.

Remember, in order to get the numbers will be needed to execute the rake tasks that populate stats table:

```
rake stat:created_plan:generate:full_all_orgs
rake stat:joined_user:generate:full_all_orgs
```


app/actions/org/total_stat.rb
```diff
@@ -0,0 +1,55 @@
+require './app/services/org/stat_joined_user'
```
Bodacious: Avoid using `require` in Rails app files. These should be autoloaded on each new request.

If the files aren't being autoloaded, you might need to add `config.autoload_paths += [ Rails.root.join('app', 'services')]` to **application.rb**

More info in the Rails docs here:
https://guides.rubyonrails.org/autoloading_and_reloading_constants.html

app/controllers/usage_controller.rb
```diff
@@ -1,14 +1,26 @@
-# frozen_string_literal: true
```
Bodacious: Please be careful to not remove _magic comments_, these serve a purpose

https://freelancing-gods.com/2017/07/27/friendly-frozen-string-literals.html

Rubocop will fail when these are removed.

config/routes.rb
```diff
@@ -123,7 +123,9 @@
     end
   end
 
-  resources :usage, only: [:index]
+  resources :usage, only: [:index] do
+    get :all, action: :all, on: :collection, as: :all, constraints: { format: :csv }
```
Bodacious: Format constraints might not work the way you expect here. From the rails docs:

> There is an exception for the format constraint: while it's a method on the Request object, it's also an implicit optional parameter on every path. Segment constraints take precedence and the format constraint is only applied as such when enforced through a hash. For example, get 'foo', constraints: { format: 'json' } will match GET  /foo because the format is optional by default. However, you can use a lambda like in get 'foo', constraints: lambda { |req| req.format == :json } and the route will only match explicit JSON requests.

I'll need to change a couple of the routes that I added recently too 🙈

config/application.rb
```diff
@@ -84,7 +84,7 @@ class Application < Rails::Application
     # parameters by using an attr_accessible or attr_protected declaration.
     #config.active_record.whitelist_attributes = true
 
-    config.autoload_paths += %W(#{config.root}/lib)
+    config.autoload_paths += %W(#{config.root}/lib #{config.root}/app)
```
Bodacious: This isn't required. Directories within the app dir are automatically autoloaded. You might need to restart the app and/or stop Spring first `spring stop`.

config/routes.rb
```diff
@@ -123,7 +123,9 @@
     end
   end
 
-  resources :usage, only: [:index]
+  resources :usage, only: [:index] do
```
Bodacious: It looks like you're just defining an index action here. Can we move this into its own controller (`UsageDownloadsControlller#index`, or into the `UsageController#index` action?

Also, remember to use plural names for Rails resources.

config/application.rb
Bodacious: Seeing the same issue @jollopre has reported here. Fine to keep this in when needed.

jollopre:
@Bodacious thanks for the comments. I've decided to remove the concept of Action which adds complexity to this codebase and use 'Rails Services' instead. In addition, I have created the UsageDownloadsController to better align with the 'retrieve all', i.e. index action. 

Although this issue did not have to deal with renaming actions such as Actions::StatCreatedPlan::Generate or Actions::StatJoinedUser::Generate because were addressed in #1173, I renamed them to services too. This has led to rename the stat tasks that were invoking the previous mentioned actions. Now, we have the following options for populating the db with stats:

```
rake stat:create:created_plan             # Creates created plan stats for every org since they joined
rake stat:create:joined_user              # Creates joined user stats for every org since they joined
rake stat:create_last_month:created_plan  # Creates created plan stats for today's last month for every org
rake stat:create_last_month:joined_user   # Creates joined user stats for today's last month for every org
```

PR #1874, opened 2018-09-13
Clean up pending and failing specs

Bodacious:
- Fix pending Template model specs
- Fix pending model specs
- Fix broken/pending specs, including random "errors" bug


app/models/template.rb
```diff
@@ -185,15 +189,10 @@ class Template < ActiveRecord::Base
   # Retrieves unarchived templates whose title or org.name includes the term
   # passed
   scope :search, lambda { |term|
-    unarchived.where("templates.title LIKE :term OR orgs.name LIKE :term",
-                     term: "%#{term}%")
+    unarchived.joins(:org).where("templates.title LIKE :term OR orgs.name LIKE :term",
```
Bodacious: This fixes a bug that was discovered in testing.

spec/support/matchers/has_errors_matcher.rb
```diff
@@ -3,7 +3,7 @@
 RSpec::Matchers.define :have_errors do |expected|
 
   match do
-    actual.body.match(/Error/)
+    actual.body.match(/Error\:/)
```
Bodacious: This was raising intermittent errors, since Lorem ipsum contains the word "error"

PR #1875, opened 2018-09-13
Fix broken tests caused by locale configuration

Bodacious:
- Add LocaleSet
- Add LocaleFormatter to force proper string formats
- Rename I18n formats to use hyphen
- Add rake task for updating existing languages on DB
- Fix instances of US english in the templates

## Locale formats 
FastGettext expects locales to be in the format 'en_GB'. I18n uses the ISO standard for language formats 'en-GB' (a hyphen instead of an underscore). FastGettext [does re-format locale strings](https://github.com/grosser/fast_gettext/blob/master/lib/fast_gettext/storage.rb#L182) when setting locale (e.g. `FastGettext.locale = "en-GB" # => "en_GB"`), but the lookups were tending to fail, causing lots of unpredictable bugs in testing. 

See this issue for more info: https://github.com/grosser/fast_gettext/issues/81

This PR adds a couple of helper classes to mediate between I18n expectations and FastGettext expectations. Namely, `LocaleFormatter` and `LocaleSet`.

By forcing the use of these classes throughout the code, it seems to have sorted most of the sporadic failures in the tests.

## Default locale

There was another issue that was happening sporadically: Some tests would fail because the text on the page was the wrong locale. I think this was some sort of race condition, in which the test was inspecting the page _before_ strings had been updated in the browser by FastGetext. This meant that strings like "Customize" were being loaded in the page in US English, whereas the tests were searching for GB English strings "Customise". Since `en-GB` is the default locale, I rewrote those strings to use GB spelling.

## Migration

I've added a rake task `rake upgrade:normalize_language_formats` to change the locale formats for existing Languages, Themes, and Templates on the DB.


config/initializers/fast_gettext.rb
```diff
@@ -24,6 +26,10 @@ def available_locales
...
 
+I18n.available_locales += available_locales.for(:i18n)
+FastGettext.default_available_locales = available_locales.for(:fast_gettext)
+
 FastGettext.default_text_domain       = 'app'
-FastGettext.default_locale            = default_locale
-FastGettext.default_available_locales = available_locales
+
+I18n.default_locale        = available_locales.for(:i18n).first
+FastGettext.default_locale = available_locales.for(:fast_gettext).first
```
xsrust: This looks like a clean way of managing the fastGettext vs I18n expectations.

config/locales/contact_us/contact_us.pt_BR.yml
```diff
@@ -1,25 +0,0 @@
-# Sample localization file for English. Add more files in this directory for other locales.
```
xsrust: We have a full translation for Brazillian due to their translation work with CDL so we'll actually want to keep and update this file rather than remove it.

lib/tasks/upgrade.rake
```diff
@@ -1,6 +1,21 @@
 require 'set'
 namespace :upgrade do
 
+  desc "Update Language abbreviations to use ISO format"
+  task :normalize_language_formats => :environment do
```
xsrust: We can flag this needs to be run with the release

PR #1887, opened 2018-09-18
Change success message on permission change

### LINKED ISSUE #1880: Change msg text for removing User permissions
As an admin, when I remove user permissions on the Users page and click 'Save' the banner message is weird (see screenshot). Please update text to the following for all permissions changes:

"Successfully saved permissions"

screen shot 2018-09-16 at 4 08 44 pm

Bodacious:
Fixes #1880  .

Changes proposed in this PR:
- Change success flash notice message when permissions are updated.

@xsrust I suggest we leave this one for @briri to review, since he's done work recently on generalising the flash notices and might want to have a say in this change.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -84,7 +84,7 @@ class UsersController < ApplicationController
       end
       render(json: {
         code: 1,
-        msg: success_message(perms.first, _("saved")),
+        msg: _("Successfully saved permissions"),
         current_privileges: render_to_string(partial: "users/current_privileges",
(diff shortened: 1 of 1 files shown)


briri:
It looks like the perms.first is a nil for some reason and being passed into that success_message function. If you pass in a `Perm` it should produce the correct error message. 

I think it would be better to send in `Perm.new` than hard code a message here so that we can keep those messages consistent in the future should we need to modify them.

Bodacious:
@briri agreed, that's probably a better solution. Have updated now.

Issue #1754, opened 2018-07-20
Add a button to download a list of users


stephaniesimms:
Feedback from DMPTool admin webinar:
Is it possible for administrators to get a dynamic or downloadable up to date list of users versus screen scraping the list we see on the admin page? This will also help us identify resources and outreach more programmatically and hopefully greater adoption of the DMPTool. 

- Add a 'Download users' button to create a CSV report from the Admin Users page, just as we have on the Plans page.

sjDCC:
This works fine - thanks for adding. Could we put the button in the same place as the 'plans' page though and call it 'Download users' to keep things consistent? I couldn't spot it at first. Thanks

![download-users](https://user-images.githubusercontent.com/6169940/46549581-76fc5f00-c8ca-11e8-8a6a-5854bf523bc8.PNG)
![download-users2](https://user-images.githubusercontent.com/6169940/46549580-76fc5f00-c8ca-11e8-8ede-8c5a88aee8aa.PNG)

stephaniesimms:
i second @sjDCC request to move and rename the button for consistency with the plans page (screenshot above). otherwise it's great @johnpinto1 
- move button to upper right above the table
- change button text to 'Download users'

sjDCC:
hi @johnpinto1 Sorry to be finnicky about this, but could we ensure buttons are aligned so it looks slick and change the text to "Download users" for consistency please. Thanks :)

![align](https://user-images.githubusercontent.com/6169940/46687154-0e76f000-cbf2-11e8-8078-3aa175bf4db1.PNG)
![align2](https://user-images.githubusercontent.com/6169940/46687155-0e76f000-cbf2-11e8-8740-a46863ad0eb2.PNG)

stephaniesimms:
button and functionality are mostly working as expected now. the only issue i noticed is with date formats, which i expect is a configuration setting. will create a new issue for that. closing this one.

johnpinto1:
@stephaniesimms At present the date format is hard-coded '%d.%m.%Y' in the code at present. I forgot that the US have a different date format. I can look at this tomorrow as part of this. Perhaps @xrust, @Bodacious, @briri could suggest how I get round date format issue depending on locale the issue in ***models/user/at_csv.rb***: 


```
def to_csv
      CSV.generate(headers: true) do |csv|
        csv << HEADERS
        @users.each do |user|
          name = "#{user.firstname} #{user.surname}"
          email = user.email
[...]

PR #1892, opened 2018-09-25
Add a button to download a list of users by Org admins in CSV format.

johnpinto1:
Issue #1754

Fixes # 1754.

Changes proposed in this PR:
- Added functionality for Org admins to download users lists as viewed on "User Accounts" page.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -15,9 +15,17 @@ class UsersController < ApplicationController
     authorize User
     if current_user.can_super_admin?
       @users = User.page(1)
+      @users_to_csv = User.all
     else
       @users = current_user.org.users.page(1)
+      @users_to_csv = current_user.org.users.order(:surname)
     end
+
+    respond_to do |format|
+      format.html
+      format.csv { send_data User::AtCsv.new(@users_to_csv).to_csv, filename: "users-accounts-#{Date.today}.csv" }
+    end
+
   end
 
   ##
--- a/app/models/user/at_csv.rb
+++ b/app/models/user/at_csv.rb
@@ -1,0 +1,35 @@
+class User
+  class AtCsv
+
+    HEADERS = ['Name', 'E-Mail', 'Created Date', 'Last Activity', 'Plans', 'Current Privileges',  'Active']
+
+    def initialize(users)
+      @users = users
+    end
+    
+    def to_csv
+      CSV.generate(headers: true) do |csv|
+        csv << HEADERS
+        @users.each do |user|
+          name = "#{user.firstname} #{user.surname}"
+          email = user.email
+          created = user.created_at.strftime('%d.%m.%Y')
+          last_activity = user.updated_at.strftime('%d.%m.%Y')
+          plans = user.plans.size
+          active = user.active ? 'Yes' : 'No'
+      
+          if user.can_super_admin?
+            current_privileges = 'Super Admin'
+          elsif  user.can_org_admin?
+            current_privileges = 'Organisational Admin'
+          else
+            current_privileges = ''
+          end
+          
+          csv << [ name, email, created, last_activity, plans, current_privileges,  active ]
+        end
+      end
+    end
+    
+  end
+end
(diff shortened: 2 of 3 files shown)


Bodacious:
That was @johnpinto1's original approach, and I asked him to change it.

My thinking was that the User model is the file with the highest churn within the codebase, and adding the CSV stuff in there was only going to further exacerbate that by adding more responsibility for that model.

Generally, I consider things like CSV to be presentation code, and so it doesn't belong in the model, but I appreciate that Rails's convention normally encourages this sort of pattern.

If you wanted to call a method on the User model, maybe a sensible compromise would be to do:

```
class User < ActiveRecord::base
   
  # ... stuff

  def self.to_csv
    # self would be whatever the current scope is.
    User::AtCsv.new(self).to_csv
  end

  # ... stuff
end

# This should work...
@csv = Org.last.users.order(:email).to_csv
```

This would make it reusable with scopes attached, etc., but keep all of the CSV-specific stuff in it's own object.

briri:
Ah ok. That does make sense. The user does get some heavy use and already has all of the Devise bloat attached to it.

I like the compromise of adding the wrapper function.

app/models/user/at_csv.rb
```diff
@@ -0,0 +1,35 @@
...
+  class AtCsv
+
+    HEADERS = ['Name', 'E-Mail', 'Created Date', 'Last Activity', 'Plans', 'Current Privileges',  'Active']
+
+    def initialize(users)
+      @users = users
+    end
+    
+    def to_csv
+      CSV.generate(headers: true) do |csv|
```
briri: I cannot get this to generate my CSV file. I'm at 20 minutes and counting on my local with 31,500 users when logged in as a super admin.

We should either hide the button for super admins, only allow it to appear and process filtered results, or at least include a message next to the button warning that it may take up to 10+ minutes to generate the CSV.

@stephaniesimms and @sjDCC is downloading the list of all users something either of you would need? The easiest option would be for us to hide this button for super admins (we did something similar on another page if I recall due to the huge size of the download/query).

app/controllers/users_controller.rb
```diff
@@ -15,9 +15,17 @@ def admin_index
     authorize User
     if current_user.can_super_admin?
```
briri: I'd propose doing something like this to prevent the code from running both queries for all requests:
```ruby
case request.format
    when 'text/csv' # there might be a better way to test for the format
      if current_user.can_super_admin?
        @users = User.all.order(:surname)
      else
        current_user.org.users.order(:surname)
      end
    else
      if current_user.can_super_admin?
        @users = User.page(1)
      else
        @users = current_user.org.users.page(1)
      end
    end
```

app/controllers/users_controller.rb
```diff
@@ -15,9 +15,17 @@ def admin_index
...
       @users = User.page(1)
+      @users_to_csv = User.all
     else
       @users = current_user.org.users.page(1)
+      @users_to_csv = current_user.org.users.order(:surname)
     end
+
+    respond_to do |format|
+      format.html
+      format.csv { send_data User::AtCsv.new(@users_to_csv).to_csv, filename: "users-accounts-#{Date.today}.csv" }
```
briri: Then once you've added the wrapper method that @Bodacious recommended, you can update this to something like:
```ruby
  format.csv { send_data @users.map(&:to_csv), filename: "users-accounts-#{Date.today}.csv" }
```

johnpinto1: I replaced this with `format.csv { send_data User.to_csv(@users), filename: "users-accounts-#{Date.today}.csv" }`

app/controllers/users_controller.rb
johnpinto1: @briri Updated code. I have excluded super admin user viewing the *Download Csv* button.

johnpinto1:
@briri, @sjDCC & @stephaniesimms  With the current changes (based on @briri comments) I have ensured that only Org admins (but not super admins) see the **Download Csv button**.

app/controllers/users_controller.rb
johnpinto1: @briri & @Bodacious : I based my changes on comments from both. All mistakes are mine.

briri:
Thanks @johnpinto1

PR #1894, opened 2018-09-25
fixed issue with autosave

### LINKED ISSUE #1878: Autosave not working on write plan page
Expected behavior is the following: previously the system would autosave anything I input into a text editor (for questions only, not comments) on the write plan page as soon as I moved focus away from the editor. This action was accompanied by the timestamp appearing below the editor (screenshot 1). 

Now I can enter text and click around, opening and closing accordions, etc without autosave being triggered and no timestamps (screenshot 2). If I navigate to a different tab and return to 'Write plan' my answers are lost. I have to click the 'Save' button in order to save my answers now. This is happening for all existing and new plans. Desired behavior is to reinstate autosave for questions.

screen shot 2018-09-16 at 2 59 01 pm

screen shot 2018-09-16 at 2 53 50 pm

briri:
Fixes #1878  .

JS was checking length of the html within the `answer-locking` element. Recent reformatting of the HTML for Rubocop/readability added some blank space to the element which was causing the length to always be greater than zero.

Updated to check the length of the element's text instead.


Bodacious:
Well caught!

xsrust:
nice catch

PR #1895, opened 2018-09-25
Fixed issues with Feedback settings on Org page

### LINKED ISSUE #1877: Unable to switch on feedback as org admin
To test #1845 I tried to configure feedback for a few different orgs and found I couldn't switch this on. Initially I got an error that there had to be an email message (my message text was deleted on save) and then I got an error about the logo.
turn-on-feedback
request-feedback

briri:
Fixes #1877 .
- Updated JS so that the Feedback message is enabled/disabled appropriately based on the status of the feedback_enabled radio buttons
- Added a `before_validation` callback that will set the feedback subject to the default subject defined in the FeedbacksHelper.rb. There is a rake task that sets this value for existing records but nothing sets it for new Orgs.

A better long term solution might be to remove this DB field completely since we no longer allow users to define their own email subject. We should discuss and create a new ticket if that sounds like a good solution.

### DIFF
--- a/app/models/org.rb
+++ b/app/models/org.rb
@@ -31,6 +31,7 @@
 class Org < ActiveRecord::Base
   include ValidationMessages
   include ValidationValues
+  include FeedbacksHelper
   include GlobalHelpers
   include FlagShihTzu
   extend Dragonfly::Model::Validations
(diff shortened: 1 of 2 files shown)


xsrust:
Looks like a good update to me.

I agree we should add an issue to remove this from the DB if it's not configurable per-org.  As is, we could use the helper when we are displaying this.  

Added:
https://github.com/DMPRoadmap/roadmap/issues/1898

PR #1896, opened 2018-09-26
Fix bug in paginable table sorting

### LINKED ISSUE #1861: Sort by columns wrong color and display field name

**Expected behaviour:**
'Sort by' is indicated by white ascending/descending arrows

**Actual behaviour:**
'Sort by' is indicated by blue text (e.g. 'Sort by updated_at'). 

**Steps to reproduce:**
- Login
- Look at My Dashboard table(s)

Bodacious:
Fixes #1861 .

Changes proposed in this PR:
- Fix bug preventing paginable links from updating when clicked
- Fix bug: JS was loading multiple nested duplicate .paginable objects
- Refactor Paginable for improved readability.

### DIFF
--- a/app/controllers/concerns/paginable.rb
+++ b/app/controllers/concerns/paginable.rb
@@ -147,7 +147,7 @@ module Paginable
          style="float: right; font-size: 1.2em;">
 
         <span class="screen-reader-text">
-          #{_("Sort by %{sort_field}") % sort_field.split(".").first}
+          #{_("Sort by %{sort_field}") % { sort_field: sort_field.split(".").first }}
         </span>
       </i>
     HTML
--- a/app/controllers/concerns/paginable.rb
+++ b/app/controllers/concerns/paginable.rb
@@ -8,6 +8,9 @@ module Paginable
   # Regex to validate sort_field param is safe
   SORT_COLUMN_FORMAT = /[\w\_]+\.[\w\_]/
 
+  PAGINATION_QUERY_PARAMS = [:page, :sort_field, :sort_direction,
+                             :search, :controller, :action]
+
   private
 
   # Renders paginable layout with the partial view passed
@@ -75,24 +78,12 @@ module Paginable
 
   # Returns the base url of the paginable route for a given page passed
   def paginable_base_url(page = 1)
-    url_for(
-      @paginable_path_params.merge(
-        controller: @paginable_params[:controller],
-        action: @paginable_params[:action], page: page
-      )
+    url_params = @paginable_path_params.merge(
+      controller: @paginable_params[:controller],
+      action: @paginable_params[:action],
+      page: page
     )
-  end
-
-  # Returns the base url of the paginable router for a given page passed together with
-  # its query_params. It is used to retain context, i.e. search, sort_field,
-  # sort_direction, etc
-  def paginable_base_url_with_query_params(page: 1, **stringify_query_params_options)
-    base_url = paginable_base_url(page)
-    stringified_query_params = stringify_query_params(stringify_query_params_options)
-    if stringified_query_params.present?
-      return "#{base_url}?#{stringified_query_params}"
-    end
-    base_url
+    url_for(url_params)
   end
 
   # Generates an HTML link to sort given a sort field.
@@ -101,8 +92,9 @@ module Paginable
     link_to(
       sort_link_name(sort_field),
       sort_link_url(sort_field),
-      "data-remote": true,
-      class: "paginable-action", "aria-label": "#{sort_field}"
+      class: "paginable-action",
+      data: { remote: true },
+      aria: { label: sort_field }
     )
   end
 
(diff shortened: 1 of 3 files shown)


lib/assets/javascripts/utils/paginable.js
```diff
@@ -4,14 +4,16 @@ export const paginableSelector = '.paginable';
 
 $(() => {
   const onAjaxSuccessHandler = (e, data) => {
-    $(e.target).closest(paginableSelector).html($(data));
+    $(e.target).closest(paginableSelector).replaceWith($(data));
```
Bodacious: @briri @xsrust Can you double check this change for me?

The behaviour I saw was that each click on a sort link added a new `.paginable` DOM tree inside the `.paginable` one. I opted for `replaceWith` to fix that, and then moved the event listener to the 'body' element to catch click events on objects just added to the page.

app/controllers/concerns/paginable.rb
```diff
@@ -8,6 +8,9 @@ module Paginable
   # Regex to validate sort_field param is safe
   SORT_COLUMN_FORMAT = /[\w\_]+\.[\w\_]/
 
+  PAGINATION_QUERY_PARAMS = [:page, :sort_field, :sort_direction,
```
briri: 👍

app/controllers/concerns/paginable.rb
```diff
@@ -101,8 +92,9 @@ def paginable_sort_link(sort_field)
     link_to(
       sort_link_name(sort_field),
       sort_link_url(sort_field),
-      "data-remote": true,
-      class: "paginable-action", "aria-label": "#{sort_field}"
+      class: "paginable-action",
+      data: { remote: true },
+      aria: { label: sort_field }
```
briri: Didn't realize you could achieve the same result by breaking up the attribute name like this. Will start using this convention going forward 👍

app/controllers/concerns/paginable.rb
```diff
@@ -152,67 +130,62 @@ def refine_query(scope)
...
     end
     <<~HTML.html_safe
       <i class="fa #{class_name}"
          aria-hidden="true"
          style="float: right; font-size: 1.2em;">
 
         <span class="screen-reader-text">
-
-        Sort by #{sort_field.split(".").first}
+          Sort by #{sort_field.split(".").first}
```
briri: We should probably wrap this for FastGetText `_("Sort by #{sort_field...}")`.

Good catch on the missing `</span>`

app/controllers/concerns/paginable.rb
```diff
@@ -152,67 +130,62 @@ def refine_query(scope)
...
+      query_params[:sort_direction] = sort_direction.opposite
     else
-      sort_url = paginable_base_url_with_query_params(
-        page: page,
-        sort_field: sort_field)
+      query_params[:sort_direction] = sort_direction
     end
-    "#{sort_url}#{stringify_nonpagination_query_params}"
+    base_url = paginable_base_url(query_params[:page])
+    sort_url = URI(base_url)
```
briri: Good to see the use of URI here

lib/assets/javascripts/utils/paginable.js
briri: If it was nesting them then this should correct the issue and keep the DOM tree clean.
Delegating the listener to the body should work as well. We'll have to test all of the various scenarios to make sure the listeners still work as the table reloads (e.g. sorting after search, the view all link, etc.)

lib/assets/javascripts/utils/paginable.js
```diff
@@ -4,14 +4,16 @@ export const paginableSelector = '.paginable';
...
   };
   $('.paginable-search form[data-remote="true"]').on('ajax:before', e => isValidText($(e.target).find('input[name="search"]').val()));
   // Event listener for Ajax success event captured in response to a paginable link clicked or
   // search form submitted. Note the presence of a selector for on (e.g. a[data-remote="true"])
   // so that descendant elements from .paginable-results that are added in future are also
   // automatically handled.
-  $(paginableSelector).on('ajax:success', 'form.paginable-action[data-remote="true"], a.paginable-action[data-remote="true"]', onAjaxSuccessHandler);
+  $('body').on('ajax:success',
+    'form.paginable-action[data-remote="true"], a.paginable-action[data-remote="true"]',
```
briri: We should also check pages where there are multiple paginable tables (e.g. My Dashboard) to make sure that firing the success handlers for both tables doesn't cause any issues

lib/sort_direction.rb
```diff
@@ -0,0 +1,56 @@
...
+# Examples:
+#
+#   @direction = SortDirection.new(:asc)
+#   @direction.to_s # => "ASC"
+#   @direction.downcase # => 'asc'
+#   @direction.opposite # => 'DESC'
+#
+#   SortDirection.new(:wrong).to_s # => 'ASC'
+#
+class SortDirection
```
briri: I really like this. It removes a lot of complexity out of the paginable code.

app/controllers/concerns/paginable.rb
xsrust: neither did I, it's much easier to read though

lib/assets/javascripts/utils/paginable.js
xsrust: Good example of something to check durring UAT. Do we have a process for flagging things like this to @sjDCC and @stephaniesimms to ensure they get checked in UAT?

app/controllers/concerns/paginable.rb
Bodacious: @briri @xsrust  very handy when you have multiple nested attributes such as:

`data: { toggle: "dropdown", target: "#my-dropdown", user_id: 1234 }`

PR #1897, opened 2018-09-26
Update default locale

### LINKED ISSUE #1890: Default language set to Deutsch/German
When I first navigate to the app https://dmponline-dev.dcc.ac.uk the default language setting is German. After I log in it remembers my profile setting. I assume the desired behavior is for the default to be English-GB on public pages.

briri:
Fixes #1890 .

Updated the FastGettext initializer to use the `default_locale` logic when setting the default locales for both I18n and FastGettext. Included logic to swap out '-' and '_' in the locale names for the appropriate localization plugin

### DIFF
--- a/config/initializers/fast_gettext.rb
+++ b/config/initializers/fast_gettext.rb
@@ -1,7 +1,9 @@
+# frozen_string_literal: true
+
 # When Travis runs this, the DB isn't always built yet.
 if Language.table_exists?
   def default_locale
-    Language.default.try(:abbreviation) || 'en-GB'
+    Language.default.try(:abbreviation) || "en-GB"
   end
 
   def available_locales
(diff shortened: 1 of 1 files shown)


config/initializers/fast_gettext.rb
```diff
@@ -31,5 +31,5 @@ def available_locales
 
 FastGettext.default_text_domain       = 'app'
 
-I18n.default_locale        = available_locales.for(:i18n).first
-FastGettext.default_locale = available_locales.for(:fast_gettext).first
+I18n.default_locale        = default_locale.gsub(/_/, '-')
+FastGettext.default_locale = default_locale.gsub(/\-/, '_')
```
xsrust: I think it would be cleaner to use the new LocaleFormatter class here:
https://github.com/DMPRoadmap/roadmap/blob/development/lib/locale_formatter.rb

```
I18n.default_locale = LocaleFormatter.new(default_locale, format: :i18n)
FastGettext.default_locale = LocaleFormatter.new(default_locale, format: :fast_gettext)
```

briri:
This fixes the German language issue but not the random travis failures for referencing UI items by their text (e.g. `click_button "Save"`). :/

Bodacious:
@briri thanks for updating the `fast_gettext.rb` initializer. 

Agreed that there is still something fishy going on with the localisation in the tests, but I would argue against adding translation to the test files. We want tests to be as direct as possible and avoid any sort of abstraction that might make it less clear what's being tested and what's expected (this is the same reason we don't DRY up tests with meta programming).

I think the need to keep to testing in a given language and spend a bit of time debugging the issue of the locale being reset incorrectly in the tests.

PR #1901, opened 2018-10-01
Add 'new window' text for Screen Readers when links have target="_blank"

briri:
Originally reported on DMPTool accessibility tests. A few of the links though are in the Roadmap code base though.
https://github.com/CDLUC3/dmptool/issues/160

- Add a '(opens in a new window)' message to image alt text for DCC/CDL logos in footer
- Add `<em class="sr-only">(new window)</em>` to other offsite links
- Skipped Oauth links to ORCID and Ship since these technically bring the user back to the original page and the text of the links indicates that the user is going to those entities to login/link their accounts


Bodacious:
Code changes look fine, happy to approve.

For features like this in future, it might be DRYer to create a helper method:

``` ruby
module LinksHelper
  def external_link_to(name, path, options)
    options[:target] = "_blank"
    name = name + content_tag(:em, "(new window)", class: "sr-only")
    link_to(name, path, options)
end
```

...just to make it easier to update in future if needed.

briri:
Yes, good point

PR #1902, opened 2018-10-02
Fix form validations on Orgs form

### LINKED ISSUE #1860: Unable to create new Orgs

**Expected behaviour:**
Super Admin is able to create a new Org

**Actual behaviour:**
User is unable to create an Org because they cannot provide all the necessary info via the form. The Language field is required by the model but there is no language selector on the form.

**Steps to reproduce:**
- Login as a super admin
- Go to Admin -> Organisations
- Click 'New Organisation'
- Fill in required fields
- Submit the form

screen shot 2018-09-05 at 11 44 41 am

We can either default the value to the installation's default language pref or add a selector. This value currently has no bearing on the rest of the system. I think the original intention may have been for it to determine new user's default language pref.

Bodacious:
Fixes #1860  .

Changes proposed in this PR:
- Update .gitignore file
- Add validations for `contact_name` field
- Remove validations for links fields

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -91,3 +91,6 @@ lib/assets/.eslintcache
 .byebug_history
 lib/data_cleanup/rules/org/fix_blank_abbreviation.yml
 .rspec
+node_modules
+.env
+public/packs
--- a/app/models/org.rb
+++ b/app/models/org.rb
@@ -81,9 +81,10 @@ class Org < ActiveRecord::Base
 
   validates :language, presence: { message: PRESENCE_MESSAGE }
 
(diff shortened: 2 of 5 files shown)


briri:
I think we may need to restore some JS validation for the links (used here and on the template details page when the org is a funder).
The links aren't required, but when one is provided it must have both a URL and Text. The URL should be validated to ensure that its a valid URL.

briri:
Lets refactor that afterward (along with the CSS changes you've made). Orgs don't update their links very often. We'll have to add a cleanup task though to capture any unvalidated links that get through between releases.

PR #1911, opened 2018-10-08
The allowed QuestionFormats when adding a new question to a customisa…

johnpinto1:
Fixes # 1771

Changes proposed in this PR:
- Ensure that "Multi select box" question format is no longer visible when customising questions.


briri:
looks good @johnpinto1

Issue #1916, opened 2018-10-09
All hyperlinks should open in a new tab


stephaniesimms:
While testing autosave (related issue #1878 ) on the write plan page I experienced some weird behavior related to hyperlinks. When I click a Guidance link (screenshot 1) it opens in the same tab. Then if I try to navigate back to the write plan page the browser seems to think I'm in a new session and doesn't display any of my previously answered and autosaved questions (screenshot 2). I have to reload the entire plan to recover my answered questions.

The desired behavior is to open links in a new tab anyway. Reinstating this behavior should avoid the problem.

<img width="935" alt="screen shot 2018-10-09 at 12 15 09 pm" src="https://user-images.githubusercontent.com/12703638/46692807-09e41480-cbbd-11e8-8d87-15212504cbf2.png">

<img width="950" alt="screen shot 2018-10-09 at 12 11 55 pm" src="https://user-images.githubusercontent.com/12703638/46692867-2da75a80-cbbd-11e8-8ca2-77426b317609.png">

stephaniesimms:
working as expected now. closing issue.

sjDCC:
Looks good. Thanks @briri

PR #1918, opened 2018-10-09
added JS to force external links to openin a new tab

briri:
We're doing this in the DMPTool already so submitting for inclusion to Roadmap to address #1916.

This JS logic will force any external link (unless its in the exceptions list) to open in a new browser tab. We went through and did this for static links, but this will also catch links embedded in user input data like guidance and questions/answers.


briri:
Yeah. could be. I'll take a look at it today

PR #1919, opened 2018-10-10
Issue 1853

### LINKED ISSUE #1853: Update template customisation to prevent duplicates
When updating a Template via `upgrade_customization!`, there's a risk that duplicate sections and annotations might be added. 

This bug may happen in cases where the Template has customised versions where the number value of sections, phases, and annotations might have changed.

Add a new attribute to these models that is persisted across Template versions and represents the position of the record within the parent.

Bodacious:
Fixes #1853  .

Changes proposed in this PR:
- Fix bug caused by clashing section numbers when transferring custom sections
- Update capybara TinyMCE helper to match other capybara matchers
- Add single point of contact for Template publishing (see Template#publish)
- Add specs for template customisation

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -232,4 +232,7 @@ group :development do
 
   # TomDoc for YARD (http://rubyworks.github.com/yard-tomdoc)
   gem "yard-tomdoc"
+
+  gem "dotenv-rails"
+
 end
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -100,6 +100,10 @@ GEM
       devise (>= 4.0.0)
     diff-lcs (1.3)
     docile (1.3.0)
+    dotenv (2.5.0)
+    dotenv-rails (2.5.0)
+      dotenv (= 2.5.0)
+      railties (>= 3.2, < 6.0)
     dragonfly (1.1.5)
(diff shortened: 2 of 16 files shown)


Gemfile
```diff
@@ -232,4 +232,7 @@ group :development do
 
   # TomDoc for YARD (http://rubyworks.github.com/yard-tomdoc)
   gem "yard-tomdoc"
+
+  gem "dotenv-rails"
```
xsrust: What's this being used for?

Bodacious: Allows for use of ENV variables which many people like to do to keep their code and dev. environment separate. For example, testing each DB_ADAPTER as described in `database_example.yml`

``` YML
defaults: &defaults
  adapter: <%= ENV['DB_ADAPTER'] || 'postgresql' %>
  encoding: <%= ENV['DB_ADAPTER'] == "mysql2" ? "utf8mb4" : "" %>
  username: <%= ENV["DB_ADAPTER"] == "postgresql" ? 'postgres' : '' %>
  database: roadmap_<%= ENV['RAILS_ENV'] %>
  pool: 5

development:
  <<: *defaults

test:
  <<: *defaults

```

PR #1921, opened 2018-10-10
Rewrite version method to return only SHA

Bodacious:
Changes proposed in this PR:
- Instead of version helper showing current tagged version, show SHA (more useful for dev releases)


briri:
seems sensible

PR #1923, opened 2018-10-11
Issues/1844

xsrust:
updated helper function logic to only deal with whether or not a custom section should be export.  Wrapped section and question headings in their own conditional block.


app/models/concerns/exportable_plan.rb
```diff
@@ -51,7 +51,7 @@ def prepare(coversheet = false)
     template = Template.includes(phases: { sections: { questions: :question_format } })
                        .joins(phases: { sections: { questions: :question_format } })
                        .where(id: self.template_id)
-                       .order("sections.number", "questions.number").first
+                       .order('sections.number', 'questions.number').first
```
briri: Rubocop prefers double quotes

PR #1927, opened 2018-10-11
Data cleanup additions

xsrust:
## Validations:
Updates guidance to only need a theme if published (it was difficult to assign a 'generic' theme)

## Removed:
Splash logs model should have been removed with Migration

## SectionSorter: 
Added a mode_name method to the class so it will respond to this and be able to be excluded from the task.

## Model Files:
Added schema annotations to models added in recent migrations

## Data Cleanup
- Updated the search for an Answer's plan to be iterative rather than as a query since this was failing on 2 of our datasets.
- Unpublish guidances without themes
- Remove guidance groups without orgs
- Remove notes without an answer
- Added tasks for user-orgs and name NOTE: have a look at the name defaults to see if you like them
- add a default to title-less templates
- update query on user-identifiers


app/models/guidance.rb
```diff
@@ -54,7 +54,7 @@ class Guidance < ActiveRecord::Base
   validates :published, inclusion: { message: INCLUSION_MESSAGE,
                                      in: BOOLEAN_VALUES}
 
-  validates :themes, presence: { message: PRESENCE_MESSAGE }
+  validates :themes, presence: { message: PRESENCE_MESSAGE }, if: :published?
```
briri: This might cause problems with the weird way we publish Guidance and Groups in the UI. We will have to make sure that UAT is thorough.

lib/data_cleanup/rules/guidance_group/remove_blank_org.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+module DataCleanup
+  module Rules
+    # Fix blank user on Answer
```
briri: comment is inaccurate. I'll clean up today. I introduced a few of my own ;)

lib/data_cleanup/rules/template/fix_blank_org.rb
```diff
@@ -11,7 +11,7 @@ def description
 
         def call
           ::Template.where("customization_of is not null and customization_of not in (?)", ::Template.all.pluck(:family_id)).each do |customization|
-            log("Setting customization_of to NULL for Template without matching family_id: #{ids}")
+            log("Setting customization_of to NULL for Template without matching family_id: #{customization.family_id}")
```
briri: good catch!

lib/data_cleanup/rules/user/fix_blank_name.rb
```diff
@@ -0,0 +1,27 @@
...
+        end
+
+        def call
+          surname = Set.new(::User.where(surname: [nil,""]))
+          firstname = Set.new(::User.where(firstname: [nil,""]))
+          both = firstname & surname
+          surname = surname ^ both
+          firstname = firstname ^ both
+          either = firstname | surname
+          log("Setting users without orgs to other_org: #{either.map(&:id)}")
```
briri: incorrect log entry.

PR #1928, opened 2018-10-12
update guidance spec to allign with changes to validation logic

xsrust:


briri:
good catch on the `includes(plan)`

PR #1929, opened 2018-10-12
Fix bug preventing new sections from merging into persisted phases

### LINKED ISSUE #1853: Update template customisation to prevent duplicates

Bodacious:
Fixes #1853 

Changes proposed in this PR:
- Fixed a bug that prevented new sections on customised template from being merged onto newly created template in `UpgradeCustomizationService` (this was caused by trying to save a Section with Questions. The presence validation on the Question model was preventing this)


briri:
failures in GTravis look like they were fixed in Sam's latest PR

PR #1930, opened 2018-10-12
Added new find_known_invalidations and updated upgrade tasks

briri:
1) Created a new Rake task, `upgrade:add_default_values_v2_0_0`,  that adds all new default values to fields that may no longer be NULL (e.g. `guidance_groups.optional_subset`

2) Replaced the `upgrade:find_known_invalidations` Rake task with one that runs through each model's attributes` validators and runs a corresponding SQL query to check their validity. (Checks the individual instance's `.valid?` for our homegrown validators due to them not providing enough info). This should be more inclusive than the old one and will hopefully remain relevant as changes occur in the future.

3) Broke `update:v2_0_0` Rake tasks into 2 parts. 
Part 1 `update:v2_0_0_part_1` does the following:
- Runs `upgrade:add_default_values_v2_0_0` 
- Runs `db:migrate`
- Runs `upgrade:find_known_invalidations` with a message afterward instructing the user to correct the issue before proceeding any further

Part 2 `update:v2_0_0_part_2` does the following:
- Runs `data_cleanup:clean_invalid_records` which will pick up any defined rules
- Runs `upgrade:add_versioning_id_to_templates` to populate the new versional_id fields
- Runs `upgrade:normalize_language_formats` to correct any ISO formatting issues
- Runs `stat:build` to build out the new Stats tables

### DIFF
--- a/lib/tasks/data_cleanup.rake
+++ b/lib/tasks/data_cleanup.rake
@@ -190,7 +190,13 @@ namespace :data_cleanup do
       filter.attributes.each do |attr|
         # skip password validaton since the field is encrypted through Devise
         unless attr == :password
-          ids = klass.where.not(attr.to_sym => filter.options[:when]).pluck(:id)
+          # If this is the users.email field send it to the EmailValidator. Devise has its own Regex
+          # but running a Regex query gets messy between different DB types
+          if klass.name == "User" && attr == :email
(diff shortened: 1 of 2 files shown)


Bodacious:
+1 for handling data migrations in the migrations files (as well as database structure migrations)

PR #1933, opened 2018-10-15
Remove unnecessary validator from OrgIdentifier model

### LINKED ISSUE #1934: OrgIdentifier model has an unnecessary validation
Discovered issue with OrgIdentifier validation when going through the output of the data_cleanup:find_known_invalidations task. The OrgIdentifier was checking for the unstopped uniqueness of the 'identifier' field. DMPTool has a scenario though where 2 institutions point to the same Shib IdP.

Remove the validation since its not necessary. The uniqueness check for org_id + identifier_scheme_id will prevent multiple entries.

briri:
Fixes #1934 

Discovered issue with OrgIdentifier validation when going through the output of the `data_cleanup:find_known_invalidations` task. The OrgIdentifier was checking for the unstopped uniqueness of the 'identifier' field. DMPTool has a scenario though where 2 institutions point to the same Shib IdP.

There was already a check for uniqueness against the identifier_scheme so the additional requirement on the identifier was not necessary.

### DIFF
--- a/app/models/org_identifier.rb
+++ b/app/models/org_identifier.rb
@@ -35,7 +35,7 @@ class OrgIdentifier < ActiveRecord::Base
                                                  message: UNIQUENESS_MESSAGE }
 
   validates :identifier, presence: { message: PRESENCE_MESSAGE },
-                         uniqueness: { message: UNIQUENESS_MESSAGE }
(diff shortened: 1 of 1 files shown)


xsrust:
Looks good to me.  Good catch.

PR #1943, opened 2018-10-16
Fix bug when sorting draggable sections

### LINKED ISSUE #1942: Able to drag 2 custom sections to top of funder template
Currently I'm able to drag two custom sections to the top of a funder template, if I drag the second to second position. If it put it first the error message is raised as expected.

Previously the tool prevented two from sitting at the top. If you tried to drag a second there it was returned to its original location. In the screengrab below you can see the two sections. If I go to preview I see the two, but when I return, only 1 is visible on the editable interface. The other disappears.

able-to-drag-2

Thoughts @Bodacious?

Bodacious:
Fixes #1942  .

Changes proposed in this PR:
- Fix bug on template dragging and sorting.

### DIFF
--- a/lib/assets/javascripts/views/org_admin/phases/show.js
+++ b/lib/assets/javascripts/views/org_admin/phases/show.js
@@ -8,9 +8,6 @@ $(() => {
   //
   // Returns Boolean
   function prefixSectionExists(draggableSections) {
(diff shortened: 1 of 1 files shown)


Bodacious:
@xsrust Just a heads-up I've pushed to this branch since you approved it.

briri:
looks good to me

PR #1950, opened 2018-10-16
Added limitations to 'Other' org user's access

### LINKED ISSUE #1935: Guidance default configurations lost
This may just be a configuration issue, but when I create a new plan the guidance behavior is incorrect. 
- I should see the "DMPTool"/UCOP guidance group checked by default and my organizational affiliation guidance also checked. Currently I only get my org guidance if there is any by default (screenshot 1 - )
- I should not see "Non partner institution" in the full list of guidance (screenshot 2)

screen shot 2018-10-15 at 3 24 02 pm

screen shot 2018-10-15 at 3 00 14 pm

briri:
Fixes #1935.

Changes proposed in this PR:
- Added additional where clause to `published` scope to exclude the `orgs.is_other` Org. Technically that org does not have admins and should never have guidance or templates. There are situations though where a Super Admin can switch to the 'other' org and still have access to the guidance page and is then able to create guidance. 
- Added additional checks to prevent users associated with the 'Other' Org from accessing the Template, Guidance and User Permissions endpoints. This should prevent the original issue from occurring in the first place.

### DIFF
--- a/app/models/guidance_group.rb
+++ b/app/models/guidance_group.rb
@@ -65,7 +65,7 @@ class GuidanceGroup < ActiveRecord::Base
     where("name LIKE ?", search_pattern)
   }
 
-  scope :published, -> { where(published: true) }
+  scope :published, -> { where(published: true).joins(:org).where('orgs.is_other = ?', false) }
 
   # =================
   # = Class methods =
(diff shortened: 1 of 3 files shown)


Bodacious:
@xsrust would you mind reviewing this PR? 

I'm not familiar enough with the requirements to assess the implementation.

briri:
just an FYI ... I need to update the Spec tests on this still

briri:
closing in favor of #1953

PR #1953, opened 2018-10-17
Alternate implimentation of 1935

xsrust:
Taclking 1935 from the 'prevent data we dont want' angle rather than adding logic to filter out this unwanted data.  I think this approach will be easier to maintain in the long run.

- `publishable?` method added to template and guidance_group models
- `can_publish` method added in org model
- Conditional validations on the published field added to template and guidance to enforce false unless publishable.
- Change template publish controller action to be in line with the error reporting across the rest of the site.
- Push down controller logic on publishability of templates into model.
- Small improvement to unpublish query performance

### DIFF
--- a/app/controllers/org_admin/templates_controller.rb
+++ b/app/controllers/org_admin/templates_controller.rb
@@ -250,18 +250,14 @@ module OrgAdmin
     def publish
       template = Template.find(params[:id])
       authorize template
-      # rubocop:disable Metrics/LineLength
-      if template.latest?
-        # Now make the current version published
-        if template.publish!
-          flash[:notice] = _("Your #{template_type(template)} has been published and is now available to users.")
-        else
-          flash[:alert] = _("Unable to publish your #{template_type(template)}.")
-        end
+      template.published = true
+      if template.save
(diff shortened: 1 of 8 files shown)


app/controllers/org_admin/templates_controller.rb
```diff
@@ -250,26 +250,22 @@ def history
...
-      # rubocop:enable Metrics/LineLength
       redirect_to request.referrer.present? ? request.referrer : org_admin_templates_path
     end
 
     # PATCH /org_admin/templates/:id/unpublish  (AJAX)
     def unpublish
       template = Template.find(params[:id])
       authorize template
-      versions = Template.where(family_id: template.family_id)
+      versions = Template.where(family_id: template.family_id, published: true)
```
briri: This feels like its better suited to life in the model too since we have `publish` methods there.

briri:
Feel free to merge once those changes are in place @xsrust

briri:
Lets push this to the next release (with the CSS rework). Its an edge case and I don't think another round of UAT warrants putting this release off any further.

PR #1954, opened 2018-10-17
Added check to ensure that the model stems from ActiveRecord::Base in data_cleanup

briri:
Discovered bug while running `data_cleanup.find_invalid_records` that was throwing an error for the SectionSorter object because it does not inherit from ActiveRecord::Base.

Added a simple check to make sure the model is an ActiveRecord object before checking its validity.


briri:
closing since it was addressed in #1955

PR #1966, opened 2018-10-23
updated auth logic on public plans

### LINKED ISSUE #1965: Public plan permission error
The PDF download links on public plans are not working currently. First you are prompted to login and then you get a 'not-authorised' message, unless you are an super-admin (or org-admin for the uni in question I assume).

Everyone should be able to download public plans without login

xsrust:
Fixes #1965.


Bodacious:
For the record, the DCC-User approval was me. I was signed in to the wrong account accidentally 🙈

PR #1969, opened 2018-10-24
Add ActiveRecord Session storage

### LINKED ISSUE #1968: Prefer database for session storage
As [described in the Rails documentation](https://guides.rubyonrails.org/security.html#session-expiry), it's safer for us to use server-side session storage than browser-side:

> One possibility is to set the expiry time-stamp of the cookie with the session ID. However the client can edit cookies that are stored in the web browser so expiring sessions on the server is safer. Here is an example of how to expire sessions in a database table. Call Session.sweep("20 minutes") to expire sessions that were used longer than 20 minutes ago.

Bodacious:
Fixes #1968

Changes proposed in this PR:
- Add ActiveRecord Session Storage for managing sessions instead.

Developers should add to their own deployment, the following cron job:

`SESSION_DAYS_TRIM_THRESHOLD=2  RAILS_ENV=production rake db:sessions:trim`

This will help to keep the sessions table small, and automatically sign out all users after 48 hours (2 days)

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20180903131335) do
+ActiveRecord::Schema.define(version: 20181024120747) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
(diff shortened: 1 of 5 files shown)


db/schema.rb
```diff
@@ -300,6 +300,16 @@
...
   add_index "sections", ["versionable_id"], name: "index_sections_on_versionable_id", using: :btree
 
+  create_table "sessions", force: :cascade do |t|
+    t.string   "session_id", limit: 255, null: false
+    t.text     "data"
+    t.datetime "created_at"
+    t.datetime "updated_at"
+  end
+
+  add_index "sessions", ["session_id"], name: "index_sessions_on_session_id", unique: true, using: :btree
```
briri: This index is having issues in MySQL:
```
ActiveRecord::StatementInvalid: Mysql2::Error: Specified key was too long; max key length is 767 bytes: CREATE UNIQUE INDEX `index_sessions_on_session_id` USING btree ON `sessions` (`session_id`)
```

briri:
Thanks for updating the col size. Feel free to merge once the tests are passing

PR #1970, opened 2018-10-24
Add Rollbar gem for catching production errors

Bodacious:
Add support for Rollbar gem to catch exceptions from the app.

To include Rollbar in your setup, use the `--with` option when running `bundle install`

`bundle install --with=rollbar`

You'll also need to provide an ENV variable with your rollbar access token:

```
ROLLBAR_ACCESS_TOKEN=73ef49021e064a23b65d637b18997912
```


Bodacious:
Good shout! thanks for the link.

PR #1987, opened 2018-11-11
Issue1679

jollopre:
- CreateCreatedPlanService and CreateJoinedUserService are now either creating or updating stats for a month
- Stat model is now extended with details JSON field
- CreateCreatedPlanService is now extended with plans by template
- Usage stats has now a horizontal bar chart for plans by template
- Usage stats has now a button to download a CSV with plans by template since org's creation

Below an example of the new horizontal bar chart for plans by template. The interface allows you to select range dates (last month, 3 months ago, etc). Note that you will need to run ```rake stat:create:created_plan ``` in order to populate stats for including counts of plans by template

screen shot 2018-11-11 at 22 03 41

### DIFF
--- a/app/controllers/stat_created_plans_by_template_controller.rb
+++ b/app/controllers/stat_created_plans_by_template_controller.rb
@@ -1,0 +1,35 @@
+# frozen_string_literal: true
+
+class StatCreatedPlansByTemplateController < ApplicationController
+
+  def index
+    check_authorized!
+
+    data = StatCreatedPlan.monthly_range(index_filter).order(date: :desc)
+
+    if params[:format] == "csv"
+      data_csvified = StatCreatedPlan.to_csv(data, details: { by_template: true })
+      send_data(data_csvified, filename: "created_plan_by_template.csv")
+    else
+      render(json: data.as_json(only: [:date, :count], methods: :by_template))
+    end
+  end
+
+  private
+
+  def index_filter
+    {
+      org: current_user.org,
+      start_date: params[:start_date],
+      end_date: params[:end_date]
+    }
+  end
+
+  def check_authorized!
+    unless current_user.present? &&
+        current_user.can_org_admin?
+      raise Pundit::NotAuthorizedError
+    end
+  end
+
+end
--- a/app/services/org/create_created_plan_service.rb
+++ b/app/services/org/create_created_plan_service.rb
@@ -1,30 +1,27 @@
+# frozen_string_literal: true
+
 class Org
+
   class CreateCreatedPlanService
+
     class << self
+
       def call(org = nil)
         orgs = org.nil? ? Org.all : [org]
 
         orgs.each do |org|
           OrgDateRangeable.split_months_from_creation(org) do |start_date, end_date|
-            create_count_for_date(start_date: start_date, end_date: end_date, org: org)  
+            StatCreatedPlan::CreateOrUpdate.do(
+              start_date: start_date,
+              end_date: end_date,
+              org: org
+            )
           end
         end
       end
 
-      private
-
-      def count_plans(start_date: , end_date: , org:)
-        users = User.where('users.org_id = ?', org.id)
-        plans = Plan.where('plans.created_at >= ? AND plans.created_at <= ?', start_date, end_date)
-        creator_admon = Role.with_access_flags(:creator, :administrator)
-
-        Role.joins([:plan, :user]).merge(creator_admon).merge(users).merge(plans).select(:plan_id).distinct.count
-      end
-
-      def create_count_for_date(start_date:, end_date:, org:)
-        count = count_plans(start_date: start_date, end_date: end_date, org: org)
-        StatCreatedPlan.create(date: end_date.to_date, count: count, org_id: org.id)
-      end
     end
+
   end
+
 end
--- a/app/services/org/create_last_month_created_plan_service.rb
+++ b/app/services/org/create_last_month_created_plan_service.rb
@@ -1,32 +1,29 @@
+# frozen_string_literal: true
+
 class Org
+
   class CreateLastMonthCreatedPlanService
+
     class << self
+
       def call(org = nil)
-        orgs = org.nil? ? ::Org.all : [org]
+        orgs = org.nil? ? Org.all : [org]
 
         orgs.each do |org|
           months = OrgDateRangeable.split_months_from_creation(org)
           last = months.last
           if last.present?
-            create_count_for_date(start_date: last[:start_date], end_date: last[:end_date], org: org)
+            StatCreatedPlan::CreateOrUpdate.do(
+              start_date: last[:start_date],
+              end_date: last[:end_date],
+              org: org
+            )
           end
         end
       end
 
-      private
-
-      def count_plans(start_date: , end_date: , org:)
-        users = User.where('users.org_id = ?', org.id)
-        plans = Plan.where('plans.created_at >= ? AND plans.created_at <= ?', start_date, end_date)
-        creator_admon = Role.with_access_flags(:creator, :administrator)
-
-        Role.joins([:plan, :user]).merge(creator_admon).merge(users).merge(plans).select(:plan_id).distinct.count
-      end
-
-      def create_count_for_date(start_date:, end_date:, org:)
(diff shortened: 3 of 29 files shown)


app/services/org/create_created_plan_service.rb
```diff
@@ -6,24 +6,54 @@ def call(org = nil)
...
-            create_count_for_date(start_date: start_date, end_date: end_date, org: org)  
+            create_or_update(start_date: start_date, end_date: end_date, org: org)
           end
         end
       end
 
       private
 
+      def users(org)
+        User.where('users.org_id = ?', org.id)
```
briri: prefer to see the use of the symbol here: `User.where(org: org)`

app/services/org/create_created_plan_service.rb
```diff
@@ -6,24 +6,54 @@ def call(org = nil)
...
+      end
+
+      def plans(start_date:, end_date: )
+        Plan.where('plans.created_at >= ? AND plans.created_at <= ?', start_date, end_date)
```
briri: prefer symbols here: `Plan.where(created_at: start_date..end_date)`. You may need to cast those values to Date.

app/services/org/create_created_plan_service.rb
```diff
@@ -6,24 +6,54 @@ def call(org = nil)
...
+      end
+
+      def creator_admon
```
briri: Think this should be called `creator_admin` :)

app/services/org/create_created_plan_service.rb
```diff
@@ -6,24 +6,54 @@ def call(org = nil)
...
+        Role.with_access_flags(:creator, :administrator)
+      end
+
       def count_plans(start_date: , end_date: , org:)
-        users = User.where('users.org_id = ?', org.id)
-        plans = Plan.where('plans.created_at >= ? AND plans.created_at <= ?', start_date, end_date)
-        creator_admon = Role.with_access_flags(:creator, :administrator)
+        users = users(org)
+        plans = plans(start_date: start_date, end_date: end_date)
```
briri: why set the results of the functions to variables of the same name here? Also prefer to use pluck here since you don't need the overhead of Rails creating the model. Do something like this:
```ruby
Role.joins([:plan, :user)
       .merge(creator_admin)
      .merge(users(org))
      .merge(plans(start_date: start_date, end_date: end_date))
      .pluck(:plan_id)
      .distinct.count
```

app/services/org/create_created_plan_service.rb
```diff
@@ -6,24 +6,54 @@ def call(org = nil)
...
 
         Role.joins([:plan, :user]).merge(creator_admon).merge(users).merge(plans).select(:plan_id).distinct.count
       end
 
-      def create_count_for_date(start_date:, end_date:, org:)
+      def by_template(start_date:, end_date:, org:)
+        users = users(org)
+        plans = plans(start_date: start_date, end_date: end_date)
+        roleable_plan_ids = Role.joins([:plan, :user]).merge(creator_admon).merge(users).merge(plans).select(:plan_id).distinct
+        template_counts = Plan.where(id: roleable_plan_ids).group(:"template_id").count
```
briri: template_id doesn't need to be in quotes

app/services/org/create_created_plan_service.rb
```diff
@@ -6,24 +6,54 @@ def call(org = nil)
...
+        plans = plans(start_date: start_date, end_date: end_date)
 
         Role.joins([:plan, :user]).merge(creator_admon).merge(users).merge(plans).select(:plan_id).distinct.count
       end
 
-      def create_count_for_date(start_date:, end_date:, org:)
+      def by_template(start_date:, end_date:, org:)
+        users = users(org)
+        plans = plans(start_date: start_date, end_date: end_date)
+        roleable_plan_ids = Role.joins([:plan, :user]).merge(creator_admon).merge(users).merge(plans).select(:plan_id).distinct
```
briri: see comments above in `count_plans`

app/services/org/create_last_month_created_plan_service.rb
```diff
@@ -2,30 +2,60 @@ class Org
...
       end
 
       private
 
+      def users(org)
+        User.where('users.org_id = ?', org.id)
+      end
+
+      def plans(start_date:, end_date:)
+        Plan.where('plans.created_at >= ? AND plans.created_at <= ?', start_date, end_date)
```
briri: See comments in `create_created_plan_service`. I think there is a more DRY solution. This code looks identical.

app/services/org/create_last_month_joined_user_service.rb
```diff
@@ -18,9 +18,17 @@ def count_users(start_date: , end_date: , org_id: )
         User.where('created_at >= ? AND created_at <= ? AND org_id = ?', start_date, end_date, org_id).count
```
briri: Use `User.where(created_at: start_date..end_date, org: org)`

app/services/org/create_created_plan_service.rb
```diff
@@ -6,24 +6,54 @@ def call(org = nil)
...
+      end
+
+      def create_or_update(start_date:, end_date:, org:)
         count = count_plans(start_date: start_date, end_date: end_date, org: org)
-        StatCreatedPlan.create(date: end_date.to_date, count: count, org_id: org.id)
+        by_template = by_template(start_date: start_date, end_date: end_date, org: org)
+        attrs = { date: end_date.to_date, org_id: org.id, count: count, details: { by_template: by_template }}
+        stat_created_plan = StatCreatedPlan.find_by(date: attrs[:date], org_id: attrs[:org_id])
+
+        if stat_created_plan.present?
```
briri: ActiveRecord has a `create_or_update` method you can use instead

app/services/org/create_created_plan_service.rb
jollopre: me too, but with the way you propose ActiveRecord is not smart enough to determine to what table that attribute refers when used through 'merge' spawn method. You have to namespace it, i.e. ```{ users: { org_id: org_id }}```

app/services/org/create_created_plan_service.rb
jollopre: good one! Forgot that use of range here translates to a sql 'between' and it is more expressive

app/services/org/create_created_plan_service.rb
jollopre: I guess so

app/services/org/create_created_plan_service.rb
jollopre: Setting to variables improves the readability and using either (select or pluck) will not change much since active record queries are lazy evaluated and at the end just the count is returned, i.e. no models are instantiated.

app/services/org/create_created_plan_service.rb
jollopre: ops! you're right it is pointless to symbolize that way

app/services/org/create_created_plan_service.rb
jollopre: ```create_or_update```, as far as I'm aware is a private method used by #save. Since save is an instance method, it means that we have to find a matching record by date and org_id in order to either create or save the instance therefore that's the reason I used it. Anyway, I have added a uniqueness validation against date/org_id for a stat type (e.g. StatCreatedPlan, StatJoinedUser) in case another piece of code tries to save to the database outside this service. I'm open to other proposals :)

app/controllers/stat_created_plans_by_template_controller.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+class StatCreatedPlansByTemplateController < ApplicationController
+
+  def index
+    check_authorized!
+
+    data = StatCreatedPlan.monthly_range(index_filter).order('date DESC')
+
+    if params[:format] == 'csv'
```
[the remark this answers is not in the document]
jollopre: I'm I'm not wrong, the strong params is used for mass assignment and this piece of code is neither creating or updating any record from the db, that's the reason is not 'strongerised'.

jollopre:
@briri, the suggested changes are now in place with another commit

PR #1988, opened 2018-11-12
Fixes bug where users could not accept invitations to plans

xsrust:
- Updated devise's invitation to set a default first/last name, and to default the org to the inviter's org.
- Updated form users are driven to after following invitation link to ask users to fill in name and org


xsrust:
will close and PR development

Issue #1161, opened 2018-02-14
Display tooltip below form fields on Create plan page


stephaniesimms:
USABILITY TESTING FEEDBACK
Only saw tip after he typed… didn’t realize it was for Project Title so he goes back and amends it.
Didn't notice mock project checkbox because it was hidden by tip.
- on create plan page display tooltips below the field instead of out to the right (screenshot)

![screen shot 2018-07-09 at 11 37 55 am](https://user-images.githubusercontent.com/12703638/42469493-c1a1b0f8-836c-11e8-8dce-13db04ecb1ec.png)

briri:
mainly an issue on the create plan page. Display below the field if possible

johnpinto1:
@briri  @xsrust This code is in current development. So I think it should be available for UAT.

sjDCC:
Looks good, thanks @johnpinto1 Closing out

PR #1989, opened 2018-11-13
Issue #1161 - Display tooltip below form fields on Create plan page.

johnpinto1:
- Required modification for the intiialization of Bootstrap tooltips in the javascript file
      /utils/tooltipHelper.js. The default behaviour is placement of tooltip to the right of element
      with attribute data-toggle='tooltip'. Now if the attribute data-placement attribute is set,
      then the placement of the tootip will be as specfied in the attribute and not on the right (the default behaviour).
    - The accessible combob-box now takes an optional property 'placement' which if present sets data-placement
      attribute to the value of the placement property.
    - Now all Create Plan page tooltips are below the input as requested.

Fix for #1161.

### DIFF
--- a/app/javascript/utils/tooltipHelper.js
+++ b/app/javascript/utils/tooltipHelper.js
@@ -6,8 +6,14 @@ $(() => {
   // uses an iframe and we can't detect when the editor window gains focus. It only works on hover.
   //
   // If the content of the tooltip contains HTML, then add `data-html="true"` to the element
-  $('[data-toggle="tooltip"]').tooltip({
+  // Default behaviour for all tootips, when attribute data-placement not set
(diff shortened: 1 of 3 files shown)


app/javascript/utils/tooltipHelper.js
```diff
@@ -6,8 +6,15 @@ $(() => {
...
+  $('[data-toggle="tooltip"]:not([data-placement])').tooltip({
     animated: 'fade',
     placement: 'right',
   });
+
+  // Don't set placement property if data-placement present
+  $('[data-toggle="tooltip"][data-placement]').tooltip({
+    animated: 'fade',
+  });
+
```
briri: The linter is expecting the file to have a blank line at the end.

johnpinto1:
Merging PR as suggested by@briri

PR #1994, opened 2018-11-25
Issue1391

jollopre:
- Fixes selector for exporting
- Additionally fixes seeds.rb since it was broken due to the option_based validations, i.e. now it has to be saved atomically.


xsrust:
Travis/pr has passed and the remote is on digital_curation_centre, but travis is running the push against the branch on roadmap(which dosent exist), so the git clone is failing.

PR #1995, opened 2018-11-26
Updated api stats controller to use Org model's  method

### LINKED ISSUE #1703: Usage dashboard reports incorrect total plans
We noticed a discrepancy between total n plans as reported in the usage dashboard using the filters and total n plans when querying the dmptool database directly for many orgs, especially those with a lot of recent use. 

One possible contributor is the related bug issue #1663 . It looks like another contributor is that a single plan with multiple contributors/roles (owner, co-owner, editor, etc) is counted multiple times, one for each role. The correct calculation would count individual plan IDs per Org - so if plan A has an owner from UCOP and two co-owners from UCSF it should be counted once for UCOP and once for UCSF.

@sjDCC do we want to include test plans in total n? I'm okay either way but at least the total n on the dashboard should reflect the total n listed in the plans table. if we exclude test plans in the usage db we should just state that somewhere.

briri:
Fixes #1703 .

The usage statistics page was calling the API controller which had not been updated to use the Org model's `plans` method to calculate the total number of plans. Switched to use the `current_user.org.plans` method so that we are not duplicating query logic

### DIFF
--- a/app/controllers/api/v0/statistics_controller.rb
+++ b/app/controllers/api/v0/statistics_controller.rb
@@ -119,12 +119,21 @@ class Api::V0::StatisticsController < Api::V0::BaseController
     if params[:range_dates].present?
       r = {}
       params[:range_dates].each_pair do |k, v|
-        range_date_plans = plans
-          .where("plans.created_at >= ?", v["start_date"])
-          .where("plans.created_at <= ?", v["end_date"])
-        r[k] = roles.joins(:user, :plan).merge(users).merge(range_date_plans)
-                    .select(:plan_id).distinct.count
+        #range_date_plans = plans
+        #  .where("plans.created_at >= ?", v["start_date"])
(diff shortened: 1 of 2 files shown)


app/controllers/api/v0/statistics_controller.rb
```diff
@@ -232,54 +176,18 @@ def plans
...
-        filtered += [obj]
-      end
-    end
-    filtered
+  # Convert start/end dates in hash to a range of Dates
+  def dates_to_range(hash)
+    today = Date.today
+    start_date = Date.parse(hash.fetch("start_date", today.prev_month.to_date.to_s))
+    end_date = Date.parse(hash.fetch("end_date", today.to_date.to_s)) + 1.day
+    start_date..end_date
```
xsrust: well that's way cleaner than my 30 lines from a couple years ago to restrict the range:)

PR #1997, opened 2018-11-27
Moved deactivate logic to Role model and added specs

### LINKED ISSUE #1663: Remove plan action should take DMP out or org view and statistics too
When a user clicks to 'remove plan' it should effectively work like a delete action and remove from all views, unless it is in active user by another user. Currently it just removes from the 'my dashboard' view it seems. The institutional plans view under org admin still shows removed plans, even if not shared / in use by another.

Can you please adjust the logic so 'remove plan' takes them out of statistics and org view too please. It's meant to act like a delete button, though it doesn't get completely wiped from the database to avoid data integrity issues.

briri:
Fixes #1663 .

- Moved deactivate logic out of the controller and into the Role model
- Updated deactivate logic


app/models/role.rb
```diff
@@ -155,6 +155,24 @@ def set_defaults
...
+    self.active = false
+    if self.save!
+      # If no other creator or administrator is attached to the plan
+      # the deactivate all the other roles
+      if Role.where(plan: self.plan, active: true)
+             .with_access_flags(:creator, :administrator).size <= 0
+        [:editor, :commenter, :reviewer].each do |access|
+          Role.where(plan: self.plan, active: true)
+              .with_access_flags(access).update_all(active: false)
+        end
```
xsrust: I think that the .with_access_flags() function is duplicating functionality built into flag_shih_tzu.
https://github.com/pboling/flag_shih_tzu#generated-named-scopes

Also the way we have defined the access fields, all creators will be admins, so we shouldnt need to check for both.

```suggestion
        if Role.where(plan: self.plan, active: true).administrator.size == 0
          Role.where(plan: self.plan, active: true).update_all(active: true)
        end
```

briri:
Updated the Role.deactivate logic to use the FlagShihTzu helper methods and made it so that we only deactivate all users if there is no creator, admin or editor left on the plan. Also added the plan.visibility change back in. 

@xsrust take a look and let me know what you think. I haven't had a chance to update the other controllers yet to adhere to the new logic.

briri:
Further refactoring of Role so that we're using FlagShihTzu more closely and have consolidated logic to the Plan and Role models.

Will finish up testing locally and updates to dependent views/controllers tomorrow

briri:
OK @xsrust hopefully this is it for the refactor.

Note that I had to run a bundle update due to new Ruby vulnerabilities with ActiveJob. I also ran yarn update for good measure.

There is also a tweak in the `download_plans` function to remove periods and commas from the plan title so that they don't make their way into file names (did this in DMPTool code a while back but seems like it would be useful for everyone).

Also note that I updated the `plan.owner` model method so that if it goes by the 'administrator' flag instead of 'creator' in the event that the create has 'removed' the plan. That way it will return the next administrator. I don't see where this would become a problem but may be something to discuss on our Tuesday call

There is now a new `rake data_cleanup: deactivate_orphaned_plans` task that will need to be run that deactivates any plans that already have no owners/coowners

app/controllers/org_admin/plans_controller.rb
```diff
@@ -54,6 +52,7 @@ def download_plans
 
     org = current_user.org
     file_name = org.name.gsub(/ /, "_")
+                        .gsub(/[\.;,]/, "")
```
xsrust: orgs with wierd chars in their names I take it?

app/controllers/roles_controller.rb
```diff
@@ -11,10 +11,10 @@ def create
     @role = Role.new(role_params)
     authorize @role
 
-    access_level = params[:role][:access_level].to_i
-    @role.access_level = access_level
+    plan = Plan.find(role_params[:plan_id])
+
     message = ""
-    if params[:user].present?
+    if params[:user].present? && plan.present?
```
xsrust: I would think the validations we added would remove the need for checks like this.
Were you running into issues without the check?

app/controllers/roles_controller.rb
```diff
@@ -40,7 +40,7 @@ def create
             email: user.email
           }
           @role.user = user
-          if @role.save
+          if @role.save!
```
xsrust: I think you do want `save` instead of `save!`.  The former returns true/false while the latter raises an exception

app/models/plan.rb
```diff
@@ -391,8 +382,8 @@ def commentable_by?(user_id)
   #
   # Returns Boolean
   def administerable_by?(user_id)
-    user_id = user_id.id if user_id.is_a?(User)
-    role?(user_id, :administrator)
+    role = Role.find_by(plan_id: id, user_id: user_id, active: true)
+    (role.present? ? role.administrator? : false)
```
xsrust: I think we could replace this style query with:
```
Role.administrator.where(plan_id: id, user_id: user_id, active: true),any?
```
In this and the surrounding methods

app/models/plan.rb
```diff
@@ -422,18 +403,54 @@ def latest_update
...
   # Returns nil
   def owner
-    vals = Role.access_values_for(:creator)
-    User.joins(:roles)
-        .where("roles.plan_id = ? AND roles.access IN (?)", id, vals).first
+    usr_ids = Role.where(plan_id: id, active: true)
+                  .administrator
+                  .order(:created_at)
+                  .pluck(:user_id).uniq
+    User.where(id: usr_ids).first
```
xsrust: This seems like a really sensible way to define the owner

app/models/plan.rb
```diff
@@ -422,18 +403,54 @@ def latest_update
...
+        role.editor = true
+      when :administrator
+        role.administrator = true
+        role.editor = true
+      when :editor
+        role.editor = true
+      when :reviewer
+        role.reviewer = true
+      end
+      role.commenter = true
```
xsrust: If we dont have any roles which cannot comment, do we need that designation anymore?
I suppose we might always add something like an 'observer' in which has read-only no-comment access. 
Similar to the super/org-admin view into a plan without review status, but with this access for some reason scoped to a subset of all plans.

app/controllers/org_admin/plans_controller.rb
briri: yep. E.g. 'University of California, Office of the President' and 'Univ. of So and So'

app/controllers/roles_controller.rb
briri: good catch

app/models/plan.rb
briri: yes, that's cleaner. will change.

app/models/plan.rb
briri: Yeah. Thought about the same as I was refactoring `reviewable_by?` and `commendable_by?` aside from the super/org admin roles and the Branding config stuff you could get away with `Role.where(plan_id: id, user_id: user_id, active: true).any?`.

As you point out though it gives us an opportunity to add further granularity in the future. I don't think the addition of the access level check adds much overhead right now

PR #2019, opened 2018-12-11
updated call to Plan.readable_by?

### LINKED ISSUE #2018: Not calling Plan.readable_by correctly
We're receiving the following deprecation warning: 
```shell
DEPRECATION WARNING: You are passing an instance of ActiveRecord::Base to `find`. Please pass the id of the object by calling `.id`. (called from readable_by? at /Users/briley/Documents/workspace/dmptool/app/models/plan.rb:351)
```

Make sure all calls to `plan.readable_by` are sending the user id instead of the User object

briri:
Fixes #2018

### DIFF
--- a/app/views/paginable/plans/_org_admin.html.erb
+++ b/app/views/paginable/plans/_org_admin.html.erb
@@ -14,7 +14,7 @@
       <% scope.each do |plan| %>
         <tr>
           <td>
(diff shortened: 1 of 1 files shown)


sjDCC:
I guess this is one for devs to UAT and close

briri:
yes, not sure how the PR ended up on the board. We typically just place issues out there

Issue #1683, opened 2018-07-06
Display issue (cropped answers & whitespace) when reviewing DMPs


sjDCC:
A few admins have experienced issues when reviewing DMPs. The answers are in read-only mode and sometimes get cropped to only show half of someone's answer (you need to reload page to open in full) or show whitespace afterwards. This has happened on Chrome for all users. 

Email text from enquirer below:

The first screen shot image attached shows that the answer has been cut off halfway through (I've just created a test template to avoid sharing our actual plans), whereas the second screenshot shows the full answer after refreshing the page four times. We get this a lot when reviewing DMPs and sometimes we don't always realise that an answer has been cut if it happens to show a full sentence.

![answer - cut off](https://user-images.githubusercontent.com/6169940/42382876-797eb6ea-812d-11e8-8f46-6686ef7a3f64.JPG)
After refreshing page:
![answer - full](https://user-images.githubusercontent.com/6169940/42382880-7c1a02ba-812d-11e8-9998-734be7e5989c.JPG)

Other answers end up with a load of blank space after them which means a lot of scrolling before we get to the next question (shown in third screenshot - you can see the end of an answer at the top of the page with a load of blank space afterwards.)
![answer - blank space](https://user-images.githubusercontent.com/6169940/42382896-8671b83e-812d-11e8-8a50-a21ddab8ed42.JPG)

@xsrust and I have looked into this but can't replicate it, but since multiple people have experienced it it warrants further troubleshooting though to see if we can identify what's happening

Further update from Hannah:
"I tried using Firefox and Explorer as well as Chrome and had the same issues with the answer boxes being cut off or being far too long."

sjDCC:
This issue has been reported again so it still causing some users trouble. User says:

"when I click inside the answer box, it fails to expand, I’ve had to select all text and copy and paste into another system to read it (that sentence continues and there’s another one after).
 
I’m working in Firefox 61.0.1 (64-bit)."

@Bodacious any chance you could see if you could replicate this one? We couldn't isolate the issue to pursue a fix before

stephaniesimms:
Here's some more info from a DMPTool user w/the same issue
![screenshot](https://user-images.githubusercontent.com/37444768/44151461-a7ff5918-a070-11e8-987e-819c12a7da5b.jpg)

[...]

PR #2023, opened 2018-12-14
Display issue (cropped answers & whitespace) when reviewing DMPs when…

johnpinto1:
… answers in

readonly mode.

Fixes issue #1683.

When the review mode is readonly, then we display the answer in in a paragraphs.

(Could not replicate whitespace issue to test if this fixes it too. But will continue to look at it.)
Before change 
selection_097
After change
selection_096


briri:
looks good to me too. thanks @johnpinto1

Issue #1749, opened 2018-07-18
Revise error message for failed attempt to change email address


stephaniesimms:
related bug fix #1714 
When a user tries to change their email address to one that is already registered they receive a vague error message (screenshot below). Revise the error text as follows:
- Error: Save unsuccessful. That email address is already registered. You must enter a unique email address.

![capture](https://user-images.githubusercontent.com/6169940/42806193-bdea8f4a-89a5-11e8-832f-8b982b606e88.PNG)

The problem may be that the email I was changing it to was already known by the system though.

sjDCC:
@xsrust @briri can you confirm whether this code has been deployed. I still get the old 'Save unsuccessful' text on DMPonline-test and roadmap-stage

sjDCC:
All working as expected now. Closing out.

PR #2051, opened 2019-01-31
Added a customised error message for the case where personal details …

johnpinto1:
…update of the email fails because of its use already.

Fix for #1749.

-

### DIFF
--- a/app/controllers/registrations_controller.rb
+++ b/app/controllers/registrations_controller.rb
@@ -189,6 +189,9 @@ class RegistrationsController < Devise::RegistrationsController
             successfully_updated = false
           else
             successfully_updated = current_user.update_with_password(password_update)
+            if !successfully_updated
+              message = _("Save unsuccessful. That email address is already registered. You must enter a unique email address.")
+            end
           end
         else
           # This case is never reached since this method when called with


app/controllers/registrations_controller.rb
```diff
@@ -189,6 +189,9 @@ def do_update(require_password = true, confirm = false)
             successfully_updated = false
           else
             successfully_updated = current_user.update_with_password(password_update)
+            if !successfully_updated
+              message = _("Save unsuccessful. That email address is already registered. You must enter a unique email address.")
```
xsrust: Is it necissarily the case that a user for whom update_with_password fails has used a duplicate email address?
I could see this triggering on an incorrect password as well

briri: Hmm. We should look at refactoring this whole controller at some point. Its overly complex and doesn't need to be IMO. That's for another day though! ;)

This code executes when the user edits their profile. Specifically when they've changed their email address. If the password were invalid Devise kicks out to its own handler I believe. So the only other failure point would be a duplicate email address. The UI form itself should be preventing invalid email addresses.

Issue #2047, opened 2019-01-29
Don't present abbreviation code on User Interface


sjDCC:
When plans are submitted for review, the email notification that will be issued is presented in the UI. You can see in the screengrab below that this sometimes includes the abbreviation code for user names etc. 
![feedback-1](https://user-images.githubusercontent.com/6169940/51900659-c7ce2100-23ad-11e9-8582-484019ee759d.PNG)

Once the 'Request feedback' button is clicked, the notification panel at the top presents the text properly (see screengrab 2). Since we know the user etc, can we ensure it always displays as proper text.
![dmp-feedback](https://user-images.githubusercontent.com/6169940/51900668-cc92d500-23ad-11e9-9055-1659b44e8cc6.PNG)

johnpinto1:
The abbreviations should no longer appear.
![selection_119](https://user-images.githubusercontent.com/8876215/52857367-e60e7d80-311e-11e9-8b81-9382a551c81d.png)
![selection_120](https://user-images.githubusercontent.com/8876215/52857395-f45c9980-311e-11e9-88a0-a7649f6ee928.png)

sjDCC:
Looks good, thanks @johnpinto1 Closing out

PR #2055, opened 2019-02-11
The "Request expert feedback" text area on the Share Plan page displayed

johnpinto1:
selection_119
Ruby strings  %{user_name} and %{plan_name} that were not interpolated.

- The strings are now interpolated.
- The function helper function feedback_constant_to_text in FeedbacksHelper
 displays the user name instead of the email.
- also Rubocop errors and warnings in FeedbacksHelper were fixed.

Fix for issue #2047.
selection_120

### DIFF
--- a/app/views/plans/_share_form.html.erb
+++ b/app/views/plans/_share_form.html.erb
@@ -128,7 +128,7 @@
       <h2><%= _('Request expert feedback') %></h2>
       <p><%= _('Click below to give data management staff at your organisation access to read and comment on your plan.') %></p>
       <div class="well well-sm">
-        <%= sanitize current_user.org.feedback_email_msg.to_s %>
+        <%= sanitize current_user.org.feedback_email_msg.to_s % { user_name: current_user.name(false), plan_name: plan.title } %>
       </div>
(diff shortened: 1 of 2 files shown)


app/views/plans/_share_form.html.erb
```diff
@@ -128,7 +128,7 @@
       <h2><%= _('Request expert feedback') %></h2>
       <p><%= _('Click below to give data management staff at your organisation access to read and comment on your plan.') %></p>
       <div class="well well-sm">
-        <%= sanitize current_user.org.feedback_email_msg.to_s %>
+        <%= sanitize current_user.org.feedback_email_msg.to_s.gsub("%{user_name}", current_user.name(false)).gsub("%{plan_name}", plan.title) %>
```
xsrust:
```suggestion
        <%= sanitize current_user.org.feedback_email_msg.to_s % {user_name: current_user.name(false), plan_name: plan.title} %>
```

I think that you could replace the two gsub calls with the above

johnpinto1: @xsrust @briri  Updated with change suggested by @xsrust and forced pushed.

briri: suggested changes were made

Issue #1743, opened 2018-07-18
Request feedback message field should be mandatory


sjDCC:
On the admin interface, users can choose to switch on the option of offering feedback on plans. If so they need to enter a message to display to end users. This field is marked as mandatory but it is possible to save it without an entry, meaning a blank box is shown on the interface.

Ideally, the stock message would pre-populate the box and be available to edit. You can't copy and paste from the tooltip as it only displays on hover-over.

![default-message](https://user-images.githubusercontent.com/6169940/42878344-a890e510-8a84-11e8-9038-329706af7a08.png)

johnpinto1:
It looks like saving an empty feedback message was already fixed since the bug was recorded. So I added the *Sample Message* from the tooltip to the page
![selection_121](https://user-images.githubusercontent.com/8876215/52857247-7ef0c900-311e-11e9-85f3-143d4a5d780f.png)

magdalenadrafiova:
Hiya - this is fixed and looks brilliant - however would it be possible to change the msg to message please? please see image 
![msg-20022019](https://user-images.githubusercontent.com/47178106/53096997-56027680-3518-11e9-919a-5704b9768f31.PNG)
@johnpinto1

briri:
Hi @johnpinto1,

If you want to get Rails to show better name for attributes like the one above, `orgs.email_msg`, you can add a replacement to the I18n locale files in `config/locales/[language].yml` in the active_record section. For example:
```yaml
en-GB:
  activerecord:
    attributes:
      org:
        email_msg: "Email Message"
  ...
```
https://github.com/DMPRoadmap/roadmap/blob/4d9665ef8c76736ac750fe6ed0763da26be2a0f3/config/locales/en-GB.yml#L3

Be sure to update all of the various locales. You can likely use Google translate for that.

johnpinto1:
PR for change requested by @magdalenadrafiova

![Selection_026](https://user-images.githubusercontent.com/8876215/59029804-2bce8d80-8857-11e9-9990-418a1ea62754.png)

sjDCC:
Looks good. Closing out

PR #2056, opened 2019-02-13
Fix for ensuring that the Edit Request Feedback form disallows saving…

johnpinto1:
… in the case where feedback is checked but no

mandatory message provided.

Fix for #1743.

(Note: I have not tried to make the sample message the default message, as doing so resulted in it being saved. @sjDCC asked if this would be possible but did not require it.)

### DIFF
--- a/app/helpers/orgs_helper.rb
+++ b/app/helpers/orgs_helper.rb
@@ -2,18 +2,18 @@
 
 module OrgsHelper
 
-  DEFAULT_EMAIL = '%{organisation_email}'
+  EMAIL_PLACEHOLDER = "[Organisation Contact Email Placeholder]"
 
-  # Tooltip string for Org feedback form.
+  # Sample message for Org feedback form.
   #
   # org - The current Org we're updating feedback form for.
   #
   # Returns String
-  def tooltip_for_org_feedback_form(org)
-    email = org.contact_email.presence || DEFAULT_EMAIL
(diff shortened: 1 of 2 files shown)


app/javascript/views/orgs/admin_edit.js
```diff
@@ -7,21 +7,44 @@ import { eachLinks } from '../../utils/links';
 $(() => {
   const toggleFeedback = () => {
     const editor = Tinymce.findEditorById('org_feedback_email_msg');
+    const saveBtn = $('#save_org_submit');
     if (isObject(editor)) {
       if ($('#org_feedback_enabled_true').is(':checked')) {
         editor.setMode('code');
+        if (editor.getContent().trim() === '') {
```
briri: Did you try just adding the `"aria-required": true` to the field? We moved away from using JS to enforce form validations and instead rely on validations defined in the model (you can require the email body there with an `validates :feedback_email_msg, presence: true, if: { feedback_enabled: true }` clause and then displaying the appropriate success/failure to the user (see this section for an example: https://github.com/DMPRoadmap/roadmap/blob/4d9665ef8c76736ac750fe6ed0763da26be2a0f3/app/controllers/orgs_controller.rb#L57)

johnpinto1: @briri The functionality you mention was already correctly coded and working. (The bug was noted in July 2018, it appears to be fixed since then). To make the sample message copy-and-paste friendly I have moved the message that was in in the tooltip to be beside the input textarea.
selection_122

Issue #2058, opened 2019-02-21
CSV download issue - missing information and mis-formatting


sjDCC:
A user from the Netherlands has reported some mis-formatting on the CSV download. There are two issues:

1. Any text in the freetext box after a radio/dropdown/multi-select isn't displaying in the export
2. Line breaks in freetext boxes are being represented as semi-colons

I tested this with the Hartstichtung (Dutch Heart Foundation) template. See screengrabs and I can share the csv download
![test](https://user-images.githubusercontent.com/6169940/53193890-0bf9bd80-360a-11e9-9153-6b503d744ea1.PNG)
![capture](https://user-images.githubusercontent.com/6169940/53193821-df45a600-3609-11e9-88b4-26b77c967684.PNG)

johnpinto1:
@sjDCC I am going to look at this.

dsisu:
This issue has been fixed. I also tested the download both numbered and bulleted lists, tables and links. The text is exported but the formatting is removed, which is to be expected of downloads in CSV formats.

PR #2060, opened 2019-02-27
Added the missing information in CSV export plans.

johnpinto1:
The answers with question options was not including the answer.text, but overwriting the variable answer_text that held the value.

Fix for Issue #2058.

Not able to replicate "    2\. Line breaks in freetext boxes are being represented as semi-colons" so can't fix it.


johnpinto1:
Changing naming of branch. I was following the convention of another project I am on. So closing this.

PR #2065, opened 2019-03-06
Issue1158

### LINKED ISSUE #1158: Date picker question type option is not working
We have an option to pick a date as a question type but this is not working.

All seems well when you first select it
first-edit

But different options are presented when you re-edit
re-edit

And on the front end, nothing is displayed aside from the comments box. Not even the question text.
how-question-displays

I think this was a code addition from Benjamin so it would be good to go back to that PR and see how it was supposed to work. it seems something has got mangled in interim and mixed up dates with multiple choice options and other styles which have the optional comments box.

briri:
Fixes #1158 

Having a small issue with the JQuery-UI datePicker polyfill. 

@Bodacious, @xsrust and @johnpinto1 maybe one of you can see what I'm doing wrong here. I was originally using the jquery-ui theme that's already located in `node_modules` but the 'prev' and 'next' text and icons were not visible in the UI (although the links were there and you could click on them, there was just no visual cue for the user). 
I then went out to the JQuery-ui site and pulled down a custom theme containing only the bits necessary for datePicker. I added them into the application.scss (although I think its probably preferable to use the node_modules one) this seemed to work but the 'prev' 'next' icons are still missing. Its at least now has a `background-url` (see screenshot) but its looking in an odd place. 

I think its probably just an issue with the webpacker config but I don't have bandwidth for the rest of the week to dig deeper. If either of you has any time to take a look I would appreciate it

To do this: open this branch locally in Safari and create a plan based off of a template that has a date field (you may need to add a question to an existing template). Then edit the plan and expand the section. The datePicker polyfill will open when the field gets focus.

screen shot 2019-03-06 at 9 58 32 am


johnpinto1:
@briri @xsrust @Bodacious I have added missing icons and fonts based on these errors I found on Safari. See image
screenshot 2019-03-07 at 11 28 24
The issue I am finding is that the **assets**  for the JQuery-UI are being precompiled to the wrong folder, e.g, images copied to **public/assets/jquery-ui** folder instead of **public/assets/jquery-ui**.
So http://localhost:3000/assets/jquery-ui/datepicker/images/ui-icons_FFF_256x240.png works but the datepicker is expects http://localhost:3000/assets/vendor/jquery-ui/datepicker/images/ui-icons_FFF_256x240.png 

Also the missing *font* which I added is not being copied so http://localhost:3000/fonts/GillSans-Light.ttf fails.

Also I had to rename *ui-icons_ffffff_256x240.png* to *http://localhost:3000/fonts/GillSans-Light.ttf*. Will continue investigating.

briri:
took a stab at adding `jquery-ui-rails` gem but it didn't work. I think due to a config or load timing issue with web packer. The jquery-ui docs instruct you to add a `//= require 'query-ui'` line to your application.js which uses some gem magic to do all the necessary wiring. We are using webpacker though so the application.js lives outside the normal asset pipeline infrastructure.

We are also already making use of jquery-ui's `sortable` functionality via webpacker so we would likely need to rethink our asset management strategy. 

Going to try to do some css hacking to get the proper images in place.

PR #2068, opened 2019-03-19
fix bugs with API

xsrust:
re-add date filtering, revert assumption of last month by default, re-add instance variables, fix other minor bugs

Fixes :
Plans and using_template endpoints broken, others responding differently to lack of date options than expected


app/controllers/api/v0/statistics_controller.rb
```diff
@@ -126,19 +126,23 @@ def created_plans
   # Displays the number of DMPs using templates owned/create by the caller's Org
   # between the optional specified dates
   def using_template
+    org_templates = @user.org.templates.where(customization_of: nil)
```
briri: ha. good catch. looks like that error was always triggered before

Issue #2080, opened 2019-04-11
Search function is case sensitive and shouldn't be


dsisu:
Please complete the following fields as applicable:

**Expected behaviour:**

Searching for a plan title by using the same term - e.g. autistic - should yield the same results irrespective of how the search term is spelt, i.e. lowercase (autistic) or title case (Autistic).

Sam says that searcheable fields should not be case sensitive.

**Actual behaviour:**

If user searches for a plan title containing the term 'Autistic' it returns three results, while if it searches for the same term all lowercase - 'autistic' - it returns one result. Ideally, both searches should return the same result, i.e. a total of four plan titles contain the term 'autistic'.

**Steps to reproduce:**

Please search on the system for plan title containing words 'autistic' or 'Autistic'.

**Special note.**
Please note that I have raised a similar issue before, see for example issue no 1870 or 1319. More recently, the issue was raised by Georgina Parsons from Cranfield University.

Ideally, we should ensure none of the searcheable fields is case sensitive. Users should be able to search for plans, users, institutions, etc, without worrying about letter case.

Let's deal with this because this query keeps popping up and it affects both ordinary users and org_admin users, particularly the latter.

Thanks!

sjDCC:
This doesn't seem to be fixed @johnpinto1 

The following record only comes up when I search "SMART" lower case or capital first letter doesn't return it
![Capture](https://user-images.githubusercontent.com/6169940/56307656-43ce3e80-613d-11e9-8b19-828ad11ee174.PNG)

johnpinto1:
@sjDCC The code is currenty only on DEV. @xsrust will be deploying it later to UAT. I tried search of users on DEV
![Selection_200](https://user-images.githubusercontent.com/8876215/56358567-e71f6200-61d6-11e9-8b54-c30b19819924.png)

sjDCC:
All working well. Closing out. Thanks

PR #2097, opened 2019-04-17
Added code to make Search functionality case-insensitive.

johnpinto1:
Changes:
   - relevant models with :search scope updated with lower casing  of the column
     values and the search parameters.
   - fixed all rubocop warnings on the changed models except
     "Style/RedundantReturn: Redundant return detected."

   Note: one could not use 'ILIKE' in WHERE clauses changed as it is not supported by MariaDB and MySql.
   Hence we use lower() in changed WHERE clauses.

 Fix for issue #2080

### DIFF
--- a/app/models/guidance.rb
+++ b/app/models/guidance.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 # Guidance provides information from organisations to Users, helping them when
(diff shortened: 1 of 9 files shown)


johnpinto1:
Fixing rubocop rspec on Travis. So don't merge yet.

PR #2121, opened 2019-04-30
updated permissions and routing around public template export

### LINKED ISSUE #2118: Template Export Does not render useful error messaging for bad id's

**Expected behaviour:**
- User gets a flash-message to let them know the id they have asked to download is bad
**Actual behaviour:**
- User gets a generic 404
**Steps to reproduce:**
pass a non-existant template_id to the template-export action of the public pages

xsrust:
Fixes #2118 .

Exported templates must additionally exist and be published

### DIFF
--- a/app/controllers/public_pages_controller.rb
+++ b/app/controllers/public_pages_controller.rb
@@ -2,8 +2,6 @@
 
 class PublicPagesController < ApplicationController
 
-  after_action :verify_authorized, except: [:template_index, :plan_index]
-
   # GET template_index
   # -----------------------------------------------------
   def template_index
@@ -22,7 +20,6 @@ class PublicPagesController < ApplicationController
     @template = Template.live(params[:id])
     # covers authorization for this action.
     # Pundit dosent support passing objects into scoped policies
-    skip_authorization
     unless PublicPagePolicy.new(@template).template_export?
       redirect_to public_templates_path, notice: "You are not authorized to export that template" and return
       #raise Pundit::NotAuthorizedError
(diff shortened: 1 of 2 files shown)


app/controllers/public_pages_controller.rb
```diff
@@ -22,10 +22,11 @@ def template_export
     @template = Template.live(params[:id])
     # covers authorization for this action.
     # Pundit dosent support passing objects into scoped policies
+    skip_authorization
```
briri: prefer to just add this to the `verify_authorized` exception list above since we're already making exceptions.

app/policies/public_page_policy.rb
```diff
@@ -14,7 +14,7 @@ def template_index?
   end
 
   def template_export?
-    @object.is_default || @object.org.funder?
+    @object.present? && ( @object.is_default || @object.org.funder? ) && @object.published
```
briri: Not necessary but you should be able to add a `?` on `is_default?` and `published?`

xsrust: Good call, I'll update those.

xsrust:
Hi Briri, I completely agree with your comments, I went back and forth on which way to do this, but had settled on this approach because it captures the different requirements of the actions.  The template and plan indexes actually require no authorization, whereas this action requires it, so we verify (and skip) authorization.  I believe the initial thought behind having the after-action verifying was to ensure we don't loose this auth.  That being said, It may make more sense to remove the verification alltogether as it's not applicable to the other two actions.

Additionally, We've had a request to allow templates to be exported from the org-admin menu, which would logically move the template_export action out of the public pages and into a different controller, at which point the other two actions remaining in the controller will not need authorization.

Issue #2089, opened 2019-04-16
Pre-fill ID on the plan details page


sjDCC:
The 5 digit ID used in the URL acts as a unique ID in the tool. This should be automatically pre-filled in the ID field on the plan details page.

![ID](https://user-images.githubusercontent.com/6169940/56212406-dabdcc80-6051-11e9-9692-46b3fa46862e.PNG)

xsrust:
@sjDCC 
Do we want to do this for all pre-existing plans which do not have anything set in this ID field?

sjDCC:
I would say so. Do any have anything in the ID field? I don't think it's being actively used

briri:
We have hidden that field in the DMPTool. I don't recall what it was used meant for. Perhaps an internal Uni identifier?

sjDCC:
It was intended as a plan ID but we never used it. Manchester uni do actively use it now so we plan to pre-populate. I guess this shouldn't have any knock-on effect for you if it's hidden but good to flag any ways in which code needs to be implemented if so.

magdalenadrafiova:
Hiya all, 

I do not still see the ID being filled automatically after saving my plan... 
![Pre-fill ID on the plan details page #2089](https://user-images.githubusercontent.com/47178106/57528762-0ff4cc00-732b-11e9-81f6-e2eadd99a34f.PNG)

johnpinto1:
@magdalenadrafiova This will only work for newly created plans. We don't want to overwrite ids of plans if they have been added.

magdalenadrafiova:
I understand - but when I follow the steps to create a new plan it does not create the id again another example - not sure whether I might be doing something wrong and maybe therefore it is not generating it 
![example-2-10052019-2089](https://user-images.githubusercontent.com/47178106/57529162-1df71c80-732c-11e9-9eaa-cb0a2eaf29e4.PNG)

magdalenadrafiova:
Works perfectly - 
![2089 - pre fill](https://user-images.githubusercontent.com/47178106/57530085-5e579a00-732e-11e9-912b-efce02d4d0ea.PNG)

xsrust:
@johnpinto1 
I actually added a rake task:
https://github.com/DMPRoadmap/roadmap/blob/0f24533af50726c7a66e88230e6ff42d0dcacf11/lib/tasks/upgrade.rake#L670-L672

To auto-fill any blank identifiers

johnpinto1:
Great @xsrust

[...]

PR #2124, opened 2019-05-01
Pre-fill ID on the plan details page.

johnpinto1:
Changes:
  - the default plan :identifier is set to the plan :id on creation in the
    PlanControllers create method.
  - added a method add_identifier to Plan model.

Fix for issue #2089

Selection_222

@xsrust and @briri given my ruby rails inexperience please comment if this is the the best way to do what was required by issue

### DIFF
--- a/app/models/plan.rb
+++ b/app/models/plan.rb
@@ -450,6 +450,14 @@ class Plan < ActiveRecord::Base
     end
   end
 
+  ## Update plan identifier.
+  #
+  # Returns Boolean
+  def add_identifier!(identifier)
+    self.update(identifier: identifier)
+    save!
+  end
+
   ##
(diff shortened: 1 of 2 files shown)


app/models/plan.rb
```diff
@@ -450,6 +450,14 @@ def add_user!(user_id, access_type = :commenter)
     end
   end
 
+  ## Update plan identifier.
+  #
+  # Returns Boolean
+  def add_identifier!(identifier)
```
xsrust: I like this being a function on the plan model so we can update it's behavior if we decide to go with some kind of identifier other than the plan's ID

briri: Yes! This may be a good spot to place any DOIs if we start minting them for DMPs.

PR #2126, opened 2019-05-02
Issue/2115

xsrust:
- Add publishability check  to templates model
- Added tests for publishability check
- refactored factory definition of templates/phases/sections to allow specification of number of sub-models to generate.  This is currently done by passing a number of sub-models for each of the current model to have.  i.e. `{phases:2 , sections:3}` will result in 2 phases, each with 3 sections, so a total of 6 sections created.  This could be re-done down the line to pass a structured object so that each of the models may have a different number of sub-models, but this granularity was not needed for my tests
- refactored templates_controller to use publishability check

### DIFF
--- a/app/models/template.rb
+++ b/app/models/template.rb
@@ -403,6 +403,36 @@ class Template < ActiveRecord::Base
     update!(published: true)
   end
 
+  def publishability
+    error = ""
+    publishable = true
+    # template must be the most recent draft
+    if published
+      error += _("You can not publish a published template.  ")
+      publishable = false
+    end
+    if not latest?
+      error += _("You can not publish a historical version of this template.  ")
+      publishable = false
+    # all templates have atleast one phase
+    end
+    if not phases.count > 0
+      error += _("You can not publish a template without phases.  ")
+      publishable = false
+    # all phases must have atleast 1 section
+    end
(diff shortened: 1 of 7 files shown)


app/models/template.rb
```diff
@@ -403,6 +403,36 @@ def publish!
...
+    unless phases.map{|p| p.sections.count > 0}.reduce(true) { |fin, val| fin and val }
+      error += _("You can not publish a template without sections in a phase.  ")
+      publishable = false
+    # all sections must have atleast one question
+    end
+    unless sections.map{|s| s.questions.count > 0}.reduce(true) { |fin, val| fin and val }
+      error += _("You can not publish a template without questions in a section.  ")
+      publishable = false
+    end
+    return publishable, error
```
briri: publishable could probably just be replaced with `error.blank?`

spec/features/templates/templates_upgrade_customisations_spec.rb
```diff
@@ -63,16 +64,22 @@
 
     click_link "Add a new section"
     within('#new_section_new_section') do
-      fill_in :new_section_section_title, with: "New section title"
-      tinymce_fill_in :new_section_section_description, with: "New section title"
+      fill_in :new_section_section_title, with: "Cool New section title"
```
briri: I like that this is now cool ;)

PR #2128, opened 2019-05-06
Add typeahead field for Research Grant ID

Bodacious:
Changes proposed in this PR:
- Adds a typeahead select to the grant ID field so that users can choose their research project from https://api.openaire.eu/projects/dspace/h2020/ALL/ALL

### DIFF
--- a/app/controllers/research_projects_controller.rb
+++ b/app/controllers/research_projects_controller.rb
@@ -9,7 +9,6 @@ class ResearchProjectsController < ApplicationController
 
   def search
     @results = research_projects.select { |r| r.description.match(params[:description]) }
-    logger.debug("Returning #{@results.count} results")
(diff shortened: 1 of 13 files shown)


app/controllers/research_projects_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+
+  DEFAULT_FUNDER_TYPE = "H2020"
+
+  def index
+    render json: research_projects
+  end
+
+  def search
+    @results = research_projects.select { |r| r.description.match(params[:description]) }
+    logger.debug("Returning #{@results.count} results")
```
briri: This log statement doesn't have much context. Either delete it (if it was just there for dev debugging purposes) or add some more context

PR #2129, opened 2019-05-08
Feature/reviewer role

xsrust:
Changes the data-structure backing the review process.
Previously, this was done with the `reviewer` `Role`, but this shifts the access to be a permission.

Submission logic is that the plan is sent to the creator's org for review.
Then, org-admins with the `reviewer` permission can read and comment on the plan.

Additionally, a bug was found that previous reviews were not actually removed, but rather the `Role` persisted, with a `nil` `user_id`.
Rake tasks to fix that issue, along with migrating any work-in-progress reviews to give the `reviewer` the correct permissions, have been added with the `upgrade:v2_1_3` task.

Finally I scrubbed all mentions of the `Role.reviewer`, and updated the tests to reflect the new logic.


app/controllers/plans_controller.rb
```diff
@@ -259,7 +259,7 @@ def share
     if @plan.present?
       authorize @plan
       # Get the roles where the user is not a reviewer
```
briri: You can remove this comment

app/helpers/perms_helper.rb
```diff
@@ -2,14 +2,15 @@ module PermsHelper
...
     {
-      :add_organisations => _('Add organisations'),
-      :change_org_affiliation => _('Change affiliation'),
-      :grant_permissions => _('Manage user privileges'),
-      :modify_templates => _('Manage templates'),
-      :modify_guidance => _('Manage guidance'),
-      :use_api => _('API rights'),
-      :change_org_details => _('Manage organisation details'),
-      :grant_api_to_orgs => _('Grant API to organisations')
+      add_organisations:  _('Add organisations'),
```
briri: thanks for converting these over

app/models/plan.rb
```diff
@@ -279,12 +278,6 @@ def request_feedback(user)
       begin
         self.feedback_requested = true
         # Share the plan with each org admin as the reviewer role
```
briri: remove this comment since its no longer applicable

spec/models/plan_spec.rb
```diff
@@ -880,18 +900,53 @@
       end
     end
 
-    it "when user is a reviewer" do
+    it "when user is a reviewer and feedback requested" do
       # All reviewers should be able to comment
-      subject.roles.reviewer.pluck(:user_id).each do |user_id|
-        expect(subject.commentable_by?(user_id)).to eql(true)
-      end
+      user = create(:user)
```
briri: not a big deal, but you could offload these duplicated lines to a `before` block to make it more DRY

PR #2133, opened 2019-05-10
Issue2088 - THIS A TEMP PR (NOT FOR MERGING)

johnpinto1:
Fixes # .

Changes proposed in this PR:
-

### DIFF
--- a/app/controllers/departments_controller.rb
+++ b/app/controllers/departments_controller.rb
@@ -1,0 +1,74 @@
+# frozen_string_literal: true
+
+class DepartmentsController < ApplicationController
+
+  after_action :verify_authorized
+  respond_to :html
+
+  # GET add new department
+  def admin_new
+    @department = Department.new
+    authorize @department
+  end
+
+  # POST /departments
+  # POST /departments.json
+  def admin_create
+    @department = Department.new(department_params)
+    authorize @department
+    @department.org_id = current_user.org_id
+
+    if @department.save
+      flash.now[:notice] = success_message(@department, _("created"))
+      # reset value
+      @department = nil
+      render :admin_new
+    else
+      flash.now[:alert] = failure_message(@department, _("create"))
+      render :admin_new
+    end
+  end
+
+  # GET /departments/1/edit
+  def admin_edit
+    @department = Department.find(params[:id])
+    authorize @department
+  end
+
+  # PUT /departments/1
+  def admin_update
+    @department = Department.find(params[:id])
+    authorize @department
+    @department.org_id = current_user.org_id
+
+    if @department.update(department_params)
+      flash.now[:notice] = success_message(@department, _("saved"))
+      render :admin_edit
+    else
+      flash.now[:alert] = failure_message(@department, _("save"))
+      render :admin_edit
+    end
+  end
+
+  # DELETE /departments/1
+  def admin_destroy
+    @department = Department.find(params[:id])
+    authorize @department
+    url = "#{admin_edit_org_path(current_user.org_id)}\#departments"
+    if @department.destroy
+      flash[:notice] = success_message(@department, _("deleted"))
+      redirect_to url
+    else
+      flash[:alert] = failure_message(@department, _("delete"))
+      redirect_to url
+    end
+  end
+
+
+  private
+
+  def department_params
+    params.require(:department).permit(:id, :name, :code, :org_id)
+  end
+
+end
--- a/app/controllers/paginable/departments_controller.rb
+++ b/app/controllers/paginable/departments_controller.rb
@@ -1,0 +1,20 @@
+# frozen_string_literal: true
+
+class DepartmentsController < ApplicationController
+
+  after_action :verify_authorized
+  respond_to :html
+
+  include Paginable
+
+  # /paginable/departments/index/:page
+  def index
+    authorize Department
+    paginable_renderise(
+      partial: "index",
+      scope: Department.by_org(current_user.org),
+      query_params: { sort_field: "departments.name", sort_direction: :asc }
+    )
+  end
+
+end
--- a/app/models/org.rb
+++ b/app/models/org.rb
@@ -73,6 +73,7 @@ class Org < ActiveRecord::Base
 
   has_many :identifier_schemes, through: :org_identifiers
 
+  has_many :departments
 
   # ===============
   # = Validations =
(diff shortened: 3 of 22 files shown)


app/controllers/paginable/departments_controller.rb
```diff
@@ -0,0 +1,17 @@
...
+class Paginable::DepartmentsController < ApplicationController
+
+  include Paginable
+
+  # /paginable/Departments/index/:page
+  def index
+
+    paginable_renderise(
+      partial: "index",
+      scope: Departments.by_org(current_user.org),
```
xsrust: I'm unfamiliar with the `.by_org` built-in, is this provided by the Paginable package?
I would have expected this to be `current_user.org.departments`

app/models/org.rb
```diff
@@ -151,6 +152,10 @@ class Org < ActiveRecord::Base
               count(distinct templates.family_id) as template_count,
               count(users.id) as user_count")
   }
+  
+  scope :departments, -> do
+    where(org_id: org.id)
+  end
```
xsrust: I don't think you should need to define this, it should be done auto-magically via the `has_many` relationship

app/views/devise/registrations/_personal_details.html.erb
```diff
@@ -29,6 +29,15 @@
   <% if org_admin %>
     <input type="hidden" id="original_org" value="<%= @user.org_id %>">
   <% end %>
+  
+    <% departments = (current_user.org.departments.present? ? current_user.org.departments : []) %>
```
xsrust: This should return [] when empty, so I don't think you need the conditional opperator

app/controllers/departments_controller.rb
```diff
@@ -0,0 +1,26 @@
...
+  def create_or_update
+    p_params = permitted_params()
+  
+    @department = Department.find_or_initialize_by(name: p_params[:name], org_id: p_params[:org_id])
+    
+    @department[:name] = p_params[:name]
+    @department[:code] = p_params[:code]
+    @department[:org_id] = p_params[:org_id]
+    authorize @department
+    @department.save!
```
briri: overall this looks fine. I think though that you could simplify it down to:
```ruby
def create_or_update
  @department = Department.find_or_initialize_by(name: permitted_params[:name], org_id: permitted_params[:org_id])
  authorize @department
  @department.save(permitted_params)
end
```

app/models/user.rb
```diff
@@ -77,6 +78,8 @@ class User < ActiveRecord::Base
   belongs_to :language
 
   belongs_to :org
+  
+  belongs_to :department
```
briri: I would add the optional flag here unless the plan is for every org to have a default department: `belongs_to :department, optional: true`

app/views/orgs/_departments.html.erb
```diff
@@ -0,0 +1,48 @@
+<%# locals: { context, title, tooltip } %>
+<% department =  department.nil? ? Department.new :  department %>
+<%= form_for(department, url: "/departments/create_or_update", html: { multipart: true, method: method, id: "edit_org_departments_form" } ) do |f| %>
```
briri: use the rails route helpers instead of the hard coded path. Do a `rake routes` and then find the line with `departments#create_or_update`. You should then see something like `POST department` at the beginning of that line. Take that value and add 'path' to the end for the route helper (e.g. `department_path`)

app/views/orgs/_departments.html.erb
```diff
@@ -0,0 +1,48 @@
...
+<%= f.hidden_field(:id, value: department.id) %>
+
+  <div class="department">
+    <div class="row">
+      <div class="col-xs-12">
+        <h3><%=_('Schools or Departments') %></h3>
+        <div class="row">
+              <div class="form-group col-xs-10">
+                <%= f.label :name, _('Name'), class: "control-label", "aria-required": true %>
+                <%= f.text_field :name, id: "org_department_name", class: "form-control" %>
```
briri: you shouldn't need to give it an id here. it should generate `department_name`. If all of the fields need to have the `org_` prefix on it, consider giving the form definition above a namespace instead: `<%= form_for(department, :org, url: department_path, html: { multipart: true, method: method, id: "edit_org_departments_form" } ) do |f| %>

spec/factories/departments.rb
```diff
@@ -0,0 +1,23 @@
...
+#  org_id     :integer
+#
+# Indexes
+#
+#  index_departments_on_org_id  (org_id)
+#
+
+FactoryBot.define do
+  factory :department do
+    name { "MyString" }
```
briri: Use Faker instead (e.g. `Faker::Education.unique.subject`)

johnpinto1:
@xrsust & @briri Thanks for the comments.

app/models/user.rb
johnpinto1: @briri I had to use `belongs_to :department, required: false` as  `:optional` only in `Rails 5`.

briri: 👍 sorry. I had been reading up on Rails 5 recently. Its hard to keep these things straight sometimes ;)

johnpinto1: @briri Good to know about when we upgrade.

johnpinto1:
@briri @xsrust Please can you have a lokk at it on Monday. Updated code. I expect a lot of suggestions for changes. The Basic functionality is there. Some screenshots
Screenshot from 2019-05-24 14-24-45
Selection_011

johnpinto1:
> Thanks for incorporating the changes @johnpinto1. I think the changes look fine.
> 
> From a UI/UX experience though this is a different approach. I think maybe @sjDCC and @magdalenadrafiova should weigh in on the design. It won't be a screen that has a lot of heavy use and I think the approach you've taken is a good/better design (fewer clicks for the user) but we may want to stick with the approach we've taken elsewhere (e.g. Themes section ... table of themes with an 'add theme' button and links to 'edit' and 'delete' which drive in to a separate page to do the add/edit)

Thanks @briri will look at the *Guidance* themes approach and try to incorporate it.

sjDCC:
Hi @johnpinto1 I agree with @briri that it's best to stick with layout / style of other pages for consistency. The reworked scrrengrabs you sent look great

johnpinto1:
Closing this to create a new request.

PR #2148, opened 2019-05-30
Bug/export txt csv

xsrust:
Fixes an issue where blank answer text could cause csv and text export to throw errors


app/models/answer.rb
```diff
@@ -98,7 +98,7 @@ def is_valid?
       if question.question_format.option_based?
         return question_options.any?
       else  # (e.g. textarea or textfield question formats)
-        return text.present?
+        return not(is_blank?)
```
briri: good use of the existing helper method 👍

Issue #2088, opened 2019-04-16
Add 'school/department' field to edit profile page


sjDCC:
At the two DMPonline user groups, there was a big demand for universities to know which School/Department/Unit their users came from. There is desire to have a question that organisations can define which would be added to the edit profile page under their institution.

Suggested workflow:
Under 'org' page on admin interface, add a tab where this question can be controlled. 
Admin users should be able to specify the label for the field e.g. "School" or "Department"
Admin users should create a controlled value list as a dropdown or checkbox
This should display as a combobox

On the front end "edit profile' page, end users should see this question beneath their organisation field. If the organisation has not set the entry, then no question will show.

Make this field editable for admins e.g. on the user page view an admin has

xsrust:
I suggest that we add a new model for a 'department' with users having one department and orgs having many.  
Org-admins can define additional departements, and change the label for their org on what they want to call these sub-groupings.

For-now, it shouldbe sufficient just to:
- allow org-admins to define the list from the edit-org-details page
- allow users (who's organisations have defined the groupings) to select from their org's list on edit-profile
- display this category in the users table of the org/super admin interfaces
- Remove this attribute when the user changes organisation

down the line, when we add more functionality to this (i.e. only reviewers of the same department get notified) may want to re-factor this so that the user gains their org through the 'department'.
I think if we do the association correctly, and generate a 'main' or 'generic' `department` for all the existing orgs, we would not need to re-write the selection logic for `user.org` anywhere in the tool.

I'm hesitant to make the department such an integral object immediately though

johnpinto1:
@xsrust @briri I found a bug for the current scenario.

- As **Org admin**, go to **Organisation details** page and create**  a **school/department** for your **org**, say **Chemistry** in 
- Go to **Edit profile** page and add **Chemistry**  as a school/department for a user
-  Now as **Org admin**, go to **Organisation details** page and delete the **school/department**  **Chemistry**

I am a way of deleting the **department_id** from users for **Chemistry** when I delete it.

johnpinto1:
@xsrust @briri Found a fix in this blog https://til.hashrocket.com/posts/jqilnp3d0h--set-foreign-keys-to-null
Will push a new PR soon.

johnpinto1:
Steps to test once last PR accepted:
- Log in as an **Org admin**.
- Go to **Organistaion details** page and select **School/Department** tab
- Initially, the page will be empty
[...]

PR #2150, opened 2019-06-03
Issue2088 - Add functionality for School/Department for an organisation.

johnpinto1:
Add functionality for School/Department for an organisation.
    
    Changes:
    - added a Department model whose instances belong to an org and require name.
    - added a DepartmentController and views for create, edit/update and a index view which is a tab
    of 'Organisational Details'.
    - added the optional attribute department_id to User model.
    - added Rspec tests.
 - Updated app/controllers/registrations_controller.rb, app/models/org.rb and
    app/models/user.rb to get rid of outstanding Rubocop warnings

Fix for issue #2088.

Based on changes requested in closed temporary PR 
https://github.com/DMPRoadmap/roadmap/pull/2133

Selection_016
Selection_017

Selection_018

### DIFF
--- a/app/policies/department_policy.rb
+++ b/app/policies/department_policy.rb
@@ -1,0 +1,37 @@
+# frozen_string_literal: true
+
+class DepartmentPolicy < ApplicationPolicy
+
+  attr_reader :user
+  attr_reader :department
+
+  def initialize(user, department)
+    raise Pundit::NotAuthorizedError, "must be logged in" unless user
+    @user = user
+    @department = department
+  end
+
+  def new?
+    @user.can_org_admin?
+  end
+
+  def create?
+    @user.can_org_admin?
+  end
+
+  def edit?
+    # Only org_admins can edit their own org's departments
+    @user.can_org_admin? && @user.org.id === @department.org_id
+  end
+
+  def update?
+    # Only org_admins can update their own org's departments
+    @user.can_org_admin? && @user.org.id === @department.org_id
+  end
+
+  def destroy?
+    # Only org_admins can delete their own org's departments
+    @user.can_org_admin? && @user.org.id === @department.org_id
+  end
+
+end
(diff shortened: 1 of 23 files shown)


app/controllers/departments_controller.rb
```diff
@@ -0,0 +1,74 @@
+# frozen_string_literal: true
+
+class DepartmentsController < ApplicationController
```
briri: Move this into the `app/controllers/org_admin` dir and the `OrgAdmin` namespace. Then remove the `admin_` prefix from the action names so that we just have: `new`, `update`, etc. 

The other `admin_` prefixed controller actions out there are legacy code we never finished cleaning up.

app/policies/department_policy.rb
```diff
@@ -0,0 +1,47 @@
...
+  attr_reader :user
+  attr_reader :department
+
+  def initialize(user, department)
+    raise Pundit::NotAuthorizedError, "must be logged in" unless user
+    @user = user
+    @department = department
+  end
+
+  def admin_new?
```
briri: update these to match the new action names in the controller (e.g. new, update, etc.)

app/policies/department_policy.rb
```diff
@@ -0,0 +1,47 @@
...
+    puts "admin_new?: @department"
```
briri: looks like you can remove these debug lines

config/routes.rb
```diff
@@ -66,6 +66,16 @@
       put 'admin_update'
     end
   end
+  
+  resources :departments, :path => 'org/admin/department', only: [] do
```
briri: you can then remove this and add it under the `org_admin` namespace as a single line: 
```ruby
namespace :org_admin do
  ...
  resources :departments
  ...
end
```

johnpinto1:
@briri will work on it tomorrow.

app/controllers/departments_controller.rb
xsrust: it would be a good bit of cleanup to go through and re-factor those to the org_admin namespace next sprint.

PR #2165, opened 2019-06-19
Fix for issue associated with super-admins and department affiliation…

johnpinto1:
… in a given organisation.

Changes made:
  - Department policies now allow both org admins and super admins to
    create update and delete departments, whilst org_admins can only do this
    for their own org. Super admins can do it for any org department.
  - When a super admin switches organisation their department_id is cleared.

### DIFF
--- a/app/views/devise/registrations/_personal_details.html.erb
+++ b/app/views/devise/registrations/_personal_details.html.erb
@@ -31,6 +31,7 @@
   <% end %>
   
     <% departments = current_user.org.departments.order(:name) %>
+    <% if departments.count > 0 %>
     <div class="form-group col-xs-8">
       <% dept_id = current_user.department.nil? ? -1 : current_user.department.id  %>
       <%= f.label(:department_id, _('Department or school'), class: 'control-label') %>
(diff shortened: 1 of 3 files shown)


app/views/devise/registrations/_personal_details.html.erb
```diff
@@ -31,6 +31,7 @@
   <% end %>
   
     <% departments = current_user.org.departments.order(:name) %>
+    <% if departments.count > 0 %>
     <div class="form-group col-xs-8">
       <% dept_id = current_user.department.nil? ? -1 : current_user.department.id  %>
```
briri: you shouldn't need to do this. does it not display the blank selection if the user's department is nil? the update on the controller should also be ok with not receiving the param if the user has left it blank

johnpinto1: @briri I did this because @sjDCC was seeing it https://github.com/DMPRoadmap/roadmap/issues/2088#issuecomment-500905694. Will re-test tomorrow. To see if my changes no longer make it necessary.

johnpinto1: @briri @xsrust That check is necessary to prevent the *School or department* selection appearing if the Org has no departments set.

xsrust: The form would be disabled if there were no departments, but this prevents it from being displayed at all, so is cleaner for the user.

PR #2178, opened 2019-07-09
Fixed link for UC3 at footer since the page was previously not found.

killiandunne2:
Fixes # .

Changes proposed in this PR:
-
The link to the UC3 page on the California Digital Library website at the footer of pretty much every  was giving me a page not found error; I believe the link is out of date. Although this annotation doesn't seem to be available to the public I updated it anyway.


briri:
looks like we need to update our gems (although the OmniAuth one has no patch yet). Going to merge this one to dev and we can update them in a separate PR

killiandunne2:
I see, cheers!

PR #2179, opened 2019-07-09
updated gems and yarn dependencies

briri:
Ran Bundle update and yarn upgrade


xsrust:
I'm adding in an explicit disable of non-post omniauth methods, and letting bundle-audit ignore the CVE so the remaining tests can run, per the suggestions from Omniauth for mitigating:
https://github.com/omniauth/omniauth/wiki/Resolving-CVE-2015-9284

xsrust:
Faker::StarWars was depricated and failing some tests so I've updated the call to Faker::Movies::StarWars
https://github.com/stympy/faker/blob/master/CHANGELOG.md#deprecation-2

xsrust:
Multiple issues with the tests regarding timeago.js,
Downgrading this to the version being used on development, 4.0.0.beta.1, from 4.0.0.beta.2
and re-running Travis

PR #2181, opened 2019-07-11
Added tooltip on share plan page next to Permissions. Used width and flexbox styling here.

killiandunne2:


xsrust:
Hey,

I had another look at this and I think it's cleaner to have the tooltip render when the user is hovered over any part of that radio-group.  Then we don't need to add any custom CSS to try to get elements to align next to the legend.  

Screenshot 2019-07-16 at 12 23 56

See https://github.com/DMPRoadmap/roadmap/pull/2187

Issue #2111, opened 2019-04-24
Creation of tooltip on share tab - explanation of the permissions (co-owner, editor, read only)


magdalenadrafiova:
When the owner of the plan creates co-owner of the project (essentially people with the ability to share the plan with others) it is not clear whether editors and read-only receives any notifications when the reviewer sends e.g. feedback on the plan. It would be useful for the users to explain the permissions and practical implications so the owner of the plan will know which role to choose for people who collaborate with her/him on the project. 

Would it be possible to create e.g. a tooltip on the share tab by the choice of permissions? 

Co-owner: Has admin-rights to the plan (can invite other users, view the plan, answer questions, or comment)
Editor: Has edit-rights to the plan (can view the plan, answer questions, or comment)
Read Only: Has read-rights to the plan(can view the plan or comment)

magdalenadrafiova:
@pherterich @mariapraetzellis I am closing this as I think with reworking the plans page we will have the 'contributros' or collaborators page where certain defintions/permisions will be defined

xsrust:
Re-opening this one as the contributors/collaborators page provides maDMP functionality and attribution via the CREDiT taxonomy, while the share_plan page will continue to provide the access-rules for our plans.

xsrust:
That being said, it looks like the tool tips have already been implemented so this can remain closed

![share_plan_tooltip](https://user-images.githubusercontent.com/7352616/86911244-e8f27f80-c112-11ea-8d9e-abc170122421.png)

PR #2187, opened 2019-07-16
updated plan sharing form according to #2111, also added text to the …

xsrust:
…email field to clarify what happens if the user is not in our system

### Email
Screenshot 2019-07-16 at 12 23 56

###  Radio group
Screenshot 2019-07-16 at 12 24 03

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -53,7 +53,7 @@ config/branding.yml
 
 # Ignore some of the initializers
 config/initializers/recaptcha.rb
-# config/initializers/devise.rb
+config/initializers/devise.rb
 config/initializers/wicked_pdf.rb
 config/initializers/fingerprint.rb
 
--- a/app/views/plans/_share_form.html.erb
+++ b/app/views/plans/_share_form.html.erb
@@ -88,42 +88,51 @@
   <% new_role = Role.new %>
   <% new_role.plan = @plan %>
   <%= form_for new_role, url: {controller: :roles, action: :create }, html: {method: :post} do |f| %>
-    <div class="form-group col-xs-8">
+    <div class="form-group col-xs-6">
       <%= f.hidden_field :plan_id %>
       <%= f.fields_for :user do |user| %>
-        <%= user.label :email, _('Email'), class: 'control-label' %>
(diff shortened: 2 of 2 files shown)


xsrust:
Rebased off development to try to get the travis to ignore the CVE in the bundle audit.

(Would have thought the pr part of the CI would do that automatically when it merged the new code in to development)

.gitignore
```diff
@@ -53,7 +53,7 @@ config/branding.yml
 
 # Ignore some of the initializers
 config/initializers/recaptcha.rb
-# config/initializers/devise.rb
+config/initializers/devise.rb
```
briri: wow. thanks for catching this.

app/views/plans/_share_form.html.erb
```diff
@@ -88,42 +88,51 @@
...
       <%= f.hidden_field :plan_id %>
       <%= f.fields_for :user do |user| %>
-        <%= user.label :email, _('Email'), class: 'control-label' %>
-        <%= user.email_field :email, for: :user, name: "user", class: "form-control", "aria-required": true %>
+        <%= user.label :email, _('Email'), class: 'control-label'%>
+        <%= user.email_field :email, for: :user, name: "user", class: "form-control",
+                            "aria-required": true,
+                            'data-toggle': 'tooltip',
+                            'data-html': true,
+                            title: _('Enter the email address of your collaborator: <ul><li> If they are already using DMPonline, they will see this plan on their dashboard, and recieve an email. </li> <li> If they are not currently using DMPonline, they will recieve an email inviting them to the tool so they can collaborate on your plan.</li></ul>') %>
```
briri: Use the app name defined in the branding.yml here instead of 'DMPonline'.

.gitignore
xsrust: Yeah, it kept overwriting my local version when I pulled a new branch down

app/views/plans/_share_form.html.erb
xsrust: Good catch,Good catch, completely slipped my mind, 
I'll update

PR #2190, opened 2019-07-23
DO_NOT_MERGE Translation.io -> Replaces FastGettext as translation gem

xsrust:
First-pass at replacing FastGettext with [translation.io gem.](https://github.com/translation/rails)
This gem library solves the i18n/gettext problem by interfacing between their locale-formats, and by automatically setting both of them when i18n's locale methods are called.

## Usage
- Still using gettext syntax `_("text")` to invoke translations
- Configuration of source/target locales, along with setup for available and default languages in `config/initilizers/translation.rb`
- Can override the locale being used with:  ` I18n.with_locale I18n.default_locale do ... end `

## work done
- Removed all references to gettext and replaced with I18n
- added configuration for open-source+Client translations (with the contents of app/views/branded going into the client)
- removed old gettext/i18n gems and their config
- added gitignore for client files
- synced latest versions of the translations to/from translation portal

## Remaining
- JS libraries (timeago.js) only in en-GB
- Investigate re-adding https://github.com/webhippie/gettext_i18n_rails_js for JS internationalization as it looks like it would just read from our po-files to generate JS-lookup versions of them.
- @briri is the formatting of the initilizer generic enough?  If not, how would be better to configure this?
  we could always do an example initilizer, or push more variables into the environment, but I wanted to get your oppinion here.  I dont think the branding.yml has been read when these get invoked, so we can't use that for configuration here.  I quite like the environment variable approach in combination with a .env file.  given we have the `dotenv-rails` gem, when calling any action with `bundle exec` all of those variables are available to the called program.
- Can likely remove the LocaleSet and LocaleFormatter from `lib` as the `translation` library super-seeds all the gettext vs I18n communication


briri:
closing in favor of @xsrust new PR

Issue #2174, opened 2019-07-09
Provide a different way of pop up guidance 


magdalenadrafiova:
It has been also reported that we need to remove the pop up guidance provided by the fields and provide this in another way such as text description under the question so it is compatible with assistive technology

sjDCC:
I guess this is an example of that. Looks good

![link](https://user-images.githubusercontent.com/6169940/65812060-e6452000-e1b9-11e9-809c-1e5b1154d335.PNG)

magdalenadrafiova:
Looks good

PR #2222, opened 2019-08-30
Fix for providing a text alternative for tooltips for assistive techn…

johnpinto1:
…ologies.

The fix consists of adding the current tooltip text to a <em> tag with the bootstrap class
sr-only that is made only available to assitive technologies.

Fix for issue #2174.

### DIFF
--- a/app/views/plans/_edit_details.html.erb
+++ b/app/views/plans/_edit_details.html.erb
@@ -1,3 +1,7 @@
+<% project_title_tooltip =  _('If applying for funding, state the name exactly as in the grant proposal.') %>
+<% project_abstract_tooltip = _("Briefly summarise your research project to help others understand the purposes for which the data are being collected or created.") %>
+<% id_tooltip = _('A pertinent ID as determined by the funder and/or organisation.') %>
+
 <div class="row">
     <div class="col-md-8">
     <%= form_for plan, html: {method: :put, class: 'form-horizontal edit_plan' } do |f| %>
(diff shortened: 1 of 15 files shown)


app/views/plans/_edit_details.html.erb
```diff
@@ -49,6 +52,7 @@
           <%= f.label(:description, _('Project abstract'), class: 'control-label') %>
         </div>
         <div class="col-md-8" data-toggle="tooltip" title="<%= _("Briefly summarise your research project to help others understand the purposes for which the data are being collected or created.") %>">
+          <em class="sr-only"><%= _("Briefly summarise your research project to help others understand the purposes for which the data are being collected or created.") %></em>
```
xsrust: I think it would be better to set the string once and re-use the variable, like you've done in other places

app/views/plans/_edit_details.html.erb
```diff
@@ -60,6 +64,7 @@
           <%= f.label(:identifier, _('ID'), class: 'control-label') %>
         </div>
         <div class="col-md-8">
+          <em class="sr-only"><%= _('A pertinent ID as determined by the funder and/or organisation.') %></em>
```
xsrust: It would be better to re-use a variable here too

johnpinto1: @xsrust missed that one. Will change and push.

johnpinto1: @xsrust Update code with two changes requested.

xsrust: Thanks @johnpinto1 
I wondered if you'd done those first

PR #2224, opened 2019-09-04
Allow Super admins to Archive and Merge users

xsrust:
Adds 2 options to the super-admin user-edit page:
`Archiving` will remove the user's name, email, and sign-in-IPs.
`Merging` will swap apply the currently edited user's ID to all of the selected user's notes, plans, and answers, then remove the selected user.

Screenshot 2019-09-04 at 11 54 14

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -411,10 +411,10 @@ class User < ActiveRecord::Base
     # merge logic
     # => answers -> map id
     to_be_merged.answers.update_all(user_id: self.id)
-    # => comments -> map id
-    to_be_merged.comments.update_all(user_id: self.id)
+    # => notes -> map id
+    to_be_merged.notes.update_all(user_id: self.id)
     # => plans -> map on id roles
-    to_me_merged.roles.update_all(user_id: self.id)
+    to_be_merged.roles.update_all(user_id: self.id)
     # => prefs -> Keep's from self
     # => auths -> map onto keep id only if keep does not have the identifier
     to_be_merged.user_identifiers.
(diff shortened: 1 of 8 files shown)


app/models/user.rb
```diff
@@ -326,10 +327,7 @@ def remove_token!
   # Returns Boolean
   def keep_or_generate_token!
     if api_token.nil? || api_token.empty?
-      new_token = loop do
-        random_token = SecureRandom.urlsafe_base64(nil, false)
-        break random_token unless User.exists?(api_token: random_token)
-      end
+      new_token = User.unique_random(field_name: 'api_token')
```
briri: nice. it seems more logical to place that functionality on the User model

app/models/user.rb
```diff
@@ -391,6 +389,41 @@ def acknowledge(notification)
...
+    recovery_email = nil
+    api_token = nil
+    encrypted_password = nil
+    last_sign_in_ip = nil
+    current_sign_in_ip =  nil
+    active = false
+    return self.save
+  end
+
+  def merge(to_be_merged)
```
briri: nice and clean. 👍

xsrust:
Thanks @briri 
I started by using a typeahead, but it was not too happy loading with all of our (32k) emails on my dev machine.  
I figured since it's only for super-admins at the moment, and they'll probably have the full email anyway, that this would do for a first-pass.

xsrust:
That being said, If we were to open it up to Org-admins, scoping it by their organisation should solve that issue of too many user accounts

briri:
makes sense @xsrust

Issue #2173, opened 2019-07-09
Making tooltips for logos more meaningful


magdalenadrafiova:
From our accessibility testing it has been pointed out that we need to make the tooltips for logos more meaningful - e.g. the University of Edinburgh logo says it is a link to the University of Edinburgh home page etc.

xsrust:
It looks like this is refering to the org-logo in the branding header:
https://github.com/DMPRoadmap/roadmap/blob/master/app/views/layouts/_branding.html.erb#L13-L17

The link is generated to the `org.target_url` field at the moment.
However, the organisations cannot currently edit this field.
@sjDCC I see two ways to proceed here,
First, we could make this editable, and also store the 'link text' as we're doing with the other links
Seccond, we could just remove this being a link, and rely on the branding links.

It looks like were going for the 2nd approach in the project at the moment, so it may be the most sane just to remove that field from the orgs table & not display any link here.

sjDCC:
Two would be the easy option but I guess this would mean the logo wouldn't be a link back to homepage or the unis org. Actually, where does the link usually direct to? My expectation when I click the logo top left is that it will take me back to the webpage home. For TUDelft's version of DMPonline I would expect to be taken back to https://dmponline.tudelft.nl not https://www.tudelft.nl/en/

Perhaps we can find a way to specify the link in a common way e.g. "DMP home page for [INSERT UNI NAME]" rather than removing the link

xsrust:
The link in the top-left (dmponline/roadmap logo) will still redirect to the homepage.

The link in question for this issue is that branded logo uploaded by organisations (in the example image, UoE).
Currently, that is only a link for a few historic orgs which happen to have data in that field, such as Edinburgh University.

![Screenshot 2019-09-02 at 12 44 06](https://user-images.githubusercontent.com/7352616/64112466-645f0400-cd7f-11e9-94e9-89a24c53a76c.png)


You bring up a great point on customizing the hover text for our different tenants, but that's outwith the scope of the Roadmap project.

sjDCC:
Ah right - then I think we're safe to remove it. We don't want to encourage people getting redirected out of DMPonline / DMPTool

johnpinto1:
Followed change suggested by @xsrust and removed the ability to click Org log link.
![My Dashboard - Chromium_152](https://user-images.githubusercontent.com/8876215/64334363-4805d600-cfd0-11e9-91f4-8eb79abb4386.png)

[...]

PR #2225, opened 2019-09-05
Removed the clickable link from the Organisation logo, as it was repo…

johnpinto1:
…rted to cause problems for Accessibility. The fact that image link would open the Organisations home page rather than return to the Roadmap page was confusing for users from an Accessibility pont of view.

Change:
  - Organisation logo if present is no longer clickable.

Fix for #2173.

### DIFF
--- a/app/views/layouts/_branding.html.erb
+++ b/app/views/layouts/_branding.html.erb
@@ -10,11 +10,10 @@
       </button>
       <% if user_signed_in? && !current_user.org.nil? %>
         <% if current_user.org.logo.present? %>
-          <%= link_to(image_tag(logo_url_for_org(current_user.org),
+          <%= image_tag(logo_url_for_org(current_user.org),
                                 alt: current_user.org.name,
                                 class: "org-logo",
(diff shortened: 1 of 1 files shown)


xsrust:
Looks good to me, Thanks John

johnpinto1:
@xsrust @briri I have mentioned this to @xsrust already:
If you try to up load a file for the logo, like an svg file (which we don't support currently)
My Dashboard - Chromium_152
Dragonfly::Shell::CommandFailed at -org-admin-1005339509-admin_update - Chromium_153

,  you get an error which is not very informative.

xsrust:
@johnpinto1 @briri 

I had a dig through the blame-history for the logo-file-type validation, and it dosent look like we had a rationelle for not supporting SVG other than it requring another processor for Dragonfly:
https://github.com/tomasc/dragonfly_svg
it looks pretty simple to configure if we want to suppor it

briri:
I second @xsrust suggestion of adding in the sag extension for dragonfly.

PR #2232, opened 2019-09-18
Sync Vulnerability patches from Master -> Development

xsrust:
fixes issue with devise, eslint-utils.

Also  attempts to fix the randomly failing annotations test


briri:
thanks @xsrust

PR #2244, opened 2019-09-27
Updated logic in Plan.num_answered_questions to allow for Phase context

### LINKED ISSUE #2095: Issue:counter of questions answered is incorrect in the 'request feedback'
It has been reported that the progress bar the numbers do not display correctly, please see below: 

In the Initial DMP tab, there are no numbers over the blue progress bar (I make it 9/9 though):

Blue progress bar

In the Detailed DMP tab, the user has cleverly answered 40 of 31 questions J (I think it’s 31/31 – I guess the first counter added the previous 9)
Progress bar - should be 31 from 31

And in the Final review DMP section, it still says 40/31 though they haven’t answered any questions (0/31):
Should be 0 from 31

briri:
Fixes #2095 .

Addresses an issue with the progress bar 'answered questions` count when a plan is for a template with multiple phases. Updated the `num_answered_questions` method to accept a Phase so that it can return either the total number of questions answered for the entire plan or a specific section.

### DIFF
--- a/app/models/plan.rb
+++ b/app/models/plan.rb
@@ -486,11 +486,13 @@ class Plan < ActiveRecord::Base
   # The number of answered questions from the entire plan
   #
   # Returns Integer
-  def num_answered_questions
-    Answer.where(id: answers.map(&:id))
-          .includes(:question_options, question: :question_format)
-          .to_a
-          .sum { |answer| answer.is_valid? ? 1 : 0 }
+  def num_answered_questions(phase = nil)
+    return answers.select { |answer| answer.answered? }.length unless phase.present?
+
+    answered = answers.select do |answer|
+      answer.answered? && phase.questions.include?(answer.question)
+    end
+    answered.length
   end
 
   # The number of questions for a plan.
@@ -526,7 +528,7 @@ class Plan < ActiveRecord::Base
                                           question: :question_format)
                                 .where(id: answer_ids)
     num_answers = pre_fetched_answers.reduce(0) do |m, a|
-      m += 1 if a.is_valid?
+      m += 1 if a.answered?
       m
     end
     num_questions == num_answers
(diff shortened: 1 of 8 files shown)


app/models/plan.rb
```diff
@@ -486,11 +486,26 @@ def authors
...
-    Answer.where(id: answers.map(&:id))
-          .includes(:question_options, question: :question_format)
-          .to_a
-          .sum { |answer| answer.is_valid? ? 1 : 0 }
+  def num_answered_questions(phase = nil)
+    if phase.present?
+      # If a phase was specified then return the number of
+      # answered questions within the context of the phase
+      question_ids = phase.sections.map do |s|
+        s.questions.pluck('questions.id')
```
xsrust: I think that the query would be more efficient if we defined the relationship between Phase and questions, or phase and answers, probably both.

app/models/plan.rb
```diff
@@ -486,11 +486,26 @@ def authors
...
+      end
+
+      Answer.where(id: answers.map(&:id))
+            .includes(:question_options, question: :question_format)
+            .where(question_id: question_ids.flatten)
+            .to_a
+            .sum { |answer| answer.is_valid? ? 1 : 0 }
```
xsrust: If the relationship between a phase and answers is defined, 

`phase.answers.where(plan_id: self.id).includes(:question_options, question: :question_format).to_a.sum`...

app/models/plan.rb
```diff
@@ -486,11 +486,26 @@ def authors
...
+    else
+      # Return the total number of answered questions across all phases
+      Answer.where(id: answers.map(&:id))
```
xsrust: Think it would be simpler to replace this with `self.answers`,  as we're finding the id's and then re-finding those answers.

Pre-fetching should still work on the association

briri:
makes sense @xsrust. Will update the model's relationships

briri:
Updated the models by adding the association shortcut as suggested. Also updated the name of `is_valid?` to `answered?` which seems more Rails-ish.

PR #2239, opened 2019-09-19
Fix for Accessibility issues with Tables (highlighted by UCOP electro…

johnpinto1:
Changes:
  - Added scope="col" to table header tags <th> that were missing them.
  - The Jquery dropdown menus in tables are closed when the table cell
containing it is out of focus.

(The issue the fontawesome sort icons are read as 'column # link'.
Add some screen reader only text. This issue was sorted by someone else
who added ".screen-reader-text" class to the sort_link_name() method in
app/controllers/concerns/paginable.rb.)

Fix for issue #1455.

**Current issue with dropdown when tabbing from cell with dropdown**: 
the dropdown remains visible
Selection_003

This commit closed the dropdown when out of focus.
Selection_001
Selection_002

Issue #1455, opened 2018-05-03
Accessibility issues for tables


briri:
The following accessibility issues were encountered when navigating across tables with a screen reader:
- The fontawesome sort icons are read as 'column # link'. Add some screen reader only text
- All of the `<th>` tags should have a `scope="col"` attribute
- The 'Actions' dropdown menus do not close when the user tabs off of the last link. They should be hidden when the menu loses focus

This feedback was provided from the UCOP electronic accessibility review. It was defined as having 'Medium' priority.

stephaniesimms:
I can't test the first two bullets w/o a screenreader. But the third bullet is not fixed. When I tab past the Actions menu it doesn't close.

briri:
@stephaniesimms I vote that we move this one into the backlog. Its likely something wrong with the bootstrap implementation of the dropdown menu or a JS conflict. I don't think we'll have time to resolve in this sprint though.

stephaniesimms:
sounds good @briri - moved to the backlog for now

johnpinto1:
The  fix "All of the <th> tags should have a scope="col" attribute" is not visible to most users, except those with assistive technologies.

The visible change is:
**Current issue with dropdown when tabbing from cell with dropdown**: 
the dropdown remains visible
![Selection_003](https://user-images.githubusercontent.com/8876215/65252498-f2ade680-daf0-11e9-897a-09a8bb7f54f1.png)

This commit closes the dropdown when out of focus.
![Selection_001](https://user-images.githubusercontent.com/8876215/65252693-5506e700-daf1-11e9-817e-82554b0f9f2e.png)
![Selection_002](https://user-images.githubusercontent.com/8876215/65252696-5506e700-daf1-11e9-91b5-b13fd61b4594.png)

sjDCC:
I can tab through to 'Actions' and click enter to open it. If I tab through all options it stays open. There's no navigate away button afaik so the only way to close currently is to select an action

![Capture](https://user-images.githubusercontent.com/6169940/65812141-d974fc00-e1ba-11e9-82d7-8b9a981fc5ae.PNG)

johnpinto1:
@sjDCC The issue above is addrerssed in  https://github.com/DMPRoadmap/roadmap/issues/1450
I have a fix. But need to fix tests for it as they break tests as currently written.

johnpinto1:
[...]

PR #2253, opened 2019-10-02
Fix for Accessibility issues with Tables (highlighted by UCOP electro…

johnpinto1:
…nic accessibility review)

Changes:
  - Added scope="col" to table header tags <th> that were missing them.
  - The Jquery dropdown menus in tables are closed when the table cell
containing it is out of focus.

(The issue the fontawesome sort icons are read as 'column # link'.
Add some screen reader only text. This issue was sorted by someone else
who added ".screen-reader-text" class to the sort_link_name() method in
app/controllers/concerns/paginable.rb.)

Fix for issue #1455.

I believe I fixed the issue causing the previous PR (that I closed #2239 yesterday)


johnpinto1:
@xsrus @briri Travis failing on
```
Vulnerabilities found!

The command "bundle exec bundle-audit check --update --ignore CVE-2015-9284" exited with 1.
```

app/javascript/utils/outOfFocus.js
```diff
@@ -0,0 +1,24 @@
+$(() => {
+  // This bit of code currently only closes dropdown menus in a table when the focus is no longer in the containing cell.
```
briri: Thanks for adding all of this commentary @johnpinto1 It will be helpful in the future!

briri:
Thanks @johnpinto1. Have you tried running `bundle update` to see if that rubzip can be updated to 1.3+? https://nvd.nist.gov/vuln/detail/CVE-2019-16892

Its used by the htmltoword gem but I haven't dug into how its being used. We don't zip files (that I know of) so it may be an ignorable issue. Its better to upgrade though if we can imo

xsrust:
Looks like a `bundle update rubyzip`, and then committing `Gemfile.lock` should  fix the problem.

It looks like it's been a while since [htmltoword](https://github.com/karnov/htmltoword) got an update

johnpinto1:
@xsrust  @briri Did a bundle update as suggested. Just need to wait for Travis report.

johnpinto1:
@xsrust @briri The `bundle update` did not fix Travis.

xsrust:
We got a new error for Travis, this  time complaining  about our cookie serialization.
I created #2255 to track the problem and will submit  a PR to development to update the insecure setting.
In the  meantime, I've re-started the remaining tests on the PR so it goes through the rails/js checks  against your  code

xsrust:
This is passing now that Development is patched.

Issue #2147, opened 2019-05-30
Bullet points - strange formatting 


magdalenadrafiova:
When user downloads the plan - the bullet points appear to be strange, see image below. 
Would you be able to look into what have triggered this and find a fixture for this? 

![bullet-points](https://user-images.githubusercontent.com/47178106/58630339-067dd480-82d7-11e9-83e4-78f41ca91efe.png)

johnpinto1:
Unable to replicate on Live or Test servers. @magdalenadrafiova  is going to contact person who raised issue. It might be a specific template.
[Test_Funny_characters.docx](https://github.com/DMPRoadmap/roadmap/files/3334885/Test_Funny_characters.docx)
[Test_Funny_characters.pdf](https://github.com/DMPRoadmap/roadmap/files/3334886/Test_Funny_characters.pdf)

magdalenadrafiova:
We have other admins now having the same issue. 

They are using Chrome and the characters appear after they simply ‘download’ the template as Word and open it (no typing etc) 

Let me know if we need to know something more .

johnpinto1:
@magdalenadrafiova Is it possible to get a copy of the downloaded templates.

magdalenadrafiova:
![tudelft-formatting](https://user-images.githubusercontent.com/47178106/60602815-725acd80-9dac-11e9-8ed0-d0537e0b2a3d.jpg)

briri:
@johnpinto1 I would suggest looking at that particular plan's answer in the database to see if there is any odd formatting (e.g. maybe the user did a copy-paste from Word that had some extra HTML markup). 

I did a quick test in DMPTool and could not replicate the issue so I think it may be related to that specific template/plan.

johnpinto1:
@magdalenadrafiova and @briri  for that input. Will check.

dsisu:
Hi, the .pdf downloads correctly, the word doc has the funny characters. John is looking at it now.

johnpinto1:
The Word docx @dsisu provided is attached
[Some_TU_Delft_test.pdf](https://github.com/DMPRoadmap/roadmap/files/3358534/Some_TU_Delft_test.pdf)
[Some_TU_Delft_test.docx](https://github.com/DMPRoadmap/roadmap/files/3358535/Some_TU_Delft_test.docx)

I can read the Word **docx** correctly on a Linux machine, but the same **docx** in Windows replaces bullet points with &#8713; characters

[...]

PR #2259, opened 2019-10-07
 Fix for format issues for bullet points in Plan docx exports. This

johnpinto1:
updates a previous fix. This time we only apply the change suggested by
@benjaminfaure to docx.

    The fix was suggested by @benjaminfaure: https://github.com/DMPRoadmap/roadmap/issues/2147#issuecomment-535935932

    Changes:
    Replaced <li> tag with <ol> in _plan.erb as suggested by @benjaminfaure
    that <li> tag being transformed as € sign in the Word export.

Fix for issue #2147

### DIFF
--- a/app/views/shared/export/_plan.erb
+++ b/app/views/shared/export/_plan.erb
@@ -35,7 +35,8 @@
                 <% if !@public_plan && @show_sections_questions%>
                   <%# Hack: for DOCX export - otherwise, bold highlighting of question inconsistent. %>
                   <% if local_assigns[:export_format] && export_format == 'docx' %>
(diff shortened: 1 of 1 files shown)


johnpinto1:
@xsrust and @briri I just ensured @benjaminfaure  fix only applies to **docx**. The fix suggested by @xsrust https://github.com/karnov/htmltoword/issues/73 would require we update the **htmltoword** gem which I am not to clear how to set about, it would need a local gem.

johnpinto1:
Closing this PR after discussions with @xsrust. He suggested a way to avoid this approach to fixing issue.

PR #2262, opened 2019-10-14
Fix issue with saving changes to a template's visibility

### LINKED ISSUE #1712: Can't remove checkbox for internal template use only
For funder/orgs (e.g. DCC, USGS) once a template has been marked for internal use only, I can't remove the status. Each time I uncheck the box and click Save, I receive a message that my change was successfully saved, but when I navigate away from the template details page and then return to it I find that the box is still checked for internal use only (screenshot). Correct behavior is to enable admins to remove the internal status.

screen shot 2018-07-11 at 3 04 18 pm

briri:
Fixes #1712 .

Updated logic to properly display and save changes to a Template's visibility. To test/replicate the issue, make an org a 'Funder' and an 'Org' org_type. Then edit one of their templates. You should see the visibility checkbox. Try to check/uncheck the box and save.


briri:
closing out. submitted PR to wrong branch

Issue #2251, opened 2019-09-30
Downloads with long names of unusual non-latin characters don't have a filename extension


magdalenadrafiova:
Issue: Downloads with long names of unusual non-latin characters don't have a filename extension

Description: Can't open the file locally until filename extension is added manually

Desired behaviour: Make sure extension is added, by truncating the name if necessary

sjDCC:
This seems to be working fine now. @magdalenadrafiova can you validate that the test I did matched the issue you encountered. if so, please close out.

![Capture](https://user-images.githubusercontent.com/6169940/68217903-575acd00-ffdb-11e9-9d2e-b0fba2a02c01.PNG)

![Capture](https://user-images.githubusercontent.com/6169940/68218083-ac96de80-ffdb-11e9-836f-f0038d0b512d.PNG)

nicolasfranck:
Should we even try to use the plan title as the file name?
Why not simply use something like `plan_<plan_id>_phase_<phase_id>.pdf`?

It is more opaque, but at least more condense, readable, and it leaves
nothing up to surprise (maybe someone will add emoticons to their title?)

cf. https://github.com/DMPRoadmap/roadmap/blob/cabe6d3201c76f51a5a9ac278bfcd2c8e694fdf2/app/controllers/plan_exports_controller.rb#L101

Users already refer to the plan url, which includes the id.

PR #2265, opened 2019-10-14
added some filename sanitization and now limiting filenames to 100 ch…

### LINKED ISSUE #2249: Dowloading a docx export of a plan with non-latin chars gives a crazy filename
Issue: Dowloading a docx export of a plan with non-latin chars gives a crazy filename

Description: Unicode not being used? 

Desired behaviour: Filename should fully match the plan title

briri:
Fixes #2249 and #2251.

- Added the Zaru gem which handles file name sanitization. 
- Updated regex so that multiple spaces convert to a single underscore (e.g. 'foo   bar' becomes 'foo_bar' instead of 'foo___bar')
- Limit the length of the filename to 100 characters. Some Windows versions have a `MAX_PATH` setting that only allows 255 characters (for the filename + path). 100 should provide the user with enough of the title to provide context while allowing for the location they save it to to be pretty deeply nested. https://stackoverflow.com/questions/265769/maximum-filename-length-in-ntfs-windows-xp-and-windows-vista

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -154,6 +154,9 @@ gem 'htmltoword'
 # A feed fetching and parsing library (http://feedjira.com)
 gem 'feedjira'
 
+# Filename sanitization for Ruby. This is useful when you generate filenames for downloads from user input
+gem 'zaru'
+
 # ------------------------------------------------
 # INTERNATIONALIZATION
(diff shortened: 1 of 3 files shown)


briri:
lol. perhaps just covering himself in the event that anyone has problems. We can look around for another.

Issue #2240, opened 2019-09-20
Improve admin view of users


xsrust:
- [Org, Super] Admins should be able to view, and visit, a list of user's plans
- [Org, Super] Admins should be able to update a user's department

johnpinto1:
Code for review. Some screenshots of views:


**Super Admins' View:**
![Selection_029](https://user-images.githubusercontent.com/8876215/67292149-89980500-f4da-11e9-9b00-aafbb3607dfd.png)

**Org Admins' Views:**
![Selection_028](https://user-images.githubusercontent.com/8876215/67292257-b2b89580-f4da-11e9-9479-bd47681f601b.png)
![Selection_027](https://user-images.githubusercontent.com/8876215/67292259-b3512c00-f4da-11e9-9287-949da6c507e8.png)
![Selection_026](https://user-images.githubusercontent.com/8876215/67292262-b3e9c280-f4da-11e9-81f3-c765dcdb1f7b.png)

sjDCC:
These look great, thanks @johnpinto1 

Two small improvement requests.
1. Can the Dept box be extended to align with the other fields
![align](https://user-images.githubusercontent.com/6169940/68217584-c683f180-ffda-11e9-9928-cb23f318acbc.PNG)

2. Can the number of plans column in the 'users' table link through to where admins can see the list of plans. They may not spot this otherwise
![Plan-link](https://user-images.githubusercontent.com/6169940/68217598-cb48a580-ffda-11e9-9742-62c76d4a3f3d.PNG)

johnpinto1:
@sjdcc @xsrust Work in progress on this 
![Selection_050](https://user-images.githubusercontent.com/8876215/69154066-3dd77c00-0ad7-11ea-8576-70a79aa58205.png)
![Selection_051](https://user-images.githubusercontent.com/8876215/69154087-4465f380-0ad7-11ea-90ed-870d6320d1da.png)

Have a query with @xsrust about plans numbers that needs to be sorted before I push.

magdalenadrafiova:
Could we make the department or school to expand like the other fields @xsrust @briri @raycarrick-ed  
![image](https://user-images.githubusercontent.com/47178106/69872678-fc508900-12ad-11ea-84b8-3c48fb2ad762.png)

the plans to be 'clickable' works well

xsrust:
Odd.  This looks fixed in John's PR (and screenshots) from a few days ago.

[...]

PR #2266, opened 2019-10-22
Updated functionality for Admin View of Users.

johnpinto1:
Changes:
 - for Super Admins: Added user department and plans.
 - for Org Admins:
        (1) The user view now has a clickable link for opening up
            a view of a user's profile and plans.
        (2) In a user's profile view the Department if it exists for the
            organisation is editable. The Organisation must have
            departments in systerm to avail of this functionality.

Fix for #2240.

**Super Admins' View:**
Selection_029

**Org Admins' Views:**
Selection_028
Selection_027
Selection_026

### DIFF
--- a/app/controllers/org_admin/users_controller.rb
+++ b/app/controllers/org_admin/users_controller.rb
@@ -1,0 +1,50 @@
+# frozen_string_literal: true
+
+module OrgAdmin
+
+  class UsersController < ApplicationController
+
+    after_action :verify_authorized
+
+    def edit
+      @user = User.find(params[:id])
+      authorize @user
+      @departments = @user.org.departments.order(:name)
+      @plans = Plan.active(@user).page(1)
+      render "org_admin/users/edit",
+             locals: { user: @user,
+                       departments: @departments,
+                       plans: @plans,
+                       languages: @languages,
+                       orgs: @orgs,
+                       identifier_schemes: @identifier_schemes,
+                       default_org: @user.org }
+    end
+
+    def update
+      @user = User.find(params[:id])
+      authorize @user
+      @departments = @user.org.departments.order(:name)
+      @plans = Plan.active(@user).page(1)
+      # Replace the 'your' word from the canned responses so that it does
+      # not read 'Successfully updated your profile for John Doe'
+      topic = _("profile for %{username}") % { username: @user.name(false) }
+      if @user.update_attributes(user_params)
+        flash.now[:notice] = success_message(@user, _("updated"))
+      else
+        flash.now[:alert] = failure_message(@user, _("update"))
+      end
+      render :edit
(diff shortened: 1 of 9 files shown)


app/controllers/org_admin/users_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+  class UsersController < ApplicationController
+
+    after_action :verify_authorized
+
+    def edit
+      @user = User.find(params[:id])
+      authorize @user
+      @plans = Plan.active(@user).page(1)
+      render "org_admin/users/edit",
+             locals: { user: @user,
```
briri: Don't bother changing your PR here @johnpinto1, I just wanted to point out that this is a pattern that got introduced some time ago and we've been carrying it forward. There's no need for us to assign these `locals` when we've already defined instance variables. You can just reference `@user` directly in the view.

I'll add it to the agenda to discuss in our next dev call.

app/controllers/paginable/plans_controller.rb
```diff
@@ -49,4 +49,25 @@ def org_admin
     )
   end
 
+  # GET /paginable/plans/org_admin/:page
+  def org_admin_other_user
+    @user = User.find(params[:id])
+    authorize @user
+    # render "super_admin/users/edit",
```
briri: go ahead and remove this commented code

app/views/org_admin/users/edit.html.erb
```diff
@@ -0,0 +1,80 @@
+<% title _('Editing profile for %{username}') % { username: @user.name(false) } %>
```
briri: this is defined but not used. either reference it on line 5 or remove this one

app/views/super_admin/users/edit.html.erb
```diff
@@ -35,6 +35,20 @@
...
 
+      <% departments = @user.org.departments.order(:name) %>
+      <% if departments.count > 0 %>
+      <div class="form-group col-xs-8">
+        <% dept_id = @user.department.nil? ? -1 : @user.department.id  %>
+        <%= f.label(:department_id, _('Department or school'), class: 'control-label') %>
+        <%= select_tag(:department_id,
+            options_from_collection_for_select(departments, "id", "name", dept_id),
+            include_blank: true,
+            disabled: departments.count === 0,
```
briri: `departments.count === 0` should never happen since we have the `if departments.count > 0` above

briri: also, consider using `length` or `any?` instead of count. Count is slower. 

It would also be better to load this in the controller and then reference it via `@departments`

johnpinto1:
@briri Thanks for these helpful comments. I have addressed all except the first as suggested.

Issue #2091, opened 2019-04-17
Allow admins to export their own templates


sjDCC:
Admins want to be able to produce a doc/pdf download of their own templates like we make visible for funders on the 'Public templates' page. This should include the questions and their guidance. The same code could be reused but applied to the admin interface.

Admin users should see an option to download their templates. This could be presented on the history table as there may be a need to produce a historical version for queries about what requirements were in place when.

![Picture1](https://user-images.githubusercontent.com/6169940/65909532-cb90c800-e3c0-11e9-86f9-cf4fefdb1e6a.png)
We suggest this is a 'download' option under template edit

johnpinto1:
Fix as a PR for review. Some screenshots of **Org Admin** *Templates* and *Template History* pages. 
Note: The files have version appended, e.g., **My_Template_v0.docx** or **My_Template_v5.pdf**.

![Template_export](https://user-images.githubusercontent.com/8876215/68022107-3d4b8280-fc9b-11e9-9d09-dce483bd2849.png)
![Selection_040](https://user-images.githubusercontent.com/8876215/68022109-3d4b8280-fc9b-11e9-90d1-1667eb21ce07.png)

sjDCC:
This works well and I particularly like having the options on the history table.

I wonder if we should just have a single 'Download' option under the action list so it's in line with the other entries and just give either docx or PDF as default. People can get the full choice under history. Thoughts @mariapraetzellis @magdalenadrafiova ?

magdalenadrafiova:
I agree with Sarah to just keep the download option and remove the icon next to it if possible - to keep all consistent

xsrust:
In the case of one option, I'd lean towards PDF as the default rather than docx, but that's just personal preference.

johnpinto1:
Updated code as suggested with single **Download** option which is a *pdf* export
![Selection_045](https://user-images.githubusercontent.com/8876215/68379714-4da1a880-0146-11ea-8560-b919a65d20cc.png)

magdalenadrafiova:
When I go into download my plan I do not see the option to download a plan option - I tested dmponline-test under uni of manchester trying to download my manchester uni template

magdalenadrafiova:
![image](https://user-images.githubusercontent.com/47178106/69868645-33b93880-12a2-11ea-86fd-993a6206a208.png)
works well - I was able to download as pdf closing this

PR #2274, opened 2019-11-07
Updated Org admins template table Action menu.

johnpinto1:
Change:
  - Re-worded "Export as pdf" to "Download".
  - Removed "Export as docx".

Fix for #2091.

Selection_045


johnpinto1:
@briri @xsrust My bad those files generated when you run **Rspec** were accidentally added to commit as I had run **Rspec** . I have removed them.

May be we should consider excluding *.docx, *.txt and *.csv from root directory in **.gitignore**.

johnpinto1:
@briri @xsrust Updated the gem for **brakeman** because Travis was failing.

PR #2286, opened 2019-11-26
fixed issue with archiving users

### LINKED ISSUE #1489: Delete user and plan and comment option
Users occasionally ask for accounts to be deleted. It would be good to add an option to super-admin to do this - and maybe even on the 'edit profile' page so users can do it themselves? This would go beyond 'deactivating' users to:
- delete email (replace with a generic stem email?)
- remove name
- remove org affiliation

We need to remove any personal and identifiable data for GDPR but don't want a hard delete as this could leave orphaned plans or affect other user data.

And co-owned plans shouldn't be deleted as others still need access to them, and potentially the institution too?

Notes / comment that the user has added could be deleted I think.

@stephaniesimms @dsisu - can we define what we want in terms of functionality?

briri:
Fixes #1489 .

Changes proposed in this PR:
- Fixed 'Archive' link/button to go to the correct controller/action
- Moved the default email suffix to the branding.yml file @xsrust you will need to add this value to your copy
- Updated the 'archive' method on the User model to update the fields (was not actually saving the changes without the `self.` bit

### DIFF
--- a/config/branding.yml.sample
+++ b/config/branding.yml.sample
@@ -2,7 +2,7 @@ defaults: &defaults
   legal_entity: 'the University of Edinburgh, University of Glasgow and the University of California'
 
   # Warning: The abbreviation here should match the org.abbreviation value registered in your database!
-  organisation:
+  organisation: &organization_defaults
     name: 'Curation Center'
     abbreviation: 'CC'  # This value is used to identify the default guidance that get auto-selected when a new plan is created
     url: 'https://github.com/DMPRoadmap/roadmap/wiki'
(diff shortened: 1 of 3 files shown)


config/branding.yml.sample
```diff
@@ -32,7 +32,7 @@ defaults: &defaults
     #          Comment this line out if you want to use DMPRoadmap's built-in contact form.
     contact_us_url: 'https://somewhere.com/contact-us'
 
-  application:
+  application: &application_defaults
```
briri: Added these variable names so you can do things like:
```yaml
development:
  <<: *defaults
  application:
    <<: *application_defaults
    name: 'DMPTool-Dev'
```
Without the `<<: *application_defaults` line, a check of `Rails.configuration.branding[:application].inspect` shows that it ONLY contains the `name:` param

xsrust: Oh cool!
That's much better than the way I was managing those differences, and it's nice to have the template in there for everybody.
Thanks @briri

Issue #2217, opened 2019-08-23
Full Text Plans API Filtering


xsrust:
When querying the Full-Text plans API `api/v0/plans`, users would like to be able to filter the content.

Suggested filterable fields are:
1) Dates (creation and update)
2) Users (scope to a list of users's plans)
3) Funders (only Plans on a specific funder's templates) 
4) ID (pass a single, or list of IDs to export)

briri:
merged in with latest release

PR #2293, opened 2019-12-03
Feature/plans api filtering

xsrust:
Small bugfix for templates API (was throwing an error previously)

Added filtering on 
- User[]
- Template[]
- plan[]
- Creation/update dates

### DIFF
--- a/app/controllers/api/v0/plans_controller.rb
+++ b/app/controllers/api/v0/plans_controller.rb
@@ -61,13 +61,13 @@ class Api::V0::PlansController < Api::V0::BaseController
     if params[:per_page].present? && params[:per_page].to_i > Rails.configuration.branding[:application][:api_max_page_size]
       params[:per_page] = Rails.configuration.branding[:application][:api_max_page_size]
     end
-    @plans = paginate @user.org.plans.includes( [ {roles: :user}, {answers: :question_options} ,
+    @plans = @user.org.plans.includes( [ {roles: :user}, {answers: :question_options} ,
       template: [ { phases: { sections: { questions: [:question_format, :themes ] } } }, :org]
       ])
-    
+
     # Filter on list of users
     user_ids = extract_param_list(params, "user")
-    @plans = @plans.where(roles: {user_id: user_ids}) if user_ids.present?
+    @plans = @plans.where(roles: {user_id: user_ids, access: Role.bit_values(:editor)}) if user_ids.present?
     # filter on dates
     if params["created_after"].present? || params["created_before"].present?
       @plans = @plans.where(created_at: dates_to_range(params,"created_after","created_before"))
@@ -81,6 +81,8 @@ class Api::V0::PlansController < Api::V0::BaseController
     # filter on id(s)
     plan_ids = extract_param_list(params, "plan")
     @plans = @plans.where(id: plan_ids ) if plan_ids.present?
+    # apply pagination after filtering
+    @plans = paginate @plans
     respond_with @plans
   end
 
--- a/app/controllers/api/v0/plans_controller.rb
+++ b/app/controllers/api/v0/plans_controller.rb
@@ -61,10 +61,44 @@ class Api::V0::PlansController < Api::V0::BaseController
     if params[:per_page].present? && params[:per_page].to_i > Rails.configuration.branding[:application][:api_max_page_size]
       params[:per_page] = Rails.configuration.branding[:application][:api_max_page_size]
     end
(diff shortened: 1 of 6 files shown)


xsrust:
Addresses #2217

app/controllers/api/v0/plans_controller.rb
```diff
@@ -61,10 +61,44 @@ def index
...
     end
-    @plans = paginate @user.org.plans.includes( [ {answers: :question_options} ,
-      template: [ { phases: { sections: { questions: :question_format } } }, :org]
+    @plans = paginate @user.org.plans.includes( [ {roles: :user}, {answers: :question_options} ,
+      template: [ { phases: { sections: { questions: [:question_format, :themes ] } } }, :org]
       ])
+    
+    # Filter on list of users
+    user_ids = extract_param_list(params, "user")
+    @plans = @plans.where(roles: {user_id: user_ids}) if user_ids.present?
```
briri: do we want to include all roles here (e.g. commenter and reviewer)?

app/controllers/api/v0/plans_controller.rb
```diff
@@ -61,10 +61,44 @@ def index
...
+    # filter on dates
+    if params["created_after"].present? || params["created_before"].present?
+      @plans = @plans.where(created_at: dates_to_range(params,"created_after","created_before"))
+    end
+    if params["updated_after"].present? || params["updated_before"].present?
+      @plans = @plans.where(updated_at: dates_to_range(params,"updated_after","updated_before"))
+    end
+    # filter on funder (dmptemplate_id)
+    template_ids = extract_param_list(params, "template")
```
briri: might be more efficient to process template and plan id filtering before the dates

app/controllers/api/v0/plans_controller.rb
```diff
@@ -61,10 +61,44 @@ def index
...
+    list = params.fetch(attribute+"[]", [])
+    val = params.fetch(attribute, [])
+    list << val if val.present?
+    list
+  end
+
+  # takes in the params hash and converts to a date-range
+  def dates_to_range(hash,start,stop)
+    today = Date.today
+    start_date = Date.parse(hash.fetch(start, today.prev_month.to_date.to_s))
```
briri: Would be useful to trap any exceptions thrown by the `Date.parse` here and either send the error back to the caller or default to a specific date like one month ago or something

app/controllers/api/v0/templates_controller.rb
```diff
@@ -25,7 +25,7 @@ def index
                                    .where(org_id: @user.org_id, published: true)
                                    .where.not(customization_of: nil)
 
-    Template.published.order(:org_id, :version).each do |temp|
+    published_templates.order(:org_id, :version).each do |temp|
```
briri: oh wow. great catch.

app/controllers/api/v0/plans_controller.rb
xsrust: I think it should be lazy-evaluating, so activerecord should apply all stages of filtering at the same time.

app/controllers/api/v0/plans_controller.rb
xsrust: We should add that in and make it apply both here and in the statistics_controller, as I've copied this method over from there.

app/controllers/api/v0/plans_controller.rb
xsrust: Reviewer is no longer a role,  But we do probably want to keep this to editors/administrators/creators.  

I have added a helper-function to the role model to allow us to extract the numbers which correspond to a bit_value

briri: ah right. forgot we had removed the reviewer role. thanks for adding the helper method

app/models/role.rb
```diff
@@ -82,6 +82,23 @@ class Role < ActiveRecord::Base
...
+  ##
+  # Get the integer values that correspond to a given access flag
+  # Convert into a condition, take the numerical half, remove formatting
+  # split on commas, and then convert each to an integer
+  #
+  # access - The symbol corresponding to the user's access, i.e. :editor
+  #
+  # Returns [Integer]
+  def self.bit_values(access)
+    Role.send(access.to_s + "_condition").split(' in ').last[1...-2].split(',').map(&:to_i)
```
briri: Try `Role.send(:chained_flags_values, 'access', access)` instead

PR #2298, opened 2019-12-03
Some minor updates for translations

briri:
@xsrust I found these while checking out the pt_BR changes (pre that external PR that's out there).

We had a few issues with some missing translations that are still there.

If these look good, merge them and then run the script that extracts all of the latest strings into the pot/po files when you create the release?

We can merge the external PR once development is in master.

### DIFF
--- a/app/controllers/roles_controller.rb
+++ b/app/controllers/roles_controller.rb
@@ -77,7 +77,7 @@ class RolesController < ApplicationController
       # rubocop:disable LineLength
       render json: {
         code: 1,
-        msg: _("Successfully changed the permissions for #{@role.user.email}. They have been notified via email.")
+        msg: _("Successfully changed the permissions for %{email}. They have been notified via email.") % { email: @role.user.email }
       }
       # rubocop:enable LineLength
     else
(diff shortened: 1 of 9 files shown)


app/controllers/roles_controller.rb
```diff
@@ -77,7 +77,7 @@ def update
       # rubocop:disable LineLength
       render json: {
         code: 1,
-        msg: _("Successfully changed the permissions for #{@role.user.email}. They have been notified via email.")
```
briri: interpolation here was causing the translation to not get picked up

app/views/org_admin/annotations/_form.html.erb
```diff
@@ -1,6 +1,7 @@
 <fieldset class="annotation_fields" id="<%= unique_dom_id(f.object, "fields") %>">
   <div class="form-group col-md-10">
-    <%= f.label(:type, f.object.type.humanize, class: "control-label") %>
```
briri: the text here is derived from the object type and not getting properly handled by get text

Issue #2314, opened 2019-12-12
[Bug] Error when accessing user's plan as an org admin


benjaminfaure:
Hello everyone, 

I found some bugs while accessing a user's plans as an **org admin** : 
 

1. On the **user profile**, when using the pagination in the **plans list**, the following error appear in the console : 
```
Started GET "/paginable/plans/org_admin_other_user/2?id=1&sort_direction=desc&sort_field=plans.updated_at" for 127.0.0.1 at 2019-12-11 16:08:14 +0100
Processing by Paginable::PlansController#org_admin_other_user as JS
  Parameters: {"id"=>"1", "sort_direction"=>"desc", "sort_field"=>"plans.updated_at", "page"=>"2"}
Completed 500 Internal Server Error in 41ms (ActiveRecord: 4.2ms)
NoMethodError - undefined method `org_admin_other_user?' for #<UserPolicy:0x00005570ecef3dc0>:
  app/controllers/paginable/plans_controller.rb:56:in `org_admin_other_user
briri:house_with_garden: 4:51 PM
```

2. Not exactly a bug, but the plans pulled from the database differs wether you use the pagination or not on the **user profile**. 
When you display the User Profile, all the user's plans are pulled : 
https://github.com/DMPRoadmap/roadmap/blob/f807b5f49ab345eec2b6192039a8fc1cf3e8d3e0/app/controllers/org_admin/users_controller.rb#L13
When you use the pagination, only the plans with public or org visibility are pulled : 
https://github.com/DMPRoadmap/roadmap/blob/f807b5f49ab345eec2b6192039a8fc1cf3e8d3e0/app/controllers/paginable/plans_controller.rb#L61
The `org_admin_other_user` function is called only when using the pagination, here in the code : 
https://github.com/DMPRoadmap/roadmap/blob/f807b5f49ab345eec2b6192039a8fc1cf3e8d3e0/app/views/org_admin/users/plans.html.erb#L12

If you want a similar working example, the org admin plan list works the same way : 
https://github.com/DMPRoadmap/roadmap/blob/f807b5f49ab345eec2b6192039a8fc1cf3e8d3e0/app/views/org_admin/plans/index.html.erb#L47

I'll try to look into the first bug, I'll keep you updated if I find something.

benjaminfaure:
The first issue seems to be resolved by adding 
```
def org_admin_other_user?
    @user.is_a?(User)
  end
```

to the `policies\user_policy.rb` file

[...]

PR #2322, opened 2019-12-13
Fixed issue #2314

benjaminfaure:
Fixes issue #2314 
- Added `org_admin_other_user?` policy in UserPolicy
- Plans displayed on the user profile, using pagination should be all the user's plan. Changed pulled plans from `organisationally_or_publicly_visible` to `active`.


benjaminfaure:
If that haven't been changed, the org admin can already access a list of plans created by the org users (on the `org_admin/plans` route) .
I thought that it would be better if the behaviour stayed coherent throughout the application. 

(We changed this behaviour in DMP OPIDoR, the org admin can't see the private plans in the list)

PR #2334, opened 2019-12-17
Update the Travis

briri:
Made the bundle-audit and brakeman checks conditional so that they only run when the PR/Push is made to the master branch

Also noticed an error that the version of bundler that Travis was trying to use was not in tune with the one defined in our Gemfile.lock, so added a line to install that version of the gem.


briri:
@raycarrick-ed this PR sets a condition on the bundle-audit and brakeman tests that we discussed last Friday. It sets them up to only run against PRs to master. That way a gem security vulnerability won't hold up dev any longer.

PR #2342, opened 2020-01-08
check for agreement to T&Cs was wrong

### LINKED ISSUE #2341: Agreement to T&Cs not enforced

**Expected behaviour:**
When creating an account, if you don't tick the "I accept the terms and conditions" checkbox you should get an error.

**Actual behaviour:**
Just goes ahead and creates the account.

**Steps to reproduce:**
Create new account and don't check the T&Cs box.

raycarrick-ed:
Fixes #2341  .

Changes proposed in this PR:
- change to test for terms and conditions checkbox

### DIFF
--- a/app/controllers/registrations_controller.rb
+++ b/app/controllers/registrations_controller.rb
@@ -53,7 +53,7 @@ class RegistrationsController < Devise::RegistrationsController
       end
     end
 
-    if params[:accept_terms].to_s == ""
+    if params[:user][:accept_terms].to_s == "0"
       redirect_to after_sign_up_error_path_for(resource),
         alert: _("You must accept the terms and conditions to register.")
     elsif params[:user][:org_id].blank? && params[:user][:other_organisation].blank?
--- a/app/controllers/registrations_controller.rb
+++ b/app/controllers/registrations_controller.rb
@@ -53,7 +53,7 @@ class RegistrationsController < Devise::RegistrationsController
       end
     end
 
-    if params[:accept_terms].to_s == "0"
+    if params[:accept_terms].to_s == ""
       redirect_to after_sign_up_error_path_for(resource),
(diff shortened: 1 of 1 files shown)


briri:
Hi @raycarrick-ed,

The tests are failing with this change and when I look at the params sent back on form submission, the `accept-terms` param is `0`:
```ruby
{ "user"=>{
  "firstname"=>"Test", 
  "surname"=>"User", 
  "email"=>"tester@example.org", 
  "org_name"=>"Example University (EU) ", 
  "org_id"=>"33", 
  "other_org_id"=>"0", 
  "other_org_name"=>"Non Partner Institution",
  "other_organisation"=>"", 
  "password"=>"[FILTERED]", 
  "accept_terms"=>"0"
}
```
I tried it in both Firefox and Safari to see if it was maybe browser specific (could still be I guess, I didn't try IE or Chrome). Interestingly all the Rails docs indicate that the checkbox shouldn't even appear in the params if it hasn't been checked 🤷‍♂ 

Anyway, maybe try something like: `if ["", "0"].include?(params[:accept_terms].to_s)`

raycarrick-ed:
Took me a ridiculous time to figure this out. I couldn't make sense of the values coming in until I realized that the accept_terms is inside user and so the original code was correct except for that i.e.
if params[:user][:accept_terms]
in place of
if params[:accept_terms]

briri:
ah yeah. That's some confusing Rails magic. Most HTML forms have an associated model and so all of the form elements are wrapped within the context of the model type (user in this case).

We haven't yet updated that controller to use 'strong parameters' (we've gotten to most others) and is something we will need to do for Rails 5/6. It makes this situation a bit clearer I think since it 'requires' a model and then 'permits' specific params: https://edgeapi.rubyonrails.org/classes/ActionController/StrongParameters.html

PR #2356, opened 2020-01-20
Conditional questions

### LINKED ISSUE #1772: Allow conditional fields / question logic / ordering
The question of conditional fields has been raised as a feature request several times (i.e. If no, skip to q6). I'd like to consider adding it to the list and scoping out how much work it is to implement.

Thoughts @stephaniesimms @briri @xsrust @Bodacious 

Has there been demand for this in the USA?

Mandatory questions were requested in the past too but there had been issues with this in DMPTool so we may want to keep that out of scope

raycarrick-ed:
Fixes #1772  .

Changes proposed in this PR:
- add conditional questions. Almost ceratinly not in final form (see comments on ticket) but probably better to get out and get feedback and then revisit.


briri:
looks like something weird happened with this PR @raycarrick-ed.

Have a lot of old commits and no conditional question changes.

xsrust:
It kindof looks like you may have committed the DMPonline branding, or atleast that looks like the changes vs development  There's also merge- msgs from Jimmy, so I thinnk it may have pulled in our Master or Development branch from DCC

PR #2357, opened 2020-01-21
added a 'context' field to IdentifierScheme and some scopes

briri:
Trying to break this maDMP ROR/Fundref work into reasonable chunks to make the code review a bit easier.

**Note that this is a PR to a parallel branch that I will be keeping in sync with development**

Part One adding a context to the identifier_schemes table to make it easier to segregate what identifiers are getting loaded at any given time:
- Added `identifier_schemes.context` with 2 contexts (for now). It may make sense to store the plan DOIs in this way later on
- Added some scopes to OrgIdentifier, IdentifierScheme and UserIdentifier to make it easier to get at a specific identifier (e.g. get me the shib id for a user)
- Updated code to use these new scopes
- Added tests for the new field and scopes
- Added a v2_2_0 rake task and 2 sub tasks that: Add the ROR and FundRef identifier schemes and then sets the new 'context' field.

The new 'by_name' scope doesn't introduce anything groundbreaking. Its an attempt to ensure that the correct scheme is found for any installations out there that might not have all lower case names. We were forcing the '.downcase' in a lot of areas but not everywhere.

### DIFF
--- a/.github/workflows/mysql.yml
+++ b/.github/workflows/mysql.yml
@@ -81,7 +81,9 @@ jobs:
 
     # Setup the database
     - name: 'Setup Test DB'
-      run: bundle exec rake db:setup RAILS_ENV=test
+      run: |
+        bundle exec rake db:migrate RAILS_ENV=test
+        bundle exec rake db:setup RAILS_ENV=test
 
     # Compile the assets
     - name: 'Compile Assets'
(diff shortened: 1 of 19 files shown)


briri:
not sure why the travis Push test is failing. All Github workflow tests are passing

briri:
removing this one in favor of new idea tifieas polymorphic table

PR #2358, opened 2020-01-21
Part 2 - maDMP ROR/Fundref - Moving from is_other to orgs.managed = true

briri:
Trying to break this maDMP ROR/Fundref work into reasonable chunks to make the code review a bit easier.

**Note that this is a PR to a parallel branch that I will be keeping in sync with development**

Part Two adding a managed flag to the orgs table. The intention is for this flag to eventually replace our 'is_other' logic. For now I am just adding it. The 'is_other' stuff will be removed in a subsequent step:
- Added `orgs.managed` with which defaults to false. The system will auto-create Org records when a user selects or shib sends us an Org we have never seen before.
- Added some scopes to Org to work with this new flag
- Renamed an old 'managing_orgs' scope that finds the 'default' Org defined in the branding.yml
- Updated admin nav bar and `User.can_org_admin?` to check that the user's org is managed
- Added managed to the Super Amin Orgs pages and the ability for the Super to change the value
- Added a new subtask to the `v2_2_0` that defaults all existing Org records to 'managed = true'

The idea behind this will be that when a user changes their Org via the edit profile page, Shibboleth sends us a new Org, or a user creates an account and enters a new Org then the system will auto-create an Org record. That org would appear in the search results for subsequent users.

By default, this org would have no admin and NOT have access to any of the features that our admins have. Basically anywhere that we do a 'is_other' check would become a 'managed?' check

By having an Org record we will better be able to handle maDMP tasks for these users (assuming we have a ROR or FundRef of course) and would also allow for an Org Admin to come along afterward but still have the benefit of stats data and all of their user's plans


app/models/org.rb
```diff
@@ -135,11 +139,18 @@ class Org < ActiveRecord::Base
             6 => :school,
             column: "org_type"
 
-  # Predefined queries for retrieving the managain organisation and funders
-  scope :managing_orgs, -> do
+  # The default Org is the one whose guidance is auto-attached to
+  # plans when a plan is created
+  def self.default_orgs
```
xsrust: Was there a reason to convert this from a scope to a function?

app/models/user.rb
```diff
@@ -241,9 +241,14 @@ def can_super_admin?
   #
   # Returns Boolean
   def can_org_admin?
-    self.can_grant_permissions? || self.can_modify_guidance? ||
-      self.can_modify_templates? || self.can_modify_org_details? ||
-      self.can_review_plans?
+    return true if can_super_admin?
```
xsrust: I think that this makes sense on the nose, but maybe calls for more thought about our role structure.

I like the idea of a super-admin implicitly having org_admin permissions, but I wonder how many other places in the code we check for the granular permission, rather than the categorical (i.e. org/super_admin).

lib/tasks/upgrade.rake
```diff
@@ -696,6 +701,11 @@ namespace :upgrade do
     Role.reviewer.destroy_all
   end
 
+  desc "Sets the new managed flag for all existing Orgs to managed = true"
+  task default_orgs_to_managed: :environment do
+    Org.all.update_all(managed: true)
```
xsrust: Should we have an exception for the `other org` or is that invalidated by your other ROR PR?

Issue #119, opened 2017-03-01
Allow template owners to make questions mandatory and set routing


sjDCC:
Some template owners would like to make certain questions compulsory or set routing through templates e.g. if yes, skip to section X

Main requirements here come from ZonMw, but several have asked for mandatory questions

stephaniesimms:
USGS began using the mandatory question option on DMPTool templates. We can discuss whether this should be included in MVP.

Yu-Chieh-Henry-Yang:
@briri @stephaniesimms 
Are we looking forward to implement the feature for template owner to make a question mandatory soon? (Or maybe this feature is already built in?)
If not, how would you suggest one go about building such feature?

sjDCC:
hi @Yu-Chieh-Henry-Yang This is just a suggestion for now, but something we'd certainly be interested in as users have requested it. If you are considering adding such a feature please let us know so we can collaborate and bring this in. It's great to get code contributions from externals like you

magdalenadrafiova:
@xsrust can we also close this?

PR #2359, opened 2020-01-22
Conditional questions

### LINKED ISSUE #1772: Allow conditional fields / question logic / ordering

### LINKED ISSUE #2304: Question Edit bug in template creation
Expected: should be able to delete a question even if the template is published.
Actual: If template is published when you edit the question you never get a delete button.

To reproduce:
    Create a template.
    Publish it.
    Edit it. Go to a question and choose "Edit" for that question.

The question changes to an editable form and at the bottom you have 2 buttons: "Edit" and "Cancel"
If you choose "Cancel" the question is removed from the screen but if you refresh it reappears.

Needs fixed and tested.

Notes:
Looking at the code, the critical bit is in
app/views/org_admin/questions/_form.html line 97 ish
where the buttons are drawn

<% if question.id.present? && !question.section.phase.template.published? %>

so the first clause is checking is it's been saved. If it has we present a Delete button and a Cancel, if not we just have a Cancel and only need to remove from the screen.

The second clause is the problem. If the template has been published we will never get the delete button and so never be able to delete the question. I *think* this might be a leftover from before versioning?? If we remove this clause then the delete button appears and the delete action triggers a deep copy of the template so I think we are all ok with this. So I think this was a protection from before versioning and can just be removed. Needs careful checking and UAT.

raycarrick-ed:
Fixes #1772 
Fixes #2304 

Changes proposed in this PR:
- adds in logic for conditional questions


briri:
Hmm. A lot of these files seem related to bootstrap-select. I wonder if there's a cleaner way to add it to our yarn/webpacker config instead of manually maintaining a copy in app/stylesheets.

This one might be worth a walk-through on Friday since its pretty involved.

db/schema.rb
```diff
@@ -68,6 +80,27 @@
     t.integer  "phase_id"
   end
 
+  create_table "file_types", force: :cascade do |t|
```
briri: looks like a bunch of defunct tables have leaked back into the schema.rb file.

you may be able to just manually remove `file_types` `file_uploads` and `splash_logs`

raycarrick-ed: Removed those tables from schema.rb so it's now the same as under development branch but with the extra tables needed for conditionals and it looks like it still works fine.

app/helpers/conditions_helper.rb
```diff
@@ -93,77 +99,88 @@ def section_info(plan, section)
     section_hash
   end
 
-
-	# returns a collection of questions to remove (hide). Choose from these which are to be removed by a given condition
+  # TODO: the logic of this is WAY too complex
```
briri: thanks for adding this. I have added some `TODO` comments in other places as well. We should do a search and create issues for any of them at some point

app/models/question.rb
```diff
@@ -65,8 +65,7 @@ class Question < ActiveRecord::Base
 
   has_one :template, through: :section
 
-  has_many :conditions, through: :question_options
-
+  has_many :conditions, dependent: :destroy, inverse_of: :question
```
briri: great! thanks for including the auto destroy here

app/models/question.rb
```diff
@@ -227,23 +214,24 @@ def update_conditions(param_conditions)
   end
 
 
-  def saveCondition(value, option_id, remove_id = nil)
-    c = self.question_options.find(option_id).conditions.build
-    c.action_type = value['action_type']
-    c.remove_question_id = remove_id
+  def saveCondition(value)
+    c = self.conditions.build
```
briri: nice. I haven't seen `.build` on an association before. Will use it in the future

app/models/question.rb
```diff
@@ -227,23 +214,24 @@ def update_conditions(param_conditions)
...
+    c.action_type = value["action_type"]
     c.number = value['number']
-    c.webhook_data = get_webhook(value)
+    c.option_list = value["question_option"]
+    if value["action_type"] == "remove"
+      c.remove_data = value["remove_question_id"]
+    else
+      c.webhook_data = {
+        name: ActionController::Base.helpers.sanitize(value['webhook-name']),
```
briri: maybe better to just do the sanitization at the controller before passing the hash to this method since the helper is meant to be used in that context

app/models/section.rb
```diff
@@ -31,6 +31,8 @@ class Section < ActiveRecord::Base
   include ActsAsSortable
   include VersionableModel
 
+  # Sort order: Number ASC
+  default_scope { order(number: :asc) }
```
briri: 👍 I think this makes a lot of sense but we may want to highlight the change for people doing UAT in case the ordering of sections is off anywhere when customizing or using the drag-drop feature

app/models/template.rb
```diff
@@ -72,7 +72,7 @@ class Template < ActiveRecord::Base
 
   has_many :question_options, through: :questions
 
-  has_many :conditions, through: :question_options
+  has_many :conditions, through: :questions
```
briri: makes much more sense!

app/helpers/conditions_helper.rb
```diff
@@ -135,105 +148,33 @@ def section_title(section)
...
-  def conditions_ordered(conditions)
-    grouped_conditions = group_show_conditions(conditions)
-    grouped_conditions.each do |option, conditions|
-      conditions.sort_by{|condition| condition.action_type.to_s.length}
-    end
-    grouped_conditions
-  end
+  # used when displaying a question while editing the template
+  # converts condition into text
+  def condition_to_text(conditions)
```
briri: Another suggestion to make things even more readable:
This helper is doing a lot for different contexts. I would suggest moving the controller specific bits to controllers/concerns/ and these view specific bits to a presenter. I think helpers should only be used for generic view stuff like formatting dates, etc.

briri:
Approved but leaving unmarked so @xsrust and @martaribeiro can review. 

@raycarrick-ed one of the new tests for this condition logic is failing

raycarrick-ed:
Before I saw @briri comment suggesting rebase I had already merged with dev which I *think* should give a valid merge into dev. So I'm assuming a rebase is not needed now?? if that's wrong come back to me.

briri:
nah, the merge is fine @raycarrick-ed 
it looks like one of the new tests for the conditional logic is failing though

xsrust:
Addresses the routing half of #119

app/controllers/org_admin/questions_controller.rb
```diff
@@ -98,18 +98,51 @@ def create
...
+      
+      # get a map from question versionable id to old id
+      old_question_ids = {}
+      question.template.questions.each do |q|
+        old_question_ids[q.versionable_id] = q.id
+      end
+
       question = get_modifiable(question)
+
+      # params now out of sync (after versioning) with the question_options
```
briri: so are these steps not necessary if the template has not been versioned?
you can determine whether or not the template will be versioned by the `get_modifiable` call above by checking `question.template.generate_version?` which just does a check on its published state

briri: looks like you might need to rebase against dev as well

app/models/question.rb
```diff
@@ -201,18 +201,20 @@ def annotations_per_org(org_id)
...
+    res = true
     self.conditions.destroy_all
 
     if param_conditions.present?
       param_conditions.each do |_key, value|
-        saveCondition(value, old_to_new_opts, question_id_map)
+        res = saveCondition(value, flash, old_to_new_opts, question_id_map)
       end
     end
+    return res
```
briri, line 213: no need for the `return` here. Ruby methods always return the output of the last line.

app/models/question.rb
```diff
@@ -201,18 +201,20 @@ def annotations_per_org(org_id)
...
   # after versioning
-  def update_conditions(param_conditions, old_to_new_opts, question_id_map)
+  def update_conditions(param_conditions, flash, old_to_new_opts, question_id_map)
+    res = true
     self.conditions.destroy_all
 
     if param_conditions.present?
       param_conditions.each do |_key, value|
-        saveCondition(value, old_to_new_opts, question_id_map)
+        res = saveCondition(value, flash, old_to_new_opts, question_id_map)
```
briri, line 210: `res` will always be whatever the result of the last condition are, so if there are 3 and the 2nd one is false this will still return true.
Maybe add a `break if res == false`?

app/models/question.rb
```diff
@@ -236,6 +238,10 @@ def saveCondition(value, opt_map, question_id_map)
         c.remove_data = new_question_ids
       end
     else
+      unless value['webhook-email'] =~ URI::MailTo::EMAIL_REGEXP
+        flash[:alert] = "bad email address"
```
briri, line 242: it's unusual to set the contents of `flash` in a model.

app/views/org_admin/conditions/_webhook_form.html.erb
```diff
@@ -18,7 +18,7 @@
         <div class="md-form mb-5">
           <i class="fa fa-envelope prefix grey-text"></i> <!-- mdbootstrap should have treatment for bad input for this -->
           <%= label_tag name_start + "[webhook-email]", _('Email'), {'data-error': 'wrong', 'data-success': 'right'} %>
-          <%= text_field_tag name_start + "[webhook-email]", nil, placeholder: _('recipient_email@example.com'), class: 'form-control validate' %>
+          <%= email_field_tag name_start + "[webhook-email]", nil, placeholder: _('recipient_email@example.com'), class: 'form-control validate' %>
```
briri, line 21: in most browsers, using the `email_field_tag` like you have here should be enough. You can also consider wiring in some JS just in case. Maybe something like:
```javascript
import { isUndefined } from '../../../utils/isType';
import { isValidText, isValidEmail } from '../../../utils/isValidInputType';

$(() => {
  const form = $('#my-form-id');

  if (!isUndefined(form)) {
    form.on('submit', (e) => {
      form.find('input[type="email"]').each((i, el) => {
        const value = el.val();
        // First ignore if it was left blank <- remove if you require a value
        if (isValidText(value)) {
          if (!isValidEmail(value)) {
            e.preventDefault();
            return false;
          }
        }
      });
    });
  }
  return true;
});
```

raycarrick-ed:
I wasn't too happy about the flash in model thing either. After some experimenting, it turns out that changing the field to being an email field rather than a text field and then removing the "data-dismiss" from the save button give the correct behaviour and validates the email address. So that is changed here:

c064fea8af61ff26ff9371b6c9dc051a92c2984f

briri:
I'm glad the `email` input type was enough. We should consider going through and making sure we're using all of the HTML5 input types across the site (after Rails 5)

PR #2360, opened 2020-01-22
Part 3 - maDMP ROR/Fundref - ROR service

briri:
Part 3 - A base service and ROR service (inherits from the base service) that handles communications with the ROR API and converts the API results into a standard hash that only includes the things we care about.

**Note that this PR is against the 'ror-fundref' parallel branch and not 'development'**
- Added a `base_service.rb` file that generically handles communications with external APIs
- Added a `ror_service.rb` which communicates with the ROR API and preps results for use in DMPRoadmap
- Added a new initializer for the `ror_service.rb`
- Enabled web mock for Rspec tests and added a Webmocks support file
- Removed unnecessary complexity from the Rspec Capybara helper and instead pulled in the capybara-webmock gem
- Fixed some Rubocop issues with `Gemfile`, `spec/rails_helper.rb` and `spec/spec_helper.rb`
- Added tests for new services
- Fixed an issue with the `app/views/shared/export/_plan_coversheet.html.erb` that was causing test failures if the branding.yml was missing info

Note that `Rails.configuration.x.[topic].[attribute]` is the preferred method for app config in Rails 5+
https://guides.rubyonrails.org/v4.2/configuring.html#custom-configuration


app/services/external_apis/base_service.rb
```diff
@@ -0,0 +1,134 @@
+# frozen_string_literal: true
+
+module ExternalApis
+
+  class ExternalApiError < StandardError; end
+
+  class BaseService
```
xsrust: I really like this base-API-service approach.  Particularly with the maDMP need to consume other APIs for the various controlled vocabularies/identifiers.

Issue #2362, opened 2020-01-24
Default language on user-account-edit differs from application default


xsrust:
When a new user to the tool first updates their profile, the language box will default to the first  returned from the database.  

Our logic for this page's default language is:
`current_user.language.nil? ? Language.id_for(FastGettext.default_locale) : current_user.language.id`

The `id_for` function is defined as:
`Language.where(abbreviation: abbreviation).pluck(:id).first`

xsrust:
It looks like this is once-again due to the fast-get-text vs i18n discrepancy in naming schemes
the default_locale according to gettext is en_GB, whereas the abbreviation is en-GB.

Given we  have  a  `Language.default` column in the database, I'm going to swap to pulling this in, rather than attempting to use  the  language configured in Gettext.

magdalenadrafiova:
I am not sure whether this is the desired behaviour 
![image](https://user-images.githubusercontent.com/47178106/73654512-d23b8e80-4683-11ea-8a5f-ff33461901f8.png)

magdalenadrafiova:
sam to look at still

magdalenadrafiova:
works well - thank you - created new account and the language was this time set up as English 
![image](https://user-images.githubusercontent.com/47178106/74328367-a1dfa880-4d85-11ea-8489-05a85b719625.png)

sjDCC:
Good for me

PR #2363, opened 2020-01-24
made the user_edit_profile lang field default to DB default language

xsrust:
see #2362 for details


briri:
Travis shows a failure above but when you click through its all passing.  🤷‍♂ 
Github workflows are happy so going to merge it

xsrust:
Yeah, the template_versioning_spec was throwing something on the funder-template-update not showing to an org-admin users (and only on PG-tests)
But I kicked it and it worked after

PR #2379, opened 2020-02-04
Fixed an issue where the tinymce content css was not being loaded

xsrust:
Webpacker outputs the _tinymce_content.css file to the public directory with an asset fingerprint.

In order to get the correct asset path, I had to hook up the integration with `.js.erb` and webpacker.  
This was done by running `bundle exec rails webpacker:install:erb`

After renaming the tinymce.js file to a `.js.erb` I also needed to give the full extension in all files which import that JS file.

Finally, in order  to access helpers such as `asset_path` you need to call them on `ActionController::Base.helpers` as they are not part of the default scope

### DIFF
--- a/app/javascript/utils/tinymce.js.erb
+++ b/app/javascript/utils/tinymce.js.erb
@@ -11,6 +11,8 @@ import 'tinymce/plugins/paste';
 import 'tinymce/plugins/advlist';
 // Other dependencies
 import { isObject, isString } from './isType';
+// Pull in the rails helper functions
+<% helpers = ActionController::Base.helpers %>
 
 // // Configuration extracted from
 // // https://www.tinymce.com/docs/advanced/usage-with-module-loaders/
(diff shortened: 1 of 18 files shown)


briri:
think you may need to add the erb loader to the `spec/karma/unit.js` file.

xsrust:
https://jaketrent.com/post/import-erb-into-karma/
Followed this guide for the above commit 
(something's  messed up with running headless chrome witthin my docker-env, so I'm having to use github-actions as my test env)

briri:
@xsrust maybe try giving the new Capybara config in the ror-fundref branch a try. I had to swap in capybara-webmock gem and remove the chrome headless browser. (not sure if it will work without Webmock itself enabled but worth a try). It requires that you add `gem "capybara-webmock"` to the Gemfile

https://github.com/DMPRoadmap/roadmap/blob/ror-fundref/spec/support/capybara.rb

nicolasfranck:
In production mode rails still generates `/assets/blocks/_tinymce_content.css` as its path,
and therefore cannot find it. If you open a plan's edit form, you'll see a bunch of requests from
the iframe's of tinymce, and all of them fail.

Maybe use the same approach as for the skins, where you've them into `public/tinymce`?

briri:
I've discovered this as well in my instance of the repo. I have capistrano copy the css file over but it would be good to just add it to the public dir here

briri:
just submitted PR #2767 to add the tinymce.css to the public/tinymce dir

PR #2384, opened 2020-02-06
New JQuery autocomplete interface and OrgSelection service

briri:
Another big one for the ROR Fundref ticket
- Adds some flags to the `identifier_schemes` table to help contextualize them
- Added some OrgSelection services that provide a bridge between Controllers and searching for Orgs in the local DB and via the RORService
- Added a new Controller Concern to make interaction with the OrgSelection services easier
- Replaced the old JQuery Accessible Autocomplete widget and html with the new one
- Added some presenters to aid in rendering Identifiers and the Org Selection autocomplete box
- Added tests
- Added some new rake tasks to the upgrade Rake task file
- ran the Annotate gem (created some noise in this ticket with comments about f_keys)
- Fixed tests broken by these changes

Concepts:
- Not all of the JQuery autocomplete boxes make calls out to ROR (admin templates org changer and the ones on the create plan page do not)
- The JQuery autocomplete stores the selected Org as a hash value that contains its name and relevant identifiers (e.g. local DB id, ROR id, Fundref id, etc.)
- That hash gets passed to the controller which in turn sends it over to the OrgSelection services to convert the value into an Org model
- Some of the controllers (registration and super_admin/orgs) will create the Org if it is new
- The OrgSelection presenter works with the OrgSelection services as well to convert Org models over to the hash mentioned above


app/models/identifier_scheme.rb
```diff
@@ -38,9 +42,21 @@ class IdentifierScheme < ActiveRecord::Base
...
 
   scope :active, -> { where(active: true) }
-  scope :for_users, -> { active.where(name: %w[shibboleth orcid]) }
-  scope :for_orgs, -> { active.where(name: %w[shibboleth ror fundref]) }
-  scope :for_plans, -> { active.where(name: %w[doi]) }
-  scope :authenticatable, -> { active.where(name: %w[shibboleth orcid]) }
+  scope :for_users, -> { active.where(for_users: true) }
+  scope :for_orgs, -> { active.where(for_orgs: true) }
+  scope :for_plans, -> { active.where(for_plans: true) }
+  scope :authenticatable, -> { active.where(for_auth: true) }
```
xsrust: I forget if these field names were needed for the model polymorphism.
If not, we might want to condiser bitflags (if not mutually exclusive), or an enum (if they are)

db/migrate/20200130160919_contextualize_identifier_schemes.rb
```diff
@@ -0,0 +1,8 @@
+class ContextualizeIdentifierSchemes < ActiveRecord::Migration
+  def change
+    add_column :identifier_schemes, :for_auth, :boolean, default: false
+    add_column :identifier_schemes, :for_orgs, :boolean, default: false
+    add_column :identifier_schemes, :for_plans, :boolean, default: false
+    add_column :identifier_schemes, :for_users, :boolean, default: false
```
xsrust: I forget if these field names were needed for the model polymorphism.
If not, we might want to condiser bitflags (if not mutually exclusive), or an enum (if they are)

db/migrate/20200203190734_add_funder_and_org_to_plans.rb
```diff
@@ -0,0 +1,6 @@
+class AddFunderAndOrgToPlans < ActiveRecord::Migration
+  def change
+    add_reference :plans, :org, foreign_key: true
+    add_column :plans, :funder_id, :integer, index: true
```
xsrust: Does this point out at a  ROR funder id?
If so, could we get rid of the `funder_name` field on the plan and just use the `name` of the funder with id `funder_id`?

app/models/org.rb
```diff
@@ -64,6 +64,10 @@ class Org < ActiveRecord::Base
 
   has_many :guidance_groups, dependent: :destroy
 
+  has_many :plans
+
+  has_many :funded_plans, class_name: "Plan", foreign_key: "funder_id"
```
xsrust: This would be a nice thing to add to the statistics page

app/models/plan.rb
```diff
@@ -26,16 +26,23 @@
...
 #
 # Foreign Keys
 #
 #  fk_rails_...  (template_id => templates.id)
+#  fk_rails_...  (org_id => orgs.id)
 #
 
+# TODO: Drop the funder_name column once the funder_id has been back
+#       filled and we're removing the is_other org stuff
+
```
xsrust: Looks like great minds think alike.
Ignore my other comment about doing this

app/controllers/concerns/org_selectable.rb
```diff
@@ -0,0 +1,61 @@
...
+  extend ActiveSupport::Concern
+
+  # rubocop:disable Metrics/BlockLength
+  included do
+    private
+
+    # Converts the incoming params_into an Org by either locating it
+    # via its id, identifier and/or name, or initializing a new one
+    def org_from_params(params_in:)
+      params_in = params_in.with_indifferent_access
```
xsrust: I didn't realise this was a type of hash, I've been slowly driven mad by the string vs symbol key thing for literally years

db/migrate/20200203190734_add_funder_and_org_to_plans.rb
briri: yes. There is data in that funder_name field (at least in DMPTool) and I had meant to write a script to try and convert it over to an actual funder. I saw the funder_name cleanup as part of a second pass where we truly eliminate the org/user_identifiers tables and remove the is_other flag and default Org

db/migrate/20200130160919_contextualize_identifier_schemes.rb
briri: yes, a bit flag would work here

app/controllers/concerns/org_selectable.rb
briri: me too! I'm hoping its the default in Rails 5-6.

briri:
thanks for reviewing @xsrust. I'll change those flags over to a bitflag, address the failing tests and write a rake task to back fill the funder_names.

If you're comfortable with that, I will merge this into the jror-fundref branch and then submit the final PR to development for final review.

Issue #96, opened 2017-02-14
Tie into the Fundref list of funder organisations


briri:
We need to tie the Crossref's Fundref registry into DMP Roadmap to provide us with an authoritative list of funders. This will be a part of the plan creation process as well as the Org details page (to allow a funding organisation to select themselves from the list).

It would probably make sense to create a table/model structure similar to our 'identifier_schemes <--> user_identifiers <--> users' setup to manage this and any other systems we want to begin tying into in the future.

More info about [Fundref](https://www.crossref.org/services/funder-registry/).

briri:
We may prefer to use the new Research Organization Registry (ROR) once its online and ready for use

pherterich:
Closing this one now that #2389 is merged - a new one can be opened if any further work is needed.

PR #2389, opened 2020-02-10
maDMP ROR/Fundref integration

### LINKED ISSUE #2339: Build integration with ROR identifiers
As an organizational admin or super user creates a new organization in Roadmap this should ping the [ROR API](https://github.com/ror-community/ror-api) and record the associated identifier.

Ideally, we should also record the ROR for the Principle Investigator & Data Contact Person.

Outstanding question:
- For the DMPTool the organizational information is generated/recorded prior to authentication; for DMPOnline it is after. So, we'll need to determine corresponding workflows with this in mind.

briri:
Fixes #2339 .

This PR removes the old jquery-accessible-autocomplete field we had been using with a new JQuery Autocomplete widget that allows you to supply it with a pre-loaded list of Orgs or a route to a controller that it will use to load search results dynamically. The new field allows the user to select an item from the list or type in their own organization name.

Places where the new Org Selector is used:
- Create an Account - searches both the local DB and ROR
- Finish creating an account from an invitation - searches both the local DB and ROR
- Edit Profile - searches both the local DB and ROR
- Create Plan - both Research Org and Funder boxes use preloaded list of Orgs from DB
- Edit Project/Plan Details - Funder now searches both the local DB and ROR but limits to Funders only
- Super Admin Org Details page - searches ROR only to update the ROR and Fundref ids
- Super Admin Templates page - Org changer uses a repopulated list of local DB orgs

The system will now auto-create new Orgs if we do not have a matching record in the `orgs` table for the Org the user selects/types on the: create account, accept invitation and edit profile pages.

DB changes proposed in this PR:
- Added `orgs.managed` flag
- Added `for_auth`, `for_orgs`, `for_plans`, and `for_users` flags to the `identifier_schemes` table
- Added a `org_id` and `funder_id` to the `plans` table
- Created an `identifiers` polymorphic table to replace the `org_identifiers` and `user_identifiers` tables

Code changes proposed win this PR:
- Updated models to handle new fields
- Added a `save_identifiers!` method to the Org model
- Added scopes to the IdentifierScheme model
- Added the new Identifier model and updated Org, Plan and User to reference the new table instead of their old identifiers tables
- Added a new Controller Concern that handles the translation of the Org Selector's hash to an Org  model.
- Replaced `funder_name` on `planes` with the new association to `funder_id`
- Added new services to Handle Org Selection searches and converting Hash <-> Org model
- Added a new ROR service which handles communication with ROR
- Updated the JS, views and controllers to work with the new JQuery Autocomplete Org selector
- Added presenters to facilitate the display of the new Org Selector and Identifiers
- Added tests and updated ones that were broken by the changes

**Note:** You should run `rake upgrade:v2_2_0` which will perform the following tasks:
- Add the ROR and FundRef IdentifierSchemes
- Update the Shibboleth IdentifierScheme's description to: 'Institutional Sign In (Shibboleth' (how it will appear to Org/Super Admins on the Org Details page IF `shibboleth_use_filtered_discovery_service` is enabled in the `application.rb`)
- Updates the various new IdentifierScheme flags
- Makes new `identifiers` records from the ones in `org_identifiers` and `user_identifiers`
- Makes all existing orgs `managed` by default
- Converts all existing `other_organisation` entries into Orgs
- Retrieves the ROR/Fundref ids for all existing Orgs - the results are written out to `tmp/ror_fundref_ids.csv`. **Note:** You should review this output and manually correct any incorrect matches in the UI (using the ROR lookup at the bottom of the Org Details page)


xsrust:
Fixes: #96

briri:
rebased with development

app/controllers/super_admin/org_swaps_controller.rb
```diff
@@ -2,20 +2,21 @@
...
-    end
+
+    # See if the user selected a new Org via the Org Lookup and
+    # convert it into an Org
+    lookup = org_from_params(params_in: org_swap_params)
+
     # rubocop:disable Metrics/LineLength
-    if @org.present?
-      current_user.org = @org
+    if lookup.present? && !lookup.new_record?
```
xsrust, line 18: Nice validation so we don't end-up creating new orgs by switching

app/models/guidance.rb
```diff
@@ -81,7 +81,7 @@ class Guidance < ActiveRecord::Base
 
   # Returns whether or not a given user can view a given guidance
   # we define guidances viewable to a user by those owned by a guidance group:
-  #   owned by the managing curation center
+  #   owned by the default orgs
```
xsrust, line 84: Thanks, this should make things a bit easier for people who pick up the codebase but aren't a `curation center`

app/models/identifier_scheme.rb
```diff
@@ -34,4 +37,26 @@ class IdentifierScheme < ActiveRecord::Base
...
+  # ===========================
+  # = Instance Methods =
+  # ===========================
+
+  # The name is used by the OrgSelection Services as a Has key. For example:
+  #    { "ror": "12345" }
+  # so we cannot allow spaces or non alpha characters!
+  def name=(value)
+    super(value&.downcase&.gsub(/[^a-z]/, ""))
+  end
```
xsrust, line 60: This is a neat little pattern for overriding the value.  It seems like it will be easier to test than callbacks as well.

Would it be worthwhile adding a validation for this above on the name's value?

app/models/org.rb
```diff
@@ -63,6 +64,10 @@ class Org < ActiveRecord::Base
 
   has_many :guidance_groups, dependent: :destroy
 
+  has_many :plans
+
+  has_many :funded_plans, class_name: "Plan", foreign_key: "funder_id"
```
xsrust, line 69: This is a neat association

app/models/org.rb
```diff
@@ -170,6 +198,10 @@ def self.human_attribute_name(attr, options = {})
     HUMANIZED_ATTRIBUTES[attr.to_sym] || super
   end
 
+  def links
+    super() || { "org": [] }
+  end
```
xsrust, line 203: Is this to set the default value for forms in-case the value is nil?

app/models/user.rb
```diff
@@ -241,9 +242,14 @@ def can_super_admin?
   #
   # Returns Boolean
   def can_org_admin?
-    self.can_grant_permissions? || self.can_modify_guidance? ||
-      self.can_modify_templates? || self.can_modify_org_details? ||
-      self.can_review_plans?
+    return true if can_super_admin?
+
+    # Automatically false if the user has no Org or the Org is not managed
+    return false unless org.present? && org.managed?
```
xsrust, line 248: We validate the presence of the user's org, so not sure if the `org.present?` is needed

app/models/user_identifier.rb
```diff
@@ -11,6 +11,7 @@
 #
 # Indexes
 #
+#  fk_rails_fe95df7db0                (identifier_scheme_id)
```
xsrust, line 14: Is the UserIdentifier model sticking around for a release to migrate the objects over to the new Identifier model?

db/migrate/20200203190734_add_funder_and_org_to_plans.rb
```diff
@@ -0,0 +1,6 @@
+class AddFunderAndOrgToPlans < ActiveRecord::Migration
+  def change
+    add_reference :plans, :org, foreign_key: true
+    add_column :plans, :funder_id, :integer, index: true
```
xsrust, line 4: Is it worthwhile enforcing this as a foreign key to the orgs table, or is this sufficiently captured with our validation on the model?

app/models/answer.rb
```diff
@@ -9,12 +9,16 @@
 #  text         :text
 #  created_at   :datetime
 #  updated_at   :datetime
+#  label_id     :string
```
xsrust, line 12: Not sure where this is coming from.
I didn't see the change in the schema or the migrations added.

lib/tasks/upgrade.rake
```diff
@@ -696,6 +708,160 @@ namespace :upgrade do
...
+    Org.all.update_all(managed: true)
+  end
+
+  desc "Attempts to migrate other_organisation entries to Orgs"
+  task migrate_other_organisation_to_org: :environment do
+    names = User.all.where.not(other_organisation: nil)
+                .pluck(:other_organisation).uniq
+    names.each do |name|
+      # Skip any blank names
+      next unless name.gsub(/\s/, "").blank?
```
xsrust, line 795: Should this be either `if name.gsub(/\s/, "").blank?` or `unless name.gsub(/\s/, "").present?`

briri: yep. will patch in the `identifier-refactor` PR

app/models/identifier_scheme.rb
briri: Hmm. This would happen before any `validate` or `valid?` calls so I think it catches any issues. There is no UI for this either (maybe there should be) so a validation would potentially start causing errors for existing records in some installations. I will add another step to the upgrade rake tasks to convert them to lower case.

app/models/org.rb
briri: yes. I encountered some errors with newly created records. Since everyone's tables likely have records with a nil there I thought it safer to do on the getter instead of the setter. We could update the setter though and add an upgrade rake task

app/models/user.rb
briri: This occurred for me with an invite record. I don't recall the exact context in which this method was called.

app/models/user_identifier.rb
briri: yep. need it for the migration of data unfortunately

db/migrate/20200203190734_add_funder_and_org_to_plans.rb
briri: it seems to be working the same at least on the Rails side. Not sure if an fkey versus a regular index are different on the DB side.

app/models/answer.rb
briri: hmm. me either. I only have 6 records in my DB with a value for it (all pre-2014). My data model must be out of sync for that table. Will remove it in the `identifier-refactor` branch

briri:
- rebased off of the latest dev branch
- removed the `label_id` from the annotate comments on the answer model and factory
- Fixed the upgrade rake task so that it checks 'if the name is blank' instead of 'unless the name is blank'

briri:
rebased to fix conflict with db/schema.rb

PR #2391, opened 2020-02-11
Bug/new account language

xsrust:
was previously passing the language object, not the id to the helper-function, causing the incorrect behavior


xsrust:
addresses #2362

Issue #2083, opened 2019-04-12
The Usage page "No of plans during last year" and "No. plans by template" total for some months differ slightly.


johnpinto1:
**Expected behaviour:**
The **No of plans during last year** for a given month should be equal to *No. plans by template* total for the same month period.

**Actual behaviour:**
In some cases the figures differ. See the **Feb-19** in images below. The figures are **5** and a total of **8** respectively.
![NoofPlansLastYear](https://user-images.githubusercontent.com/8876215/56036887-e9d90d80-5d25-11e9-80b1-86f94297c44b.jpeg)
![noOfTemplates](https://user-images.githubusercontent.com/8876215/56037225-aa5ef100-5d26-11e9-89ac-ba1adfa4ead8.jpeg)

The calls 
http://localhost:3000/api/v0/statistics/created_plans?range_dates... returns
{"Apr-18":47,"May-18":98,"Jun-18":116,"Jul-18":0,"Aug-18":0,"Sep-18":0,"Oct-18":3,"Nov-18":0,"Dec-18":0,"Jan-19":1,**"Feb-19":5**,"Mar-19":2}
http://localhost:3000/stat_created_plans_by_template?start_date=2018-04-30 (
[{"count":2,"date":"2019-03-31","by_template":[ ... **{"count":8,"date":"2019-02-28","by_template":[{"name":"AHRC Data Management Plan","count":6},{"name":"Standard CRUK Template","count":1},{"name":"Hartstichting","count":1}]}**, ...

**Steps to reproduce:**
Choose the **Time period** _"Last 12 months"_ for **No. of Plans by template**. Now total the number of plans for each month by moving the mouse over the stacked bar. Compare these figures with the the corresponding months on the **No. plans during last year**.
Or you can compare the Ajax requests

raycarrick-ed:
I think the reworking moving the logic from JS to ruby looks good and could serve as a model for AJAX calls going forward.
@briri 

Two very small points:

in _filter.html.erb the "Go" button can overlap with the organisation menu when that menu is quite wide (i.e there are long org names) would it be better perhaps on the next line?

The "Download" buttons all have "open in a new window" attached to them when they in fact open a popup dialog. Is this worthwhile or better just left off as it's not really what it's doing?

that's it. The rest all looks fine.

briri:
thanks @raycarrick-ed. I had done most of my testing as an org admin so didn't notice the issue with the Org dropdown. 

[...]

PR #2392, opened 2020-02-11
fixed an inconsistency in  the template counts for the usage dashboard

xsrust:
After some digging, it turned out the inconsistency was twofold.  

First, when generating the  statistics objects, we organise the templates by their `family_id` but we save the stats objects by `template.title`.  This means when both customized and non-customized versions of a template  are used within an organisation, we save  two separate counts, with the same name, ex:
```
{"count"=>8,
  "date"=>"2019-05-31",
  "by_template"=>
   [{"name"=>"BBSRC Template", "count"=>1},
    {"name"=>"EPSRC Data Management Plan", "count"=>1},
    {"name"=>"University of Manchester Generic Template", "count"=>3},
    {"name"=>"Wellcome Trust Template", "count"=>1},
    {"name"=>"Wellcome Trust Template", "count"=>1},
    {"name"=>"TEST New funder template", "count"=>1}]},
```

Then, when we're re-martialing the data for  the plans_by_template chart, we iterate, identifying the different templates by name, and thus over-write the data with the last count.  

It might be worthwile to adjust the way we save the statistics to differentiate between raw and customized funder templates, but for the  meantime, I adjusted  the template-chart helper to just  add the counts when it encounters duplicate entries.

### DIFF
--- a/app/helpers/usage_helper.rb
+++ b/app/helpers/usage_helper.rb
@@ -43,7 +43,7 @@ module UsageHelper
         # Replace any of the month/year plan counts for this template IF it has
         # any plans defined
         template_hash[:data] = template_hash[:data].map do |data|
-          data[:y] == date ? { x: template["count"], y: data[:y] } : data
+          data[:y] == date ? { x: template["count"] + data[:x], y: data[:y] } : data
         end
         datasets[template["name"]] = template_hash
       end


xsrust:
Addresses #2083

PR #2401, opened 2020-02-18
added wkhtmltopdf package and updated postgres image to latest

briri:
Trying to fix a couple of new Github Workflow issues that popped up


briri:
this addresses an issue that started coming up with the Postgres image we use in Github Workflows. Apparently Postgres added a change to their default security that prevents the postgres from running with the default username/password. We are ok with it using those for testing, so followed instructions in the following Github issue: https://github.com/docker-library/postgres/issues/681

briri:
looks like we have a new issue with wkhtmltopdf though :/

Issue #2349, opened 2020-01-13
Support for multiple contributors


mariapraetzellis:
Currently the application allows for 1 PI and 1 "Data Contact Person." In order to support machine actionability of DMPs, we need to expand the number of people that can be linked to a DMP. This could be a repeatable field so that a user could add as many PIs and Data Curators as applicable. This may need a specific own tab in the application. We'd like to make the PI & data contact mandatory.

mariapraetzellis:
@briri  to look at the Credit taxonomy- was a recommendation from a user

briri:
UI work will be done as part of #2408

sjDCC:
I'm happy with the separate tab and think the functionality to add contributors works well but have a couple of suggestions.
![Capture](https://user-images.githubusercontent.com/6169940/80204446-c925e200-8620-11ea-9af1-1b857ad0118e.PNG)
![new-contrib](https://user-images.githubusercontent.com/6169940/80204777-62ed8f00-8621-11ea-8870-b56b00cf84c1.PNG)

1. We should rename the tab "Collaborators" not "Contributors" as that's more even. Maria, Magdalena and I discussed this on a Product Manager call

2. We should pre-enter the plan owner as PI like we did previously on project details and allow them to amend entry. Pre-fill as much as possible from edit profile e.g. ORCID etc

3. Can we do a look-up integration with the ORCID field instead of this being a text entry?

4. Are the terms "Data Manager" "Principal Investigator" and "Project Administrator" from CRedIT Taxonomy? May be good to link out to that here or show them as indexed, controlled terms somehow? On the overview table like for RoR entry?

5. We discussed merging this with the table under "Share" and having the Collaborators tab be where users can list team members and assign plan editing roles. This may be a secondary piece of work to avoid delays now. It could fit under actions feasibly.

![credit-link](https://user-images.githubusercontent.com/6169940/80204889-98927800-8621-11ea-96bf-27e91134eb38.PNG)
![collab](https://user-images.githubusercontent.com/6169940/80204957-b8c23700-8621-11ea-97fc-8a938e24cf7f.PNG)

briri:
Added a new ticket #2481 for these adjustments. 

See #2350 for the ORCID question. ORCID does not provide a lookup API. We can pre-populate the value for the user who is logged in. ORCID's workflow though requires that the individual in question asserts that they are associated with the plan. Dryad has solved this by sending out an email to any contributors with a link that lets them assert this information. 

[...]

PR #2402, opened 2020-02-20
Adds contributors for plans

briri:
Addresses #2349.

These changes are dependent on the ROR/Fundref PR's new `identifiers` table

Changes proposed in this PR:
- Adds a `contributors` table (with f_key to `orgs`)
- Adds a `plans_contributors` join table that sets a bitflag `roles` field based on the CRediT Taxonomy (thanks for alerting us to that @xsrust)
- Adds the 3 new `ethical_issues` fields to the `plans` table
- Added factories/tests for the new models
- Rake upgrade task to convert all `data_contact` ('Data curation' role) and `principal_investigator` ('Investigation' role) information on a plan over to `contributors` and adds a 'Writing - original draft' role to all of the plan's authors.

Changes to model/UI forthcoming. Need to update the code that creates a plan or shares a plan to add these contributors.

### DIFF
--- a/app/models/plan.rb
+++ b/app/models/plan.rb
@@ -26,6 +26,9 @@
 #  created_at                        :datetime
 #  updated_at                        :datetime
 #  template_id                       :integer
+#  ethical_issues                    :integer
+#  ethical_issues_description        :text
+#  ethical_issues_report             :string
 #  org_id                            :integer
 #  funder_id                         :integer
 #
@@ -116,6 +119,10 @@ class Plan < ActiveRecord::Base
   has_many :roles
 
   has_many :identifiers, as: :identifiable
+
+  has_many :plans_contributors, dependent: :destroy
+
+  has_many :contributors, through: :plans_contributors
 
   # =====================
   # = Nested Attributes =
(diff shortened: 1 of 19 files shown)


app/models/plan.rb
```diff
@@ -420,7 +427,7 @@ def owner
                   .administrator
                   .order(:created_at)
                   .pluck(:user_id).first
-    User.find(usr_id)
+    usr_id.present? ? User.find(usr_id) : nil
```
briri: this was an edge case error that came up with some test records (a plan with no owner)

raycarrick-ed:
Looks good to me as far as I can tell.
My one (very minor) comment is that we use "firstname" in users but "given_name" here.
Not a big deal but a mild incnsistency.
Not merging so Sam and Marta can take a look.

briri:
yes, good point @raycarrick-ed I'll change it to firstname

briri:
rebased with workflow fixes

xsrust:
Looks like everything is now passing the tests.

I have a question about the implimentation, It makes sense to me that a Plan has_many contributors, but it's unclear to me why a  contributor would `have_many` plans.  It seems like this would allow anybody who's editing a contributor to update all references to that contributor, which I believe would be unintended.

briri:
Hmm. The original intention with using the plans_contributors join table was to keep the size of the contributors table down. I see your point though on a belongs_to relation being a better fit. I have switched it over in the api-v1 branch so closing this one out

PR #2403, opened 2020-02-21
issue#2375 allow change of delimiter for csv download

### LINKED ISSUE #2375: Delimiter to pipe in usage statititics instead of comma
is it possible to change the delimiter in the export of the usage statistics to,for example a ‘pipe’ | instead of comma

raycarrick-ed:
Fixes #2375  .

Changes proposed in this PR:
- add a select menu to the usage page which gives a choice of three separators
- this could be extended to allow for them to be specified in branding.yml but it'll do for starters
- each download then also has a "sep" field which gets changed by a JS event listener attached to the select menu
- the various CSV handling functions then get an extra sep parameter whch gets passed in from the download url

### DIFF
--- a/app/controllers/usage_controller.rb
+++ b/app/controllers/usage_controller.rb
@@ -37,7 +37,7 @@ class UsageController < ApplicationController
     authorize :usage
 
     data = Org::TotalCountStatService.call
-    sep = params["sep"]
+    sep = sep_param
     data_csvified = Csvable.from_array_of_hashes(data, true, sep)
 
     send_data(data_csvified, filename: "totals.csv")
@@ -74,7 +74,7 @@ class UsageController < ApplicationController
     authorize :usage
 
     user_data(args: default_query_args)
-    sep = params["sep"]
+    sep = sep_param
     send_data(CSV.generate({:col_sep => sep}) do |csv|
       csv << [_("Month"), _("No. Users joined")]
       total = 0
@@ -93,7 +93,7 @@ class UsageController < ApplicationController
     authorize :usage
 
     plan_data(args: default_query_args)
-    sep = params["sep"]
+    sep = sep_param
     send_data(CSV.generate({:col_sep => sep}) do |csv|
       csv << [_("Month"), _("No. Completed Plans")]
       total = 0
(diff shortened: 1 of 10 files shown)


app/javascript/views/usage/index.js
```diff
@@ -2,6 +2,24 @@ import { isObject } from '../../utils/isType';
...
+  const changeStatFnGen = (str) => {
+    const fn = (item, index) => {
+      item.href = item.href.replace(/sep=.*/, str);
+    };
+    return fn;
+  };
+
+  // attach listener to separator select menu
+  // on change look for "stat" elements and chnage their query param
+  document.getElementById('csv-field-sep').addEventListener('click', (e) => {
```
briri: we should look into converting this over to a `<form>` submission instead of a `<a>` link, that way we wouldn't need to do any manual JS it would just come through in the submission params.

Something to do after the Rails 5 upgrade

lib/csvable.rb
```diff
@@ -16,7 +16,7 @@ def from_array_of_hashes(data = [], humanize = true)
           .map(&:to_s)
       end
 
-      CSV.generate do |csv|
+      CSV.generate({:col_sep => sep}) do |csv|
```
briri: preferred style is `{ col_sep: sep }` instead of `{ :col_sep => sep }`. 
No need to worry about it now though, when we run Rubocop on the entire codebase it will auto-change these for us.

briri:
looks like there are some JS linter issues in the checks. feel free to merge yourself once those are cleared up. You can run `yarn run eslint --ext .js app/javascript/**/*.js` to run the linter locally

xsrust:
I fixed up the eslint issues, and also tweaked the styling to make the new drop-down display inline with the button text.  
Also noticed (because I'm working on a fresh roadmap DB from seeds) that the usage-dashboard breaks when there are no plans, so I added a default-case to the `start_plan_date` util in the usage_controller

app/controllers/usage_controller.rb
```diff
@@ -188,7 +188,8 @@ def total_users(args:)
   end
 
   def first_plan_date
-    StatCreatedPlan.all.order(:date).limit(1).pluck(:date).first
+    StatCreatedPlan.all.order(:date).limit(1).pluck(:date).first \
+    || Date.today.last_month.end_of_month
```
xsrust: fixes a problem when there are no stats in the db (for example after a fresh seed)

PR #2412, opened 2020-02-27
removed old unused Feedjira code and dependency

### LINKED ISSUE #2411: RSS news feed is broken
Hello everyone, 

DMP OPIDoR have been using the RSS feed for our own news. It seems to be broken since Feedjira 3.0.
The `Feed.fetch_and_parse` method has been removed in this version. 

https://github.com/DMPRoadmap/roadmap/blob/a54f85a85ddd50d5d0e0a0814c120d6d8dc6d1f8/app/controllers/static_pages_controller.rb#L5-L12

There is two ways of fixing this : 
1. Fixing the gem version to V2.2.0
2. Include the HTTParty gem and change the code as shown in the README : https://github.com/feedjira/feedjira/blob/master/README.md 

I tested the second solution, it seems to work without too much changes. 
Should I send you a PR ?

briri:
Fixes #2411 .

The old Feedjira code on the `static_pages_controller.rb`is no longer used. Removing it from the codebase.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -156,9 +156,6 @@ gem 'wicked_pdf', '~> 1.1.0'
 # This simple gem allows you to create MS Word docx documents from simple html documents. This makes it easy to create dynamic reports and forms that can be downloaded by your users as simple MS Word docx files. (http://github.com/karnov/htmltoword)
 gem 'htmltoword', '1.1.0'
 
-# A feed fetching and parsing library (http://feedjira.com)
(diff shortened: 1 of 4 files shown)


briri:
looks like we were using the Feedjira `String.sanitize` method in the footer. Switched it to use Rails' `sanitize` method.

magdalenadrafiova:
Ask benjamin to test @benjaminfaure

PR #2416, opened 2020-03-04
Ensure explicit plan sharing always works

### LINKED ISSUE #2415: Org admins cannot view plans explicitly shared by a user of another org
**Expected behaviour:**
When a user of any organisation shares a plan with any other user, they should be able to view the plan.

**Actual behaviour:**
When an org-admin has a plan shared by a user of another organisation, they do not have permissions to view the plan.

**Happening Because:**
Plan.readable_by(u_id) is defined below
```
def readable_by?(user_id)
    current_user = User.find(user_id)
    if current_user.present?
      # If the user is a super admin and the config allows for supers to view plans
      if current_user.can_super_admin? &&
          Branding.fetch(:service_configuration, :plans, :super_admins_read_all)
        true
      # If the user is an org admin and the config allows for org admins to view plans
      elsif current_user.can_org_admin? &&
          Branding.fetch(:service_configuration, :plans, :org_admins_read_all)
        owner_and_coowners.map(&:org_id).include?(current_user.org_id)
      else
        commentable_by?(user_id)
      end
    else
      false
    end
  end
```
As you can see, if the user is an org-admin, we don't check if the plan is shared with them to determine access.  I'll move that up to a guard-statement, and see if I can simplify the logic.

This is happening on live-services, so it should probably be submitted as a patch to Master

xsrust:
Fixes #2415   and adds some wider test coverage.

See ticket for details of what was going wrong before

### DIFF
--- a/app/models/plan.rb
+++ b/app/models/plan.rb
@@ -344,19 +344,17 @@ class Plan < ActiveRecord::Base
   #
   # Returns Boolean
   def readable_by?(user_id)
+    return true if commentable_by?(user_id)
     current_user = User.find(user_id)
-    if current_user.present?
-      # If the user is a super admin and the config allows for supers to view plans
-      if current_user.can_super_admin? &&
-          Branding.fetch(:service_configuration, :plans, :super_admins_read_all)
-        true
-      # If the user is an org admin and the config allows for org admins to view plans
-      elsif current_user.can_org_admin? &&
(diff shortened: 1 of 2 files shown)


xsrust:
**NOTE:**  PR against master branch

xsrust:
fixes to test-running from development for PG don't appear to be merged into Master yet, so that Action is expected to fail

xsrust:
merged in the github-workflow fixes

PR #2423, opened 2020-03-10
Changes were made to the HTML h tags - Issue 1451

martaribeiro:
Fixes # 1451.

Changes proposed in this PR:
- Changes were made to the HTML h tags in order to make it more accessibility friendly and compliant


app/views/answers/_locking.html.erb
```diff
@@ -1,6 +1,9 @@
+<!--
+  This partial displays when a question was edit by two or more users. The local variables are: answer, question, readonly, locking
```
briri: thanks for adding these comments!

app/views/devise/registrations/new.html.erb
```diff
@@ -1,19 +1,25 @@
...
               <%= _("Do you have a %{application_name} account?") % {
                 application_name: Rails.configuration.branding[:application][:name]} %>
-            </h3>
-            <p class="text-center">
+            </p>
+            <p class="text-center fontsize-h3 color-heading-text">
               <i class="fa fa-arrow-circle-down" aria-hidden="true"></i>
             </p>
             <h2 class="text-center">
-                <%= _("Sign in") %></i>
```
briri: good catch

app/views/phases/_overview.html.erb
```diff
@@ -1,7 +1,7 @@
 <%# locals: { phase } %>
 <div class="row" id="<%= dom_id(phase) %>">
   <div class="col-xs-12">
-    <h4>
+    <h4 class=''>
```
briri: don't need the class here if its empty

martaribeiro:
I have fixe the empty class. Which should now be a <p> tag with a class and not an <h> tag

Issue #2345, opened 2020-01-09
Changes to downloading usage reports 


mariapraetzellis:
We would like organizational admins to have the ability to download **all usage reports** in a single location and/or CSV file. 

A prior ticket (https://github.com/DMPRoadmap/roadmap/issues/1690) discussed including the following stats:

- Org users per month
- Org plans per month
- Org plans by template #1679
- Org plans downloads and shares #1745
- Org plans for which feedback was requested #1759

Should cover all historical data for the org, numbers/month

Please Investigate gems to see if we could have one downloadable CSV with tabs for each report.

Alternatively, we could have a dropdown list of reports within the application. 

Additionally, **please remove the ability to "Download global usage" CSV from the web application for organizational admins**. **This feature should only be present for Super Admins**. We can replace the "Download global usage" button with a new "Download usage" button that only contains the statics above related to the specific organization. 

![screenshot-docs google com-2020 01 09-14_07_43](https://user-images.githubusercontent.com/27511575/72108897-73cffa00-32e9-11ea-8489-471a44ae6838.png)

briri:
We will need to verify that the `exported_plans` table is collecting all plan downloads. My production table just has 'pdf' entries (it could be that our users aren't downloading in other formats). We do not collect historical statistics on # of shares, so would need to add that.

xsrust:
Re: `explorted_plans` only having PDF entries
This is the same on our services as well from early 2018.  I had a look to see where we create the entries, and the helper-function looks ok:
https://github.com/DMPRoadmap/roadmap/blob/84eb63fad24d6d4d67aee388e58b429949c3beaf/app/models/concerns/exportable_plan.rb#L177-L189
So I dug a bit deeper, and it's only being called in the above `prepare(coversheet: false)` method
Which is called from both the `as_pdf` and `as_csv` methods. 
These get called from the `plan_exports_controller` 
https://github.com/DMPRoadmap/roadmap/blob/84eb63fad24d6d4d67aee388e58b429949c3beaf/app/controllers/plan_exports_controller.rb#L30
with a second call for CSV exports in:
https://github.com/DMPRoadmap/roadmap/blob/84eb63fad24d6d4d67aee388e58b429949c3beaf/app/controllers/plan_exports_controller.rb#L55
But they all get a format of PDF from the final call in the concern.

We should do a quick refactor to ensure we aren't double-generating, and that we pass through the format for recording properly.

xsrust:
[...]

Issue #1682, opened 2018-07-06
Usage stats for funder view


sjDCC:
The current usage page provides number of users and number of plans for the organisation. This makes sense for universities, but funders would want to see all DMPs using their template, irrespective of which organisations the creating users belong to.

Can the view be adjusted by org type so that funders see the following:

1. Usage: total DMP figures by template for the template(s) they own (see related ticket #1679). This may need to be a separate graph specific to funders to differentiate it from the proposed plan/template graph for org-affiliated users

2. Plans: a list (and download) of all plans created using their template(s)

sjDCC:
Consider permissions for funder to download plan
1. User has a setting to grant permission (default on)
2. Only allow it to be shared after 50% complete

mariapraetzellis:
We will consult with Dutch and Irish funders to flesh this out. Have meetings scheduled in Jan & Feb

Agreed actions:
Consult with existing funders
Simple usage dashboard, similar to organisation one focusing on plans and template
Access to actual plan content can only come when the DMP has been marked as complete or accessible to funder, not in draft

xsrust:
I've done a first-pass on this by re-using the plans_by_template chart (bottom of current usage-dashboard page).  I've also generated the information for all organisations, as we may  determine down the line that we want to expose tool-wide usage of every org's templates.  

UI-wise, the view for a funder has become quite busy:
![usage_statistics_1](https://user-images.githubusercontent.com/7352616/76213897-8e4c2400-6203-11ea-9465-dc489249808c.png)

We could use one of our existing motifs to clean this up, either accordions or tabs.  I lean towards a view with accordions for each section.  I quite like the idea of putting the `Statistics on your Templates` accordion both first and open for Funders, and putting the `Statistics on your Users` first and open for universities.  If we decide not to show the `statistics on your templates` section to normal orgs, we could just omit the accordion markup entirely.

@sjDCC @magdalenadrafiova @mariapraetzellis   Let me know your thoughts on UI, as well as if this information would be interesting to all organisations, rather than just funders.

magdalenadrafiova:
@xsrust 

Is it possible to see the plan by the individual user - can they go into the plan and read the plan? Would that be a separate functionality for which we should raise a ticket?  

We do not want the funder to see an empty template. Is this selecting only plans that were completed by at least 50%? Or is this taking all the plans, even if they are empty.

xsrust:
@magdalenadrafiova 
[...]

PR #2427, opened 2020-03-12
Issue/2345

xsrust:
Addresses #2345

Overview of changes:
- Make `Download Global Usage` super-admin only
- Create `Download Monthly Usage` button to get a CSV of:
   - `Month, new plans, new users, downloads, plans shared`  
   - For each month since creation
- New models/services for statistics on exported and shared plans
  - Added these to the stats rake tasks
- Add parallel gem
  - Add parallelized tasks for `rake stat:build` and `rake stat:build_last_month`
  - Added threading to the other `create_` services
  - Change DB default pooling to 16

Note:
UI with the multiple buttons looks funny at the moment, but I will refactor the layout in another feature's PR as I'd started this work there.


config/database.yml.sample
```diff
@@ -3,7 +3,7 @@ defaults: &defaults
   encoding: <%= ENV['DB_ADAPTER'] == "mysql2" ? "utf8mb4" : "" %>
   username: <%= ENV["DB_ADAPTER"] == "postgresql" ? 'postgres' : '' %>
   database: roadmap_<%= ENV['RAILS_ENV'] %>
-  pool: 5
+  pool: 16
```
briri: Is this for the parallel gem?

lib/tasks/stat.rake
```diff
@@ -4,8 +4,51 @@ namespace :stat do
...
+    tasks = ["stat:create:created_plan",
+             "stat:create:joined_user",
+             "stat:create:shared_plan",
+             "stat:create:exported_plan",
+             "stat:create_last_month:created_plan",
+             "stat:create_last_month:joined_user",
+             "stat:create_last_month:shared_plan",
+             "stat:create_last_month:exported_plan"]
+
+      Parallel.each(tasks, progress: "Building Stats", in_processes: 4) do |task|
```
briri: 👍 this is great

briri:
looks like you have some failing stats tests too that need to be addressed

config/database.yml.sample
xsrust: yeah, essentially each thread/process takes an additional db connection, so it was flooding the default 5.  16 is a bit overkill for development, but it means we don't end up with connection-timeout errors

xsrust:
Looks like the github-workflows are getting confused when checking things out from the DCC fork.
I've updated the `mySQL` side of our workflows to the `checkout@v2` to patch the problem.  
If that works, I'll apply the same fix to the remaining workflow files.

https://github.com/actions/checkout/issues/23

xsrust:
Going to merge this so I can rebase the work on https://github.com/DMPRoadmap/roadmap/issues/1682 
They both affect the services and usage-dashboard heavily, so there's likely to be conflicts to fix on that branch.

PR #2430, opened 2020-03-13
added api_clients table and JWT

briri:
This 1st phase is limited in scope see below for details. Phase 2 will pull in all of the existing functionality from v0.

- Added new `api_clients` table that will be used to store credentials and contact info for external systems
- Added a `last_api_access` field to the `users` table
- Added a `contributors` table to store plan contributor information
- Added new Super Admin area to manage the `api_clients`
  - Pages allow the admin to 'refresh' the client's credentials and has a button to email the designated contact the credentials and information about API use.
- Updated the User profile so that a user can refresh their API token.
- Added 1st phase of the new API v1 that follows the RDA Common Metadata Standard. This 1st phase includes:
  - `POST api/v1/authenticate` which takes in the api_client or user credentials and issues a JWT
  - `GET api/v1/templates` which returns a list of all public published templates (high level metadata only ... future 'show' endpoint will return questions, guidance, etc.)
  - `POST api/v1/plans` creates a new plan. The specified `contact` is made the owner of the plan and is either emailed to let them know the plan was created (if they are an existing user) or are invited (if they are not). This expects the RDA Common Standard format
  - `GET api/v1/plans/:id` which returns the plan's metadata in the RDA Common Standard format
- Added JSON Web Token (JWT) authentication services (for use with both user and application/system authentication)
- added `start_date` and `end_date` to plans, added `grant_id` as an fkey to the `identifiers` table, `funder_id` and `org_id` as an fkey to the `orgs` table  to the `plans` table
- refactored `Identifier` a bit to make it a concern that is included by `Plan`, `Org` and `Contributor`. It also allows for a nil IdentifierScheme so that we can store identifiers that are not derived from an IdentifierScheme (e.g. a URL or ARK)
- updated `IidentifierScheme` to remove all of the `for_[x]` flags and replace with a BitFlag called `context`
- Added presenters for the V1 Jbuilder views
- Added Jbuilder views for V1
- Added new upgrade rake tasks that convert `grant_number` to `identifiers` and convert existing `data_contact`, `principal_investigator` and plan authors into `contributors`
- Added lots of tests
- Added rswag as a dev gem. Started swaggerizing V1 but have not finished

Also:
- Made OpenAire grant typeahead a configurable on/off option (on by default)
- Added `appendTo` logic to new Org AutoComplete Selector. Found an issue while testing with the DMPTool where Autocomplete appears in a few modal dialogs and default suggestion list was appearing outside the modal
- Fixed a bug in the usage dashboard's JS that was throwing errors when not on the usage page


briri:
this has gotten to be pretty monstrous. closing and will submit more atomic PRs instead

PR #2433, opened 2020-03-17
Issue/1682

xsrust:
addresses #1682 

**Changes Made:**
- generate usage-statistics for plans based on organizational templates to existing `StatCreatedPlan` model by extending the details hash
- extend plans_by_template helper functions and JS logic to generate the new template-usage chart
- fix test logic to deal with extended StatCreatedPlan models
- Re-work usage-dashboard to show the downloads buttons, then accordions for the template and user statistics, and then the filter below those.
- Some minor alignment fixes to the usage-dashboard
- If the org is a funder, the template_usage accordion starts expanded, otherwise the user_statistics accordion starts expanded


app/javascript/views/usage/index.js
```diff
@@ -37,25 +37,40 @@ $(() => {
...
   //       all other JS libraries were available to the js.erb files. Reevaluate
   //       this JS once we move to Rails 5 and properly configure webpacker
-  let drawnChart = null;
-  const monthlyPlanTemplatesChart = document.getElementById('monthly_template_plans');
+  let drawnChartByTemplate = null;
+  const monthlyPlanTemplatesChart = document.getElementById('monthly_plans_by_template');
   // Add event listeners that draw and destroy the chart
   monthlyPlanTemplatesChart.addEventListener('renderChart', (e) => {
-    drawnChart = drawHorizontalBar($('#monthly_template_plans'), e.detail);
+    drawnChartByTemplate = drawHorizontalBar($('#monthly_plans_by_template'), e.detail);
```
briri: such a weird mix of JQuery and regular JS. Will be happy to spend some time reevaluating all of our JS after we move to Rails 5!

app/models/stat_created_plan/create_or_update.rb
```diff
@@ -37,6 +38,11 @@ def plans(start_date:, end_date:)
         Plan.where(plans: { created_at: start_date..end_date })
       end
 
+      def own_template_plans(org)
+        template_ids = Template.where(org_id: org.id ).pluck(:id)
+        Plan.where(plans: { template_id: template_ids })
```
briri: Might be more efficient to get this in one query. Something like:
```ruby
Plan.includes(:template).joins(:template).where(template: { org_id: org.id })
```

app/views/usage/index.html.erb
```diff
@@ -1,98 +1,21 @@
...
-           locals: { user_count: @total_org_users, plan_count: @total_org_plans, separators: @separators } %>
-
-<%= render partial: 'usage/filter' %>
-
-
 <div class="row">
-  <div class="col-md-12">
-    <p class="red">&#42;&nbsp;<%= _('Move the mouse pointer over the bars of a chart to see numbers.') %></p>
+  <div class="col-md-12" >
+    <h1>Usage statistics</h1>
```
briri: Missing the internationalization `_("text")`

spec/services/org/create_last_month_created_plan_service_spec.rb
```diff
@@ -6,6 +6,9 @@
   let(:org) do
     FactoryBot.create(:org, created_at: DateTime.new(2018, 04, 01))
   end
+  let(:org2) do
+    FactoryBot.create(:org)
```
briri: we shouldn't need the `FactoryBot.` in these. Ok to leave for now since all of these specs have them. Can clean up later

app/javascript/views/usage/index.js
xsrust: Yeah, I figured I'd just mimic the existing syntax for now and then we could refactor all down the line.

spec/services/org/create_last_month_created_plan_service_spec.rb
xsrust: Good to know.
I think I need to spend a bit of time reading about factorybot/rspec best practices/intended usage.

briri: me too. There seem to be dozens of different approaches!

PR #2434, opened 2020-03-18
Issue 2346 - Create a super admin view of the Plans report

### LINKED ISSUE #2346: Create a super admin view of the Plans report
Currently it is difficult to get a high level view into all plans being created via the application. 

It would be very helpful to quickly get a report of all plans created. I think this can be done simply by creating a new Super Admin view for the Plans page that generates the same table but with data from all Organizations. 
screenshot-dmptool org-2020 01 09-14_55_15

martaribeiro:
Fixes #2346  .

Changes proposed in this PR:
-These changes will allow a super admin to view, search and sort all plans in the data base. The "download" button and the "view all' link have been deactivated as this would overload the code and it would take several sec if not minutes to complete these tasks.

### DIFF
--- a/app/controllers/paginable/plans_controller.rb
+++ b/app/controllers/paginable/plans_controller.rb
@@ -42,9 +42,15 @@ class Paginable::PlansController < ApplicationController
     unless current_user.present? && current_user.can_org_admin?
       raise Pundit::NotAuthorizedError
     end
+    #check if current user if super_admin
+    @super_admin = current_user.can_super_admin?
+    plans = @super_admin ? Plan.all : current_user.org.plans
+    plans = plans.joins(:template, roles: [user: :org]).where(Role.creator_condition)
+
     paginable_renderise(
       partial: "org_admin",
-      scope: current_user.org.plans,
+      scope: plans,
+      view_all: !current_user.can_super_admin?,
       query_params: { sort_field: 'plans.updated_at', sort_direction: :desc }
     )
   end
--- a/app/models/plan.rb
+++ b/app/models/plan.rb
@@ -164,8 +164,11 @@ class Plan < ActiveRecord::Base
 
   scope :search, lambda { |term|
     search_pattern = "%#{term}%"
-    joins(:template)
+    joins(:template, roles: [user: :org])
+    .where(Role.creator_condition)
     .where("lower(plans.title) LIKE lower(:search_pattern)
+            OR lower(orgs.name) LIKE lower (:search_pattern)
+            OR lower(orgs.abbreviation) LIKE lower (:search_pattern)
             OR lower(templates.title) LIKE lower(:search_pattern)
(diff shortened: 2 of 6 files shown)


app/controllers/paginable/plans_controller.rb
```diff
@@ -42,9 +42,15 @@ def org_admin
     unless current_user.present? && current_user.can_org_admin?
       raise Pundit::NotAuthorizedError
     end
+    #check if current user if super_admin
+    @super_admin = current_user.can_super_admin?
+    plans = @super_admin ? Plan.all : current_user.org.plans
+    plans = plans.joins(:template, roles: [user: :org]).where(Role.creator_condition)
```
briri: you'd likely want to do this joins+where in the plans_controller as well . Also consider using `includes` if the roles are displayed in the table layout. Includes forces eager loading whereas joins implements an `INNER JOIN` on the relation and uses lazy loading

app/models/plan.rb
```diff
@@ -164,8 +164,11 @@ class Plan < ActiveRecord::Base
 
   scope :search, lambda { |term|
     search_pattern = "%#{term}%"
-    joins(:template)
+    joins(:template, roles: [user: :org])
+    .where(Role.creator_condition)
```
briri: Not sure, but would be good to double check if this has an impact on searching for plans shared with someone who is not a creator? For example if you shared a plan with me as a commenter will this `where` prevent your plan from showing up in my search results?

xsrust: I think this would prevent any plans which were not created by the organization from showing up.  
Before, the initial scope was `current_user.org.plans` which selects along the following logic: https://github.com/DMPRoadmap/roadmap/blob/2aec36cfff386e79882584ce541d0fc46f452db4/app/models/org.rb#L241-L247

I think if this was changed to the `administrator_condition` it would have the side-effect of adding non-unique `plan` objects to the produced scope as there can be multiple `administrators` but only one `creator`.  In theory, this would be fine, but the extension is needed to query along the org-name column of the org-admin-plans-table.  Since we're displaying:
 https://github.com/DMPRoadmap/roadmap/blob/163668cea3b3545a83cde215a487ab0d3441076a/app/views/org_admin/plans/index.html.erb#L27
I think `creator_condition` might fit better here.

app/controllers/paginable/plans_controller.rb
xsrust: > you'd likely want to do this joins+where in the `plans_controller` as well

The joins+where is only needed for the sort/filter options (to ensure that the plan's creator's org can be queried), so I think should only be relevant to the paginable controller.  

> consider using `includes` if the roles are displayed in the table layout

That table's calling the `plan.owner` function to get an owner's name, which does it's own queries to `Role` and `User`, so I don't think the includes would help with access in this case

PR #2435, opened 2020-03-18
Issue 2382 - Add a active flag to the notification model 

### LINKED ISSUE #2382: Add an deactivate option to the notification system
There isn't an option to deactivate the notification once one has finished the downtime. Therefore the notification needs to be delete it instead.

One way to fix the issue would be to set an hour for the notification to expire, but it can be a problem with the time stamp in the server. Might be simpler to just had an active/inactive flag that can be changes as needed.

martaribeiro:
Fixes #2382  .

Changes proposed in this PR:
-Add a active flag to the notification model to allow super_admin to make a notification visible or not.


app/controllers/super_admin/notifications_controller.rb
```diff
@@ -55,6 +55,29 @@ def update
       render :edit
     end
 
+
+    # edit active field displayed in the table
+    def set_active
+
+      notification = Notification.find(params[:id])
+      notification.active = (params[:active] === "1")
+authorize(Notification)
```
xsrust: Please indent the `authorize` line to match the others.
It's mildly pedantic, but it would be good to check the authorization before modifying the record.

app/javascript/views/super_admin/notifications/edit.js
```diff
@@ -1,5 +1,39 @@
+<<<<<<< HEAD
+import { Tinymce } from '../../../utils/tinymce';
+// add the info on selecting the check from notification suitable
+=======
 import { Tinymce } from '../../../utils/tinymce.js.erb';
 
+//add the info on selecting the check from notification suitable
+>>>>>>> 8cebbdfe52128dbb2af7c3e1f7f1407be62795c4
```
xsrust: Looks like there's some lingering merge-conflicts in this bit.  Just take the latter section as it has the tinymce changes I'd made in development

app/javascript/views/super_admin/notifications/edit.js
```diff
@@ -1,5 +1,39 @@
...
+
+
+  $(paginableSelector).on('click, change', '.set_active_notification input[type="checkbox"]', (e) => {
+    const form = $(e.target).closest('form');
+    form.submit();
+  });
+
+
+  $(paginableSelector).on('ajax:success', '.set_active_notification', (e, data) => {
+    // const form = $(e.target);
```
xsrust: If this line isn't needed please remove instead of commenting out

app/javascript/views/super_admin/notifications/edit.js
```diff
@@ -1,5 +1,39 @@
...
+      notifier.renderAlert(data.msg);
+    }
+    /* remove or display the div
+    if (form.find('input[type="checkbox"]').is(':checked')) {
+      form.parent().siblings('.plan-visibility').html(getConstant('PLAN_VISIBILITY_WHEN_TEST'))
+        .attr('title', '');
+     } else {
+       form.parent().siblings('.plan-visibility').html(getConstant('PLAN_VISIBILITY_WHEN_NOT_TEST'))
+        .attr('title', getConstant('PLAN_VISIBILITY_WHEN_NOT_TEST_TOOLTIP'));
+    } */
```
xsrust: Please remove block comment

app/models/notification.rb
```diff
@@ -55,15 +58,15 @@ class Notification < ActiveRecord::Base
   # ==========
 
   scope :active, (lambda do
-    where('starts_at <= :now and :now < expires_at', now: Time.now)
+    where('starts_at <= :now and :now < expires_at', now: Time.now).where(active: true)
```
xsrust: This threw me for a loop for a minute.
I guess an instanciated class gets an accessor `active` to refer to the model,
And then the Plan class (or an activerecord relation) will have a `active` method referring to this scope.
I think the method/attribute are defined in different contexts so there shouldnt be any conflict

@briri have you ever run into any trouble doing something like this?

app/models/notification.rb
```diff
@@ -55,15 +58,15 @@ class Notification < ActiveRecord::Base
...
   end)
 
   scope :active_per_user, (lambda do |user|
     if user.present?
       acknowledgement_ids = user.notifications.map(&:id)
-      active.where.not(id: acknowledgement_ids)
+      active.where.not(id: acknowledgement_ids).where(active: true)
```
xsrust: The `where(active: true)` should be unnecessary as this chains off the `active` method above

app/models/notification.rb
```diff
@@ -55,15 +58,15 @@ class Notification < ActiveRecord::Base
...
     else
-      active.where(dismissable: false)
+      active.where(dismissable: false).where(active: true)
```
xsrust: The `where(active: true)` should be unnecessary as this chains off the `active` method above

app/views/paginable/notifications/_index.html.erb
```diff
@@ -14,7 +15,16 @@
...
+            <%= f.submit(_('Update'), style: 'display: none;') %>
+          <% end %>
+        </td>
       </tr>
     <% end %>
   </tbody>
 </table>
+
+
+<%# = notification.active? ? _('Yes') : _('No') %>
```
xsrust: Please remove debug line

app/javascript/views/super_admin/notifications/edit.js
xsrust: Ignore this, you've already made the change

app/javascript/views/super_admin/notifications/edit.js
martaribeiro: The webpacker was complaining about a variable that was not being used. I comment it out because I wasn't sure what is meant to do and couldn't see it being used anywhere

app/models/notification.rb
martaribeiro: @xsrust I don't know about this. I don't think I changed this?!

app/javascript/views/super_admin/notifications/edit.js
xsrust: Yeah, I'd just remove it instead of leaving the comment in.
If somebody wants to see what used to be there, they can always go through the github history, and it keeps the codebase a bit cleaner

app/models/notification.rb
xsrust: You've defined a field `active` on the model, which has the same name as this `scope :active`  
Your change to this method is just to ensure that the field is set to true, which makes sense.  I was more questioning if the overloading of that name would cause issue.  
I dont think it will, but brian's worked on more rails projects than I have:)

briri: Hmm. Yeah, doing `Notification.active` would invoke the scope over the attribute as far as I know. Would be interesting to test. `Notification.where(active: true)` though would refer to the attribute.

It may be cleaner to either rename the scope to something else or change this attribute to `enabled` or something.

app/views/super_admin/notifications/_form.html.erb
```diff
@@ -28,6 +28,13 @@
     </div>
   </div>
 
+  <div class="row">
+    <div class="form-group col-xs-12">
+      <%= f.check_box :active, style: 'width: auto' %>
```
briri: don't use style. Use a class instead. reducing `col-xs-12` down to 8 or 6 or whatever should do the trick

db/schema.rb
```diff
@@ -439,28 +438,12 @@
...
-  add_foreign_key "answers", "users"
-  add_foreign_key "answers_question_options", "answers"
-  add_foreign_key "answers_question_options", "question_options"
-  add_foreign_key "guidance_groups", "orgs"
-  add_foreign_key "guidances", "guidance_groups"
-  add_foreign_key "notes", "answers"
-  add_foreign_key "notes", "users"
-  add_foreign_key "notification_acknowledgements", "notifications"
-  add_foreign_key "notification_acknowledgements", "users"
-  add_foreign_key "org_identifiers", "identifier_schemes"
```
xsrust: Just noticed this removes a bunch of foreign_keys

db/schema.rb
```diff
@@ -126,7 +126,6 @@
   end
 
   add_index "notification_acknowledgements", ["notification_id"], name: "index_notification_acknowledgements_on_notification_id", using: :btree
-  add_index "notification_acknowledgements", ["user_id"], name: "index_notification_acknowledgements_on_user_id", using: :btree
```
xsrust: We shouldn't remove this index (not sure why the change in this commit would do that though)

briri:
looks like all of your requested changed on this one were addressed @xsrust

PR #2436, opened 2020-03-20
add user identifier column to user table

### LINKED ISSUE #2353: Add Orcid/Shibboleth display to admin display of user-accounts
Currently, it is not possible for an org/super-admin to tell if a user has connected their shibboleth or ORCiD accounts to their profile.

We should add either a display of this information, or a display with the ability to modify/remove (if that's desired), to the admin view of editing the user-profile.

raycarrick-ed:
Fixes #2353  .

Changes proposed in this PR:
- add column to org admin/superadmin users table which displays user identifiers both scheme name and actual identifier.


config/initializers/devise.rb
```diff
@@ -89,7 +89,7 @@
   	config.stretches = Rails.env.test? ? 1 : 10
 
   # Setup a pepper to generate the encrypted password.
-  config.pepper = "EXAMPLE OF PEPPER TO GENERATE THE ENCRYPTED PASSWORD"
+  config.pepper = "de451fa8d44af2c286d922f753d1b10fd23b99c10747143d9ba118988b9fa9601fea66bfe31266ffc6a331dc7331c71ebe845af8abcdb84c24b42b8063386530"
```
briri: careful to not commit things like this!

PR #2437, opened 2020-03-20
fix for XSS vulnerability

briri:
Added Rails4 monkey patch to address [CVE-2020-5267](https://nvd.nist.gov/vuln/detail/CVE-2020-5267) which is an XSS vulnerability in ActionView's `escape_javascript` method.

The patch was shared with us from the Dryad team. Thanks @sfisher

Since Rails 4.x is no longer receiving updates we need to patch this ourselves.

We will need to remove this initializer after moving to Rails 5+


xsrust:
Thanks for pulling this in @briri 
Manually patching things like this is another motivator for prioritizing moving to Rails 5.x :)

PR #2440, opened 2020-03-26
Issue/1682

xsrust:
- refactor the body of the accordions into a separate partial
- Conditionally call the accordions or just the body based on whether the user's org is a funder
- Made the JS more robust against the chart-data not existing on the page
- Remove the 'Monthly Usage' download for funders, and add in some spacing to make the top buttons align if removed.


app/javascript/views/usage/index.js
```diff
@@ -1,4 +1,4 @@
-import { isObject } from '../../utils/isType';
+import { isObject, isUndefined } from '../../utils/isType';
```
briri: 👍 I probably should have been using this one!

app/javascript/views/usage/index.js
```diff
@@ -24,33 +24,38 @@ $(() => {
   initializeCharts();
 
   // Create the Users joined chart
-  const usersData = JSON.parse($('#users_joined').val());
-  if (isObject(usersData)) {
-    createChart('#yearly_users', usersData);
+  if (!isUndefined($('#users_joined').val())) {
```
briri: good. I hope these checks become necessary in future versions of Rails' asset pipeline. It would be better if it just sent the client browser the JS it actually needs for the page.

PR #2444, opened 2020-04-02
Updates to Identifier model

briri:
Updates for initial Identifier work for #2339 

- Removed individual `for_[x]` flags on `identifier_schemes` and replaced with a `context` Bitflag 
- Created an Identifiable model concern and updated classes to us it
- Updates to Plan to convert `funder_name` into use of `funder.name`
- Various updates to address some bugs discovered while testing the Org Selector throughout the site
- Update to autoComplete.js to place suggestions within the context of the org selection box instead of in `<body>`. Issue was discovered while testing when selector was in a modal
- Added a new ApplicationService that handles retrieval of default Language and the Application name
- Refactored external_apis BaseService to use the Httparty gem (Openaire response was coming back chunked and existing Http::Net implementation would have needed additional work. Was easier to just use one of the popular http gems)
- Added a new stub DoiService and initializer
- Added an OpenAire initializer and refactored to move it inline with the other external_api services so that it can be enabled/disabled per installation
- Updated Plan pages to use the new Doi and OpenAire config


app/controllers/registrations_controller.rb
```diff
@@ -38,7 +38,7 @@ def new
           application_name: Rails.configuration.branding[:application][:name]
         }
         # rubocop:enable Metrics/LineLength
-        scheme = IdentifierScheme.by_name(oauth["provider"].downcase)
+        scheme = IdentifierScheme.by_name(oauth["provider"].downcase).first
         Identifier.create(identifier_scheme: scheme,
```
briri, line 42: These were bugs introduced in the original ROR work

app/javascript/views/plans/new.js
```diff
@@ -37,6 +37,7 @@ $(() => {
         if (data.templates.length === 1) {
           $('#plan_template_id option').attr('selected', 'true');
           $('#multiple-templates').hide();
+          $('#available-templates').fadeOut();
```
briri, line 40: Some JS bugs on the Plan create page from the initial ROR work (if a funder was selected, then the 'no funder' box was ticked it was not removing the loaded templates)

app/javascript/views/devise/registrations/new.js
```diff
@@ -0,0 +1,8 @@
+import { initAutocomplete } from '../../../utils/autoComplete';
+
+$(() => {
+  // Org selector on the /users/sign_up page that loads after a user
+  // signs in via institutional credentials but has no matching user record
```
briri, line 5: forgot to add this in the initial ROR work

app/javascript/views/usage/index.js
```diff
@@ -14,12 +14,15 @@ $(() => {
...
   // attach listener to separator select menu
   // on change look for "stat" elements and chnage their query param
-  document.getElementById('csv-field-sep').addEventListener('click', (e) => {
-    const statElems = document.getElementsByClassName('stat');
-    const newSep = 'sep='.concat(encodeURIComponent(e.target.value));
-    const changeStatFn = changeStatFnGen(newSep);
-    Array.from(statElems).forEach(changeStatFn);
-  });
+  const fieldSep = document.getElementById('csv-field-sep');
+  if (fieldSep !== null) {
```
briri, line 18: Lack of nil check was causing errors on non-usage pages

app/models/plan.rb
```diff
@@ -419,7 +418,7 @@ def owner
                   .administrator
                   .order(:created_at)
                   .pluck(:user_id).first
-    User.find(usr_id)
+    usr_id.present? ? User.find(usr_id) : nil
```
briri, line 421: was a bug in scenarios where the user_id (owner) is no longer attached to the plan

app/views/orgs/_profile_form.html.erb
```diff
@@ -48,7 +48,7 @@
 
       <% if org.logo.present? %>
         <div class="clearfix"></div>
-        <%= image_tag logo_url_for_org(org), alt: "#{org.name} #{_('logo')}" %>
+        <%= image_tag logo_url_for_org(org), alt: "#{org.name} #{_('logo')}", class: "org-logo" %>
```
briri, line 51: restricts logo size on the org details page

briri:
Travis had one of its random errors. Can't seem to restart it though on their site 🤷‍♂

app/javascript/views/usage/index.js
xsrust: thanks for adding that

app/models/identifier.rb
```diff
@@ -27,21 +27,27 @@ class Identifier < ActiveRecord::Base
 
   belongs_to :identifiable, polymorphic: true
 
-  belongs_to :identifier_scheme
+  # TODO: uncomment 'optional: true' once we are on Rails 5
+  belongs_to :identifier_scheme #, optional: true
```
xsrust, line 31: Thanks for adding these forward-looking TODO's

app/models/identifier.rb
```diff
@@ -27,21 +27,27 @@ class Identifier < ActiveRecord::Base
...
 
+  validate :value_uniqueness_with_scheme, if: :has_scheme?
+
+  validate :value_uniqueness_without_scheme, unless: :has_scheme?
+
+  # =============
+  # = Callbacks =
+  # =============
+
+  before_save :append_scheme_prefix
```
xsrust, line 49: Would overriding the default accessors work here like you'd done in the other ror-fundref PR?  I liked how that could be tested by changing the value on an instanciated record rather than needing to test the function that the callback calls.

https://api.rubyonrails.org/classes/ActiveRecord/Base.html#class-ActiveRecord::Base-label-Overwriting+default+accessors

app/models/plan.rb
xsrust: Thanks.
I've noticed we do some inconsistent access around the display of a plan's owner due to the same thing

app/models/identifier.rb
briri: 👍 we should run a search after the upgrade and address them (or remove if they are no longer relevant)

app/models/identifier.rb
briri: yess absolutely. will change this

briri:
- rebased off of ror-fundref 
- Converted callback on identifier model to a `value=` accessor override

briri:
looks like Travis' postgres test failed when trying to install node. I'm not given the option to restart the run 🤷‍♂

briri:
travis is failing randomly. Not sure why I can't restart the job. I'm thinking maybe we should just jettison Travis now that we have the Github Workflow tests going

briri:
apparently needed to sign out and then back in to Travis for the restart button to appear

Issue #2419, opened 2020-03-06
plan.funder_name not specified for all funder templates


sjDCC:
We recently changed the DMP cover page to list the funder name and the template name. See #2371 This is working in most cases but occassionally the funder_name attribute is not always specified

According to @xsrust this gets dynamically set on plan-creation. It'd be good to get a dev to have a look at why that might not have been set. The dev can create a rake:task to clean up our data.

sjDCC:
May no longer be needed after @briri recent work on RoR.

Devs to check and close

briri:
Updated to use the name of the Org who owns the Template if the user has not supplied a Funder on the Project Details page and the Org is a funder.
Added fix to the existing project-start-end PR since it is built off of the plan.funder relationship instead of the old plan.funder_name string

sjDCC:
All seems good. Closing out

PR #2446, opened 2020-04-03
Added Project start and end dates and made grant an identifier

### LINKED ISSUE #2409: Add Start and End dates to Project Details
The [RDA Common Standard](https://github.com/RDA-DMP-Common/RDA-DMP-Common-Standard/tree/v1.0#funding_tree) includes a Project start and end date. The fields have been added to the table as part of #2390.

We should add fields to the Project Details page to capture this information from users.

briri:
Fixes #2409 
- Added `start_date` and `end_date` to `plans`
- Converted `grant_number` over to an `identifier`
- Updated model, tests, views accordingly


briri:
Added a fix for #2419 so that the plan export coversheet will show the template.org.name if the plan.funder is not set

app/models/concerns/exportable_plan.rb
```diff
@@ -97,7 +97,9 @@ def prepare_coversheet
     hash[:affiliation] = self.owner.present? ? self.owner.org.name : ""
 
     # set the funder name
-    hash[:funder] = self.funder.present? ? self.funder.name : ""
+    hash[:funder] = self.funder.name if self.funder.present?
+    template_org = self.template.org
+    hash[:funder] = template_org.name if !hash[:funder].present? && template_org.funder?
```
xsrust, line 102: Thanks for adding the funder_name setting here

app/models/identifier.rb
```diff
@@ -83,8 +83,7 @@ def value_without_scheme_prefix
                         identifier_scheme.identifier_prefix.present?
 
     base = identifier_scheme.identifier_prefix
-    value.gsub(base, "").sub(%r{^\/}, "")
-  end
+    value.gsub(base, "").sub(%r{^\/}, "")  end
```
xsrust, line 86: is removing the newline a typo?

app/models/plan.rb
```diff
@@ -28,20 +28,24 @@
...
 # Foreign Keys
 #
 #  fk_rails_...  (template_id => templates.id)
 #  fk_rails_...  (org_id => orgs.id)
 #
 
-# TODO: Drop the funder_name column once the funder_id has been back
-#       filled and we're removing the is_other org stuff
+# TODO: Drop the funder_name and grant_number columns once the funder_id has
+#       been back filled and we're removing the is_other org stuff
```
xsrust, line 48: We can do a pass through the code for all these TODO's after the Rails 5 migration.
Thanks for keeping these up to date

app/views/plans/new.html.erb
```diff
@@ -49,7 +49,6 @@
       <div id="research-org-controls" class="row">
         <div class="form-group col-md-6">
            <em class="sr-only"><%= research_org_tooltip %></em>
-           <%# TODO: create a relationship on the plans table to org for this (org_id)! %>
```
xsrust, line 52: glad we can be removing some of these as well:)

app/models/plan.rb
```diff
@@ -566,6 +570,13 @@ def landing_page
     identifiers.select { |i| %w[doi ark].include?(i.identifier_format) }.first
   end
 
+  # Returns the identifier associated with the grant_id
+  def grant
+    return nil unless grant_id.present?
+
+    identifiers.select { |identifier| identifier.id == grant_id }.first
+  end
```
xsrust, line 578: I'd like to handle this through Rails associations if we can.  would something like the following work?
```
has_one :grant, class_name: 'Identifier'
```
The id on the identifier should still be unique to that table, so it should be enough to build-out that reference.

db/migrate/20200218213414_add_start_and_end_dates_to_plans.rb
```diff
@@ -0,0 +1,7 @@
+class AddStartAndEndDatesToPlans < ActiveRecord::Migration
+  def change
+    add_column :plans, :grant_id, :integer, index: true
```
xsrust, line 3: If you end up changing grant to an association, this may need to change to a reference to enforce the foreign_key 
`add_reference :plans, :grant`

app/models/identifier.rb
briri: yes. good catch

app/models/plan.rb
briri: agreed. will give it a try

briri: The following seems to work:
`has_one :grant, as: :identifiable, dependent: :destroy, class_name: "Identifier"`
I did not need to update the schema migration. 

I also added some sanity checks in the `models/plan_spec.rb` to ensure things work properly

briri:
going to hold off refactoring and resubmitting this one until `api-clients` and `contributors` have been merged into `development`

Issue #2450, opened 2020-04-07
Add a filter to the user / plans page tables


sjDCC:
On the two admin pages for "users" and "plans" add a filter that allows admins to search statistics

The primary reason we're adding this is to enable a click through from the usage dashboard so admins can interrogate specific peaks or troughs in the data. We expect most people to simple land on the results page from the filter, rather than using the filter directly themselves.
![peaks](https://user-images.githubusercontent.com/6169940/79246846-a0a82600-7e71-11ea-94d5-5abf86dfed88.PNG)

On the two tables, we feel the filter would be best to be fairly discrete e.g. a cogwheel in the table header that opens a modal which allows you to set search or select the month from a dropdown list. We may face a couple of issues in adding this as the plans table doesn't have a "month created" column, rather "last updated". 
![filter in table](https://user-images.githubusercontent.com/6169940/79247168-0f857f00-7e72-11ea-89d0-78291a30ca53.PNG)

Another option is to add the filter as another search bar entry but we think this could get confusing and clutter up the UI so expect something in the table header would be better as the filter will only work on certain fields. @mariapraetzellis will share some example options from xls and salesforce
![cogwheel](https://user-images.githubusercontent.com/6169940/79247191-17452380-7e72-11ea-9456-c0c94d520b74.PNG)

More spec ideas to come

mariapraetzellis:
Here's a screenshot of what I was thinking about in terms of making the columns filterable w/ the cog icon and sortable w/ the arrow. I'm not sure this would work in terms of pulling out users from a specific date range though. 
![screenshot-partner archive-it org-2020 04 14-13_54_17 (1)](https://user-images.githubusercontent.com/27511575/79273576-130c0c80-7e58-11ea-8c66-b66771d48949.png)

briri:
spoke with @mariapraetzellis and agreed that we can update the existing search box to allow users to provide a 'month and year' or a 'year' (e.g. 'Oct 2016' or '2016').

briri:
Here is the text that appears on the Plans and Users pages:

New message about filtering (will change this in the new ticket and include a list of fields that are searchable): "_Note: You can filter this table by 'Created' dates. Enter the month abbreviation and a 4 digit year into the search box above.<br>For example: 'Oct 2019' or 'Jun 2013'._"

[...]

PR #2453, opened 2020-04-09
Usage Dashboard: remove filter section and allow click-through from charts

### LINKED ISSUE #2343: Improvements to the Usage Statistics page and graphs
The current Usage Statistics page layout is not intuitive and it is not obvious that the filter is not connected to the graphs below. 

We propose that the filter lets the admin define settings which are then visualized in graphs, with possibility to click through to get more information / interrogate the underlying data. 

Usage Graph

This involves:
1. linking the filter and graphics below
2. Remove the "Topic" from the filter. Users will then only be able to select a date span. 
3. Investigate JS library through which charts are created to see if we can offer a click through experience to investigate the underlying data

This issue builds on https://github.com/DMPRoadmap/roadmap/issues/1678.

briri:
Fixes #2343 #2450.

Changes proposed in this PR:
- Removed the old filter section
- Added a new DateRangeable model concern and included it on Plans and User and updated their `search` scopes
- Updated JS to allow user to click-through to the plans/users pages
  - calls to the plans/users page now sends a `search=[mm yyyy]` arg in the query string to tie into the existing paginable search logic
  - updated controllers to handle carry through a `@clicked_through` variable that will determine whether or not to display a message that explain why the number of records in the table may differ from the number shown on the usage dashboard (historical versus current data)
  - added tests for the new concern


app/views/paginable/plans/_org_admin.html.erb
```diff
@@ -1,3 +1,13 @@
+<% if @filter.present? %>
+  <p><%= _(<<-TEXT
+    The data on the usage dashboard is historical in nature. This means that the number of records below may not
+    match the count shown on the usage dashboard. For example if one of your users created a plan in October and
+    then removed that plan in November, it would have been included on the usage dashboard's total for October but
+    would not appear in the list below.
+  TEXT
+  ) %></p>
```
xsrust, line 8: I'll be interested to see if our gettext rake tasks will pick up on this text to translate

briri:
@xsrust I made some modifications to this one to incorporate the work for the filter on the Plans and Users pages.

xsrust:
Thanks for refactoring that logic out to the `daterangeable` conern.
The updates look good to me

PR #2455, opened 2020-04-13
removed debug  from monthly usage stat service and pass in actual cu…

briri:
Removed a debug/test override of `current_user` in the MonthlyUsageStatsService and updated controller so that it passes this value in to the service.

### DIFF
--- a/app/services/org/monthly_usage_service.rb
+++ b/app/services/org/monthly_usage_service.rb
@@ -6,19 +6,15 @@ class Org
 
     class << self
 
-      def call
-        total = build_from_joined_user
-        build_from_created_plan(total)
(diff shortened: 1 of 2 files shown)


app/services/org/monthly_usage_service.rb
```diff
@@ -6,20 +6,16 @@ class MonthlyUsageService
...
+        build_from_exported_plan(current_user, total)
         total.values
       end
 
       private
 
-      def current_user
-        User.find_by_email 'xsrust@gmail.com'
-      end
-
```
xsrust, line 22: oops, good catch, Thanks @briri

Issue #2390, opened 2020-02-10
Build out API v1 - Phase one


briri:
Phase one of the API v1 changes involves setup of the basic infrastructure and a few initial endpoints.

### Infrastructure:
- Introduce JSON Web Token (JWT) security for system to system communications. Which will use a `client_id` and `client_secret` for initial auth and then a JWT in further requests.
- Create a new `api_client` table to handle JWT machine to machine security
- Allow access to the API for both ApiClients (systems) and Users
- Create new views that follow the [RDA common metadata standard for DMPs](https://github.com/RDA-DMP-Common/RDA-DMP-Common-Standard/tree/v1.0)
- Create views that output the RDA common metadata standard for DMPs

### Endpoints
- Build out `POST /api/v1/authenticate` - to authenticate the user and return a JWT
- ~~Build out `GET /api/v1/plans` - to retrieve a list of DMPs~~ This one will come later
- Build out `GET /api/v1/plans/[:id]` - to retrieve a DMP
- Build out `POST /api/v1/plans` - to create a DMP

briri:
This ticket will also provide a `GET /api/v1/templates` endpoint that returns the titles, descriptions, versions and owning organization names for all public templates.

Will create further tickets for carrying the existing API v0 functionality into API v1 (e.g. full text API, etc.)

briri:
The v1 API uses JWT authorization and allows both the existing User `email` + `api_token` or the new ApiClient (for systems that need to communicate with DMPRoadmap) `client_id` + `client_secret`.

sjDCC:
Deployed out to DMPtool stage for testing with RAMS

sjDCC:
Is this one the devs can approve please @xsrust Need a tutorial on APIs to do this myself!

sjDCC:
Closing this based on dev check

PR #2456, opened 2020-04-16
API Clients table and UI administration screens

briri:
Part of #2390 

- Adds an `api_clients` to store access credentials for systems (instead of individual users)
- Adds a model for ApiClient
- Adds a controller, policy, routes and corresponding views (accessible via the Super Admin menu)
  - Main Index page (index and paginable index)
  - Edit/Add pages
  - Ability to Remove an ApiClient from the index page
  - Ability to refresh client_id and client_secret on the Edit Page
  - Ability to auto-email credentials to the contact on the Edit Page
- Adds tests 
- Added a `last_api_access` to User for future use with API V1

### DIFF
--- a/lib/tasks/upgrade.rake
+++ b/lib/tasks/upgrade.rake
@@ -772,7 +772,6 @@ namespace :upgrade do
       ui.user.identifiers << Identifier.new(
         identifier_scheme: ui.identifier_scheme,
         value: ui.identifier,
-        #type: (ui.identifier_scheme.name == "orcid" ? "url" : "other"),
         attrs: {}.to_json
       )
     end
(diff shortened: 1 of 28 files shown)


lib/tasks/upgrade.rake
```diff
@@ -772,7 +772,6 @@ namespace :upgrade do
       ui.user.identifiers << Identifier.new(
         identifier_scheme: ui.identifier_scheme,
         value: ui.identifier,
-        #type: (ui.identifier_scheme.name == "orcid" ? "url" : "other"),
```
xsrust, line 775: Thanks for cleaning these up

PR #2457, opened 2020-04-16
Support for multiple Contributors

### LINKED ISSUE #2349: Support for multiple contributors

briri:
Fixes #2349.

- Add new `contributors` table with f-keys on Plan and Org
- Add Contributor model
- Add new routes, controllers, javascript and corresponding views
- Added tests


app/controllers/api/v0/plans_controller.rb
```diff
@@ -27,13 +27,15 @@ def create
...
 
-    @plan.data_contact = plan_user.email
+    # Attach the user as the PI and Data Contact
+    @plan.contributors << Contributor.new(
+      name: [plan_user.firstname, plan_user.surname].join(" "),
+      email: plan_user.email,
+      investigation: true,
+      data_curation: true
+    )
+
```
xsrust, line 38: Thanks for updating the old API to be compliant with the new structure

app/controllers/contributors_controller.rb
```diff
@@ -0,0 +1,190 @@
...
+
+  # =============
+  # = Callbacks =
+  # =============
+  def fetch_plan
+    @plan = Plan.includes(:contributors).find_by(id: params[:plan_id])
+    return true if @plan.present?
+
+    redirect_to root_path, alert: _("plan not found")
+  end
```
xsrust, line 151: I really like this style of fetching data.  It would be nice to convert some of the older controllers to conform to this style rather than fetching/validating the data per-action

app/views/plans/_edit_details.html.erb
```diff
@@ -1,288 +1,23 @@
...
-            <%= f.hidden_field :visibility %>
-            <%= f.label(:is_test, class: 'control-label') do %>
-              <%= check_box_tag(:is_test, 1, @plan.is_test?, "aria-label": "is_test") %>
-              <%= _('mock project for testing, practice, or educational purposes') %>
-            <% end %>
-          </div>
-        </div>
-      </div>
+      <%= render partial: "plans/project_details",
+                 locals: { form: f, plan: plan } %>
```
xsrust, line 8: I did not realize that all of this functionality was in the one view-file.
Thanks for breaking this out into partials

PR #2459, opened 2020-04-16
Bugfix: Template creation was failing due to invalid visibility value

### LINKED ISSUE #2458: Bug with creating additional custom templates
An org admin from DOT reported that she was only able to create 5 custom templates and then the system prevented her from making any others.

Brian confirmed there is a bug in Roadmap, he submitted a fix to DMPTool and will move to Roadmap.

briri:
Fixes #2458 .

This only impacts Orgs that have multiple Org types (e.g. both a Funder and an Institution). 

Changes proposed in this PR:
- Add line back into the templates controller that converts the `:visibility` checkbox from a '0' or '1' into `organizationally_visible` or `publicly_visible`.

### DIFF
--- a/app/controllers/org_admin/templates_controller.rb
+++ b/app/controllers/org_admin/templates_controller.rb
@@ -150,8 +150,12 @@ module OrgAdmin
     # POST /org_admin/templates
     def create
       authorize Template
+      args = template_params
+      # Swap in the appropriate visibility enum value for the checkbox value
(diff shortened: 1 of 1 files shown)


xsrust:
Thanks @briri  
We'd had people's templates which had converted to `publicly_visible` and were not showing up from within the create-plan wizard, but I'd thought it was just a data-issue with some of the older templates

PR #2484, opened 2020-04-24
DO NOT MERGE

briri:
Just resubmitting to see what changes were made so that I can adjust them for the new Identifier table


briri:
never mind. looks like it was based off the conditional questions branch

PR #2486, opened 2020-04-27
Allow creation and management of departments from the API

xsrust:
I've added 4 endpoints in v0 of the api to a new controller: `api/v0/departments_controller.rb`
- `index` displays an array of departments for any user's organization
- `create` allows an org-admin to specify a name and code to create a department
- `users` allows an org-admin to see which accounts (by email) are associated with each department.
- `assign_users` allows an org-admin to specify a list of users (`user[]` in params) to be assigned to a specific department (using the id returned by `index` and the email addresses from `users`)

### DIFF
--- a/app/controllers/api/v0/departments_controller.rb
+++ b/app/controllers/api/v0/departments_controller.rb
@@ -1,0 +1,71 @@
+# frozen_string_literal: true
+
+class Api::V0::DepartmentsController < Api::V0::BaseController
+
+  before_action :authenticate
+
+  ##
+  # Create a new department based on the information passed in JSON to the API
+  def create
+    unless Api::V0::DepartmentsPolicy.new(@user, nil).index?
+      raise Pundit::NotAuthorizedError
+    end
+    @department = Department.new(org: @user.org,
+                                 code: params[:code],
+                                 name: params[:name])
+    if @department.save
+      redirect_to api_v0_departments_path
+    else
+      # the department did not save
+      self.headers["WWW-Authenticate"] = "Token realm=\"\""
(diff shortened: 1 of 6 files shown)


app/controllers/api/v0/departments_controller.rb
```diff
@@ -0,0 +1,71 @@
...
+      raise Pundit::NotAuthorizedError
+    end
+    @department = Department.new(org: @user.org,
+                                 code: params[:code],
+                                 name: params[:name])
+    if @department.save
+      redirect_to api_v0_departments_path
+    else
+      # the department did not save
+      self.headers["WWW-Authenticate"] = "Token realm=\"\""
```
briri, line 20: 👍 this great. The v1 json responses contain the error messages but we should probably also list them in the 'realm' as well.

app/controllers/api/v0/departments_controller.rb
```diff
@@ -0,0 +1,71 @@
...
+    @users = @user.org.users.includes(:department)
+  end
+
+  def assign_users
+    @department = Department.find(params[:id])
+
+    unless Api::V0::DepartmentsPolicy.new(@user, @department).assign_users?
+      raise Pundit::NotAuthorizedError
+    end
+
```
briri, line 47: Should probably do a nil check on `@department` here to make sure it doesn't accidentally blank out associations if the id was incorrect for some reason.

xsrust: Sure, I've added that on the policy check for assign_users?

briri:
this looks fine @xsrust looks like it just needs a rebase. feel free to merge after that

PR #2493, opened 2020-04-28
fixes bug on the org_admin plans page 

xsrust:
where plans without owners would cause the page load to fail
noted in #2450 and #2346


briri:
looks fine @xsrust so I am going to merge. 

The tests were failing with weird errors (one is an intermittent error we've been seeing. I added a new ticket to address the failure in the Rails 5 work) so I restarted them.

PR #2508, opened 2020-05-13
Issue 2352 Additional filters for super-admin user index

### LINKED ISSUE #2352: Additional filters for super-admin user index
We recently had a bug report from an org-admin user who was missing a permission.  The account associated with the email address they had contacted us on had full org-admin perms.  After some digging on the Rails Console, it turned out the user had 2 different accounts, based off different email aliases, and they logged in with shibboleth credentials, so were unnaware of which account they were logging into.

I first attempted to interrogate this issue through our UI, but was quickly stopped by our super-admin `users` index page.  While I could sort by organisation, I could not filter to only see one organisation's users.  Further, I could not sort the users by the ones with permissions, to enable me to see how many admin users there are for a given organisation.

The organisational-filter, along with either a sort by permissions, or a filter for permissions (to easily see all org/super-admin users on your instance) would have been helpful.

martaribeiro:
Fixes #2352 

Changes proposed in this PR:
To add an organisational-filter, along with a filter for permissions (to easily see all org/super-admin users on your instance) 

A solution to this problem was to allow users to search by organisation and by adding a check box that filters users with permissions. For example, a “super-admin’ can search for an UoE user with  the ability  to “add organisations”, “modify templates” and “modify guidance”. The way the filter by permissions work is by looking for users with one or more types of permissions. 

There was a javascript validation to only submit the search for when there was search text. This was stopping the permissions filter to submit. The decision to remove this validation was not only to allow the filter to work but as well to allow the user to clear the search box in order to clear the search results.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -149,8 +149,11 @@ class User < ActiveRecord::Base
               "lower(email) LIKE lower(?)",
               search_pattern, search_pattern)
       else
-        where("lower(firstname || ' ' || surname) LIKE lower(?) OR " +
-              "email LIKE lower(?)", search_pattern, search_pattern)
+        joins(:org)
+          .where("lower(firstname || ' ' || surname) LIKE lower(:search_pattern)
+              OR lower(email) LIKE lower(:search_pattern)
+              OR lower(orgs.name) LIKE lower (:search_pattern)
+              OR lower(orgs.abbreviation) LIKE lower (:search_pattern) ", search_pattern: search_pattern)
       end
     end
   }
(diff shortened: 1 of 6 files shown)


briri:
think this one should be a PR to development (it looks like it was based off of that branch).

app/models/user.rb
```diff
@@ -149,8 +149,11 @@ class User < ActiveRecord::Base
               "lower(email) LIKE lower(?)",
               search_pattern, search_pattern)
       else
-        where("lower(firstname || ' ' || surname) LIKE lower(?) OR " +
-              "email LIKE lower(?)", search_pattern, search_pattern)
+        joins(:org)
+          .where("lower(firstname || ' ' || surname) LIKE lower(:search_pattern)
+              OR lower(email) LIKE lower(:search_pattern)
+              OR lower(orgs.name) LIKE lower (:search_pattern)
```
briri, line 155: Thanks for fixing this

PR #2510, opened 2020-05-13
fixed bug preventing the read only version of the template

### LINKED ISSUE #2509: Unable to view a historical version of a template

**Expected behaviour:**
Admins should be able t view historical versions of a template.

**Actual behaviour:**
The sections/questions are not visible on the 'Write Plan' tab
Screen Shot 2020-05-13 at 10 14 37 AM

**Steps to reproduce:**
- Log in as an Admin
- Navigate to the Templates page and click on the 'History' link for one of the Organizational templates 
- Click on the 'View' link of an old version of the template
- Navigate to the 'Write Plan' tab

briri:
Fixes #2509 .

Changes proposed in this PR:
- Allow an admin to view the historical contents of a template

### DIFF
--- a/app/views/org_admin/phases/container.html.erb
+++ b/app/views/org_admin/phases/container.html.erb
@@ -63,7 +63,7 @@
                         <% unless phase.sections.all?(&:modifiable?) %>
                           <%= _("You may place them before or after the main template sections.") %>
                         <% end %>
-                      <% else %>
+                      <% elsif template.latest? %>
                         <%= link_to _('Re-order sections'),
                             org_admin_template_phase_versions_path(phase.template, phase),
                             method: "post",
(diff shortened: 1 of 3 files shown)


app/views/org_admin/phases/container.html.erb
```diff
@@ -63,7 +63,7 @@
                         <% unless phase.sections.all?(&:modifiable?) %>
                           <%= _("You may place them before or after the main template sections.") %>
                         <% end %>
-                      <% else %>
+                      <% elsif template.latest? %>
```
xsrust, line 66: So this hides the link to re-order sections on Historic Templates?
Good catch, Just checked on DMPonline and hitting that throws a 404

PR #2512, opened 2020-05-13
fixed issue with giving OrgAdmin permission to Change Org Affiliation

### LINKED ISSUE #2511: Granting one SuperAdmin permission gives user all SuperAdmin permissions

**Expected behaviour:**
Giving an OrgAdmin the 'Manage Org Affiliation' permission should allow the user to see the 'Change Affiliation' box on the Templates page, not remove their existing permissions when they change Org affiliation, and NOT provide access to all other SuperAdmin functionality (e.g. Notifications and Themes)

**Actual behaviour:**
Giving the user the one permission results in access to all SuperAdmin functionality.

This is due to the fact that we just do a check for `.can_super_admin?` in most places which verifies that the user has at least one SuperAdmin permission. We should: 
- Update the `can_super_admin?` method so that it only returns true if ALL SuperAdmin permissions are granted.
- Update the code to check the individual permission where appropriate instead of just `can_super_admin?`
- Consider adding additional permissions

**Steps to reproduce:**
- Give a test OrgAdmin user account the 'Manage Org Affiliation' permission.
- Login as that User
- Access the Notifications tab

briri:
Fixes #2511 

- Update `can_super_admin?` to verify that user has ALL SuperAdmin permissions
- Update template page's Org Selector to care about `can_change_org?` instead of `can_super_admin?`

We may want to consider a re-eval of our permissions and policies. This was sort of an edge case scenario.

Hold off on merging this one @xsrust until we have a conversation about the scenario outlined in the ticket and any security implications created by these changes.

### DIFF
--- a/app/javascript/views/usage/index.js
+++ b/app/javascript/views/usage/index.js
@@ -14,12 +14,15 @@ $(() => {
 
   // attach listener to separator select menu
   // on change look for "stat" elements and chnage their query param
-  document.getElementById('csv-field-sep').addEventListener('click', (e) => {
-    const statElems = document.getElementsByClassName('stat');
-    const newSep = 'sep='.concat(encodeURIComponent(e.target.value));
-    const changeStatFn = changeStatFnGen(newSep);
-    Array.from(statElems).forEach(changeStatFn);
-  });
+  const fieldSep = document.getElementById('csv-field-sep');
+  if (fieldSep !== null) {
+    fieldSep.addEventListener('click', (e) => {
+      const statElems = document.getElementsByClassName('stat');
+      const newSep = 'sep='.concat(encodeURIComponent(e.target.value));
+      const changeStatFn = changeStatFnGen(newSep);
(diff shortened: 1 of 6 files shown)


app/javascript/views/usage/index.js
```diff
@@ -14,12 +14,15 @@ $(() => {
 
   // attach listener to separator select menu
   // on change look for "stat" elements and chnage their query param
-  document.getElementById('csv-field-sep').addEventListener('click', (e) => {
-    const statElems = document.getElementsByClassName('stat');
-    const newSep = 'sep='.concat(encodeURIComponent(e.target.value));
-    const changeStatFn = changeStatFnGen(newSep);
-    Array.from(statElems).forEach(changeStatFn);
-  });
+  const fieldSep = document.getElementById('csv-field-sep');
```
briri, line 17: We patch this in development as well. Just pulled it forward for master

PR #2533, opened 2020-05-29
init list of maDMP endpoint

xsrust:
Also proposes a pundit-based method for scoping access to

### DIFF
--- a/app/controllers/api/v1/plans_controller.rb
+++ b/app/controllers/api/v1/plans_controller.rb
@@ -87,6 +87,22 @@ module Api
       # rubocop:enable Metrics/AbcSize, Metrics/CyclomaticComplexity
       # rubocop:enable Metrics/PerceivedComplexity, Metrics/MethodLength
 
+      # GET /api/v1/plans
+      def index
+        # ALL can view: public
+        # ApiClient can view: anything from the API client
+        # User (non-admin) can view: any personal or organisationally_visible
+        # User (admin) can view: all from users of their organisation
+        plans = Api::V1::PlansPolicy::Scope.new(client, Plan).resolve
+        if plans.present? && plans.any?
+          @items = paginate_response(results: plans)
+          @minimal = true
+          render "api/v1/plans/index", status: :ok
+        else
+          render_error(errors: [_("No Plans found")], status: :not_found)
+        end
+      end
+
       private
 
       def dmp_params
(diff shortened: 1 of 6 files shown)


app/controllers/api/v1/plans_controller.rb
```diff
@@ -87,6 +87,22 @@ def create
       # rubocop:enable Metrics/AbcSize, Metrics/CyclomaticComplexity
       # rubocop:enable Metrics/PerceivedComplexity, Metrics/MethodLength
 
+      # GET /api/v1/plans
+      def index
+        # ALL can view: public
```
briri, line 92: Thanks for putting this outline of permissions here. I think it will be really useful going forward when we're trying to debug issues

app/controllers/api/v1/plans_controller.rb
```diff
@@ -87,6 +87,22 @@ def create
...
+        # ApiClient can view: anything from the API client
+        # User (non-admin) can view: any personal or organisationally_visible
+        # User (admin) can view: all from users of their organisation
+        plans = Api::V1::PlansPolicy::Scope.new(client, Plan).resolve
```
briri, line 96: 👍

app/models/api_client.rb
```diff
@@ -25,6 +25,12 @@ class ApiClient < ActiveRecord::Base
   include DeviseInvitable::Inviter
   include ValidationMessages
 
+  # ================
+  # = Associations =
+  # ================
+
+  has_many :plans
```
briri, line 32: thanks for adding this missing association

app/views/api/v1/plans/_show.json.jbuilder
```diff
@@ -34,32 +34,34 @@ if presenter.data_contact.present?
   end
 end
 
-if presenter.contributors.any?
-  json.contributor presenter.contributors do |contributor|
-    json.partial! "api/v1/contributors/show", contributor: contributor,
-                                              is_contact: false
+unless @minimal
```
briri, line 37: nice. makes a lot of sense to return just enough info for this one to keep the response size down

briri:
the postgres test is failing and it looks like it's maybe due to an ordering issue (running migrate before setup). 
I will have a look at that today and try to pull in some of the things I learned from the guithub conf

PR #2534, opened 2020-06-01
Google analytics

### LINKED ISSUE #2344: Google Analytics integration
We have received many requests for further usage statistics. Many of these requests have been tracked in https://github.com/DMPRoadmap/roadmap/issues/1745 For example:
No. of sessions
Average time in session / editing
No. of times working on one DMP

Adding a feature to enable organizational level Google Analytics would allow for an easy way to track usage data. 

The organizational admin would need to be able to enable google analytics on the Roadmap web application by adding their GA tracking code. The user would then have access to the stats via the GA application and it would be the responsibility on universities to analyze the results and create the required reports.

raycarrick-ed:
Fixes #2344 .

Changes proposed in this PR:
- add google analytics tracker to org edit page
- configure root org in application.rb
- if root org has a tracker defined, plant it on every page
- if current user org has a tracker defined plant that too


config/application.rb
```diff
@@ -122,6 +122,9 @@ class Application < Rails::Application
       config.branding = config_for(:branding).deep_symbolize_keys
     end
 
+    # org abbreviation for the root google analytics tracker that gets planted on every page
+    config.tracker_root = "UoE"
```
briri, line 126: Probably want something more generic here like 'DMPRoadmap' so that other installations aren't using UoE by default.

app/models/org.rb
```diff
@@ -64,6 +64,10 @@ class Org < ActiveRecord::Base
 
   belongs_to :region
 
+  has_one :tracker
+  accepts_nested_attributes_for :tracker 
+  validates_associated :tracker
```
briri, line 69: Nice! I did not know about this `validates_associated`.

can you add a `dependent: :destroy' to the `has_one` statement? This should prevent any orphaning of records

spec/models/tracker_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
+
+RSpec.describe Tracker, type: :model do
+  pending "add some examples to (or delete) #{__FILE__}"
```
briri, line 4: Would be great to add some simple tests here. Check out some of the other model tests like `spec/models/contributor_spec.rb` for examples of association and validation tests.
You can test that enum with something like this: `it { is_expected.to define_enum_for(:category).with(Identifier.categories.keys) }`

config/application.rb
raycarrick-ed: Agh! that wasn't intentional

PR #2536, opened 2020-06-02
API bug fixes (from hackathon) and pundit changes

briri:
Fixed a few issues that came out of the hackathon testing.

- Re-removed Feedjira gem from Gemfile.lock. Not sure how/when it was reintroduced

Pundit:
- Update existing endpoints to use the new Pundit Policy @xsrust created.
- Added Pundit Policies `api/v1/template_policy.rb`, `contributor_policy.rb` and `super_admin/api_client_policy.rb` and also added tests
- Added hacks to ApplicationController and Api::V1::BaseApiController to get them to play nicely with namespaces policies
- Added override to the `current_user` method on Api::V1::BaseApiController so that it returns the client (User or ApiClient) because Pundit uses this by default

Hackathon Bugfixes
- Update to plan permissions to allow an ApiClient to retrieve publicly visible plans
- Fixed an issue in the plans_controller when the contributor identifier does not map to an IdentifierScheme (e.g. `{ type: "other", identifier: "12345" }`)
- Fixed contributor deserializer which referenced `roles` instead of `role`
- Added a nil check to the funding deserializer to allow empty `grant_id`
- Added a `distribution` to the `dataset` in the JSON output that contains the `download_url` which points to the plans pdf export url


briri:
closing this one as the pundit changes were getting too big. Will need to resubmit them in a separate ticket someday

PR #2565, opened 2020-06-17
Fix the stats migration to add filtered; re-commit schema (lots of ch…

xsrust:
After pulling in the development db changes, my environment was complaining about the migration missing it's version tag.  
Additionally re-committed schema.rb 

I've noticed that it's changing the ID from `integer` to `serial` which appears to be a difference in rails-version migrations:
https://stackoverflow.com/a/54694863

### DIFF
--- a/db/migrate/20200601121822_add_filtered_to_stats.rb
+++ b/db/migrate/20200601121822_add_filtered_to_stats.rb
@@ -1,4 +1,4 @@
-class AddFilteredToStats < ActiveRecord::Migration
+class AddFilteredToStats < ActiveRecord::Migration[4.2]
   def change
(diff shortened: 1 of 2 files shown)


briri:
just noticed the issues now in the `db:setup` in the tests. 

Maybe try doing a `rails db:schema:dump` (when on rails 5.2 and ruby 2.6.x) to regenerate the schema.rb file from scratch

xsrust:
Ok, I've started from a production-dump, migrated it with the old-app-code, and then done a db:schema:dump with the correct rails/ruby versions.

It looks like mysql and pg are using slightly different columns for the ID and foreign-key fields

PR #2566, opened 2020-06-18
fix issue with kaminari pagination

### LINKED ISSUE #2464: Fix Kaminari / Paginable pages
The [Kaminari](https://github.com/kaminari/kaminari) config is broken after upgrading.

The `Paginable` controller concern has been updated to use strong parameters. The ERB files are failing (`app/views/kaminari`) though when trying to access some of the helper functions relating to the fact that the `@options` variable in the `Paginator` helper class is not a hash.

This is possibly caused by the following change:
```
5.11 ActionController::Parameters No Longer Inherits from HashWithIndifferentAccess
Calling params in your application will now return an object instead of a hash. If your parameters are already permitted, then you will not need to make any changes. If you are using map and other methods that depend on being able to read the hash regardless of permitted? you will need to upgrade your application to first permit and then convert to a hash.
```

We should review the kaminari docs and our own implementation. Kaminari went from 1.1 to 1.2 so some of the config may have changed.

xsrust:
Fixes #2464 

Looks like the refactor of strong params requires us to call the helper method rather than an instance variable when passing in the pagination options

### DIFF
--- a/app/views/kaminari/_paginator.html.erb
+++ b/app/views/kaminari/_paginator.html.erb
@@ -10,16 +10,16 @@
   <nav class="pagination" style="margin: 0;">
     <%= first_page_tag unless current_page.first? %>
     <%= prev_page_tag unless current_page.first? %>
-    <%# each_page do |page| %>
-      <%# if page.display_tag? %>
-        <%#= page_tag page %>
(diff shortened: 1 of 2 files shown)


briri:
ha ha. good catch @xsrust

Issue #1974, opened 2018-10-30
Make the timeago.js strings translatable


briri:
The timeago.js file currently only exists in the `lib/assets/javascripts/locale/en_GB` directory and contains a lot of hard-coded strings that are not being handled properly for other locales.

There may be ways to handle it by adding a translated version to the other language directories in `lib/assets/javascripts/locale` and then updating `lib/assets/javascripts/utils/timeagoFactory.js` to grab the current locale and import the appropriate file.

A more convenient approach though might be to place those hardcoded strings into the `views/layouts/application.html.erb` JSON section that gets fed into `lib/assets/javascripts/constants.js` so that they're picked up and added to the app.pot file and handled like our other translations.

PR #2570, opened 2020-06-22
Rails5 ajax fixes

briri:
- Updated `application.js` to share Rails UJS with our custom JS and JS.ERB files
- Fixed up the pagination concern, js, controllers and views to use strong params and to render JSON when the user has interacted with the pagination, sorting or searching functionality

@raycarrick-ed note that in order for a scenario where we were programmatically calling `.submit` for a form, we now should use `Rails.fire(form[0], 'submit');` (where `form` is the JQuery element). I found this fix in this article: https://stackoverflow.com/questions/12683524/with-rails-ujs-how-to-submit-a-remote-form-from-a-function


briri:
@raycarrick-ed  @xsrust  lets hold off on this one. I'm going to play around some more with it to see if starting rails ups in the `packs/application.js` is behind our puma issues. These scenarios where we do a form submit through JS may be good candidates for switching over to the js.erb approach (for these specific situations)

briri:
ok @raycarrick-ed and @xsrust I think I may have stumbled upon the puma hanging issue. I noticed that it was trying to load the JSON file for the ResearchProjectsController which uses the Rails cache and the Thread library. When I commented out the 'Thread' bit things started puma stopped hanging.

That JSON load may be a good candidate for ActionCable's web-socket functionality. Something to investigate down the road I guess.

This PR should be good to go. Note that I enabled Turbolinks. I think now that we have a handle on the Ajax UJS issue that we can enable it again. 

I went through more of the pages and fixed up the ajax. Note the checkboxes on the Plans page and Users page which were previously checkboxes wrapped inside a form. This are now just 'remote' checkboxes and I had to add `authenticity_token: true` to get it to pass that token along to the controller.

briri:
ok some final commits on this one @raycarrick-ed and @xsrust. I fixed up all of the tests broken by adding the `optional: true` to model associations and also addressed #1974 (make timeago localized). Also made some minor tweaks to get the feature specs working (except for the templates/annotations ones which should start working once @raycarrick-ed updates for the JS on those pages is merged in)

app/controllers/application_controller.rb
```diff
@@ -45,6 +45,7 @@ def set_gettext_locale
 
   def current_locale
     session[:locale] || FastGettext.default_locale
```
briri, line 47: switch to ||=

app/controllers/orgs_controller.rb
```diff
@@ -4,6 +4,8 @@ class OrgsController < ApplicationController
 
   include OrgSelectable
 
+  skip_before_action :verify_authenticity_token, only: %w[search]
```
briri, line 7: switch to pass auth token

app/controllers/research_projects_controller.rb
```diff
@@ -31,7 +31,8 @@ def funder_type
 
   def fetch_projects
     Rails.cache.fetch(["research_projects", funder_type], expires_in: expiry) do
-      Thread.new { ExternalApis::OpenAireService.search(funder: funder_type) }.value
+      #Thread.new { ExternalApis::OpenAireService.search(funder: funder_type) }.value
```
briri, line 34: create ticket to investigate this

PR #2571, opened 2020-06-23
Rails5 ActiveRecord monkey patch for SchemaDumper

briri:
Fixes the issue with the SchemaDumper outputting non DB agnostic information.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -71,9 +71,6 @@ gem "pg", group: :pgsql, require: false
 # Bit fields for ActiveRecord (https://github.com/pboling/flag_shih_tzu)
 gem "flag_shih_tzu" #, "~> 0.3.23"
 
-# Required for the `import` statements at the top of the Stat generators
-gem "activerecord-import"
-
 # ======== #
 # SECURITY #
 # ======== #
--- a/Gemfile
+++ b/Gemfile
@@ -71,6 +71,9 @@ gem "pg", group: :pgsql, require: false
 # Bit fields for ActiveRecord (https://github.com/pboling/flag_shih_tzu)
 gem "flag_shih_tzu" #, "~> 0.3.23"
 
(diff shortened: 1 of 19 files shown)


xsrust:
The above commit replaces the `import` statements (which I can't find any ruby-documentation on) with the suggested circular-dependancy-fix from the parallelization library.

It seems that the ESlint is failing now preventing tests from running.
The AirBNB config is incompatible with eslint v`7.3.0`:
https://github.com/airbnb/javascript/issues/2245 
Suggested fix is to downgrade to `7.2.0` or upgrade to `7.3.1`

xsrust:
Pull in the latest from the Rails5 branch (ajax was causing some conflicts/issues on the PR but not the Push tests)

xsrust:
Success,
The tests now run for PR and Push.
there are some failures, but we can split that work off into other tickets after we've merged into Rails5

Gemfile
```diff
@@ -71,6 +71,9 @@ gem "pg", group: :pgsql, require: false
 # Bit fields for ActiveRecord (https://github.com/pboling/flag_shih_tzu)
 gem "flag_shih_tzu" #, "~> 0.3.23"
 
+# Required for the `import` statements at the top of the Stat generators
+gem "activerecord-import"
```
xsrust, line 75: I think this gem is for the bulk-import of data and generating minimal SQL statements rather than for the Ruby-import keyword.

I've reverted the import statements from the services, but the commit that added this references `imagemagick` so I wanted to double check before removing this

briri: ah yes. I thought I had reverted that change.

briri:
Ok thanks for updating this PR @xsrust. I have reverted out that gem and can see that the tests are failing on other issues. I will open a new ticket for that. It looks like the addition of `optional: true` to some of the associations on the models have invalidated our unit tests.

PR #2577, opened 2020-06-24
Rails 5/2472

xsrust:
First pass at refactoring callbacks, many still remain.

If the value being set did not rely on any other data within the model, the callback is replaced with the `attribute` to set a default value.

If the value being set does rely on the state of other data within the model, or if the callback serves to transform the data (i.e. `downcase`) then it is replaced with a custom accessor method.

There's a third case of callbacks that I suspect should remain as such, but I wanted to check with you all on that logic.
Here, data is modified on Object 1, which  caused other data on Object 2 to need to change,
> i.e. updating the plan.complete flag after saving an answer, or stripping permissions when a user changes organization. 

In those cases, the `after_save` or `after_update` callbacks are more applicable due to the potential to delete/change Object 2
 and then have validation fail on Object 1, invalidating the changes we've already made on `Object 2`


app/models/org.rb
```diff
@@ -185,6 +198,10 @@ def self.human_attribute_name(attr, options = {})
     HUMANIZED_ATTRIBUTES[attr.to_sym] || super
   end
 
+  # ===========================
+  # = Public instance methods =
+  # ===========================
+
   def links
     super() || { "org": [] }
   end
```
xsrust, line 207: Just noticed this, should we refactor to a default and remove this method?

app/models/org.rb
```diff
@@ -171,9 +178,15 @@ def self.default_orgs
...
-  before_validation :set_default_feedback_email_subject
+  # =============
+  # = Callbacks =
+  # =============
+
+  # this is basically checking the file-store for the dragonfly image each time
+  # before we validate in-case the db/system have moved and the link to the logo files
+  # is therefore different
+  # TODO: evaluate the need for this
   before_validation :check_for_missing_logo_file
```
xsrust, line 189: This one is a wierd-case.

The method effectively checks the validity of the image-file each time we save an org, and then does a seccondary search for the file if it cannot find the logo, finally nullifying the UID if it cannot find one.:
```
  # If the physical logo file is no longer on disk we do not want it to prevent the
  # model from saving. This typically happens when you copy the database to another
  # environment. The orgs.logo_uid stores the path to the physical logo file that is
  # stored in the Dragonfly data store (default is: public/system/dragonfly/[env]/)
  def check_for_missing_logo_file
    if self.logo_uid.present?
      data_store_path = Dragonfly.app.datastore.root_path

      if !File.exist?("#{data_store_path}#{self.logo_uid}")
        # Attempt to locate the file by name. If it exists update the uid
        logo = Dir.glob("#{data_store_path}/**/*#{self.logo_name}")
        if !logo.empty?
          self.logo_uid = logo.first.gsub(data_store_path, "")
        else
          # Otherwise the logo is missing so clear it to prevent save failures
          self.logo = nil
        end
      end
    end
  end
```

I don't think there's a way to do the same thing with either an accessor or default, so if we want to retain this logic, we'll need to keep the callback here.  We may find that this is unneccissary if we add some abstraction on ActiveStorage which would remove the reference to the file if it cannot find it.

app/models/org.rb
```diff
@@ -171,9 +178,15 @@ def self.default_orgs
...
-  before_validation :ensure_language
   after_create :create_guidance_group
```
xsrust, line 190: I wasn't sure how to remove the callback this case either.
After we make a new organisation, we want to create the default guidance_group for that org.

We could house that logic in the controller where we create organizations instead

app/models/org.rb
xsrust: We would also need a rake:task to fixup any existing data

app/models/plan.rb
```diff
@@ -585,18 +579,6 @@ def landing_page
...
-  # Returns nil
-  # Returns String
-  def set_creation_defaults
-    # Only run this before_validation because rails fires this before
-    # save/create
-    return if id?
-
-    self.title = "My plan (#{template.title})" if title.nil? && !template.nil?
-  end
-
```
xsrust, line 599: I'm not sure if there's a way to get this to work via an accessor as we would need a valid template which will not be available (unless specified as part of the initilizing call).
Additionally, we override this value in the `plans_controller#create` action
@briri do you know if that assumption holds for the new API, or would a default set in the accessor work for that case, as we'll have all the params at the time we make a call to `Plan.new()` ?

app/models/question.rb
```diff
@@ -83,9 +83,14 @@ class Question < ApplicationRecord
...
                                    message: UNIQUENESS_MESSAGE }
 
+  # =============
+  # = Callbacks =
+  # =============
 
+  # TODO: condition.remove_list needs to be serialized (from Array) before we can check
+  # for related conditions, so this can't be replaced by :destroy on the association
   before_destroy :check_remove_conditions
-
```
xsrust, line 88: Because we've used a serialized-array field for the list of questions a condition removes, (rather than a join table), we'll need to create an instance of the conditions which may refer to this question (any condition in the template) and iteratively remove this question from each:
```

  def check_remove_conditions
    id = self.id.to_s
    self.template.questions.each do |q|
      q.conditions.each do |cond|
        cond.remove_data.delete(id)
        if cond.remove_data.empty?
          cond.destroy if cond.remove_data.empty?
        else
          cond.save
        end
      end
    end
  end
```

I think we may want to refactor this to an `after_destroy` to ensure the delete has actually worked,
But I'm again not sure how to re-create this functionality without a callback

app/models/question_option.rb
```diff
@@ -49,6 +49,12 @@ class QuestionOption < ApplicationRecord
   validates :is_default, inclusion: { in: BOOLEAN_VALUES,
                                       message: INCLUSION_MESSAGE }
 
+  # =============
+  # = Callbacks =
+  # =============
+
+  # TODO: condition.option_list needs to be serialized (from Array) before we can check
+  # for related conditions, so this can't be replaced by :destroy on the association
   before_destroy :check_condition_options
```
xsrust, line 58: See the above comment on `app/models/question.rb` as this is effectively the same but on the `condition.option_list` attribute

app/models/template.rb
```diff
@@ -469,25 +482,6 @@ def new_family_id
...
-    self.published ||= false
-    self.archived ||= false
-    self.is_default ||= false
-    self.version ||= 0
-    unless id?
-      self.visibility = if (org.present? && org.funder_only?) || is_default?
-                          Template.visibilities[:publicly_visible]
-                        else
-                          Template.visibilities[:organisationally_visible]
-                        end
```
xsrust, line 483: I've removed this from the defaults for now as we set this in the `templates_controller#create` method, and it would not necessarily have access to the `org` attribute at creation time

app/models/template.rb
```diff
@@ -90,13 +101,14 @@ class Template < ApplicationRecord
...
   # = Callbacks =
   # =============
 
-  before_validation :set_defaults
-
+  # TODO: leaving this in for now, as this is better placed as an after_update than
+  # overwriting the accessors.  We want to ensure this template is published
+  # before we remove the published_version
+  # That being said, there's a potential race_condition where we have multiple-published-versions
   after_update :reconcile_published, if: -> (template) { template.published? }
```
xsrust, line 112: Just wanted to make sure the logic of the above comment makes sense, and check if there's a better approach

app/models/user.rb
```diff
@@ -162,8 +165,8 @@ class User < ApplicationRecord
   # = Callbacks =
   # =============
 
-  before_validation :ensure_language
-
+  # TODO: should the following-three actions be replaced by overriding the
+  # org= method?
   before_update :clear_department_id, if: :org_id_changed?
 
   after_update :delete_perms!, if: :org_id_changed?, unless: :can_change_org?
```
xsrust, line 172: The third action is:
`after_update :remove_token!, if: :org_id_changed?, unless: :can_change_org?` 

We could remove the perms and token when the values are changed, rather than after the save completes, but I think they may make more sense as callbacks

app/models/language.rb
```diff
@@ -46,11 +46,17 @@ class Language < ApplicationRecord
...
-  # = Callbacks =
-  # =============
-
-  before_validation :format_abbreviation, if: :abbreviation_changed?
+  # =========================
+  # = Custom Accessor Logic =
+  # =========================
+
+  # ensure abbreviation is downcase and conforms to I18n locales
+  # TODO: evaluate the need for the LocaleService after move to Translation.io
```
briri, line 54: 🙏 we're having so many problems with Fastgettext right now that I'm wondering if its worth moving now

app/models/org.rb
briri: We can ditch this code. If I remember correctly it was an attempt to try and find the logo (back when logos were stored on local disk) when copying the DB from one machine to another

app/models/org.rb
briri: I think it makes sense to move that outside the model. If we do move this to the controller we will likely need to do the same in the `/app/services/org_selection/hash_to_org.rb`

app/models/org.rb
briri: yes and yes :)

app/models/plan.rb
briri: The `title` is required in the new api and we reject the request if its nil or empty

briri: I think the Create Plan page requires the title as well. Not sure what the scenario is with copying

app/models/question.rb
briri: the placement of this code feels wrong ... an object manipulating it's peers before it's destroyed :/ ... maybe better if it lived in a service that the controller interacts with.

anyway, I think just leave as a callback for now

app/models/template.rb
briri: makes sense to leave this one as is

app/models/user.rb
briri: agreed Sam. We could make the change on `def org_id=` but I'm not sure whether or not that would trigger properly if we had `user.org = org` somewhere.

app/models/question.rb
raycarrick-ed: I think the logic is correct but having to do this seems wrong. Seems to me that the "correct" thing to do here would be to use a join table. I stayed with the serialisation as closer to the original Killian implementation but it seems weird. Not sure it would be a huge task to rewrite it. Do now? Do later? Happy to go for it if consensus is do it now or leave till rails5 is done....

briri: agreed @raycarrick-ed. I think we should hold off on it until after the Rails 5 migration is complete

app/models/org.rb
xsrust: The `hash_to_org` service actually dosen't persist the `Org` object, but rather populates a `Org.new`, and then the persistance is done in the `OrgSelectable`.  

It got me thinking about the purpose of creating the default-guidance-group.  It's unpublished, so really just serves as a way to ensure theres one `GuidanceGroup` with the org's abbreviation and `optional_subset: false` for an org-admin to begin attaching guidance to.

I moved the logic for the method over to the `GuidanceGroup` model, and added a method to the `GuidancesController` to ensure that this default group gets created for any org which is managed and does not currently have a default org

app/models/plan.rb
xsrust: Looks like the copying scenario is to prepend the title with `"Copy of " ` so this is fine to remove as it's always set at plan-creation

PR #2578, opened 2020-06-25
Bug missing indexes

briri:
Adds a few missing indexes to the plans table for the new Org, Funder and Grant Id relationships. Discovered this during the DMPTool deploy of v2.2.0


briri:
ah yes. will fix

PR #2609, opened 2020-07-10
Rails 5 - Fix Errors stopping tests from running

xsrust:
Fixes an issue with the database failing to seed due to validations


xsrust:
Started this from the Rails5 branch rather than development,

xsrust:
Additionally fix a secondary error with the AirBnB Linting which was stopping the assets from compiling

PR #2617, opened 2020-07-17
Fix some remaining test failures

xsrust:
So far the issues with the `spec/services/template/upgrade_customization_service_spec` have been resolved.  

My local environment does not run all the feature tests, so I may need to use this PR as a runner service

### DIFF
--- a/app/mailers/user_mailer.rb
+++ b/app/mailers/user_mailer.rb
@@ -71,10 +71,9 @@ class UserMailer < ActionMailer::Base
   end
 
   def feedback_notification(recipient, plan, requestor)
-    return unless @user.org.present? && recipient.active?
+    return unless recipient.active?
 
     @user = requestor
-    @org = @user.org
     @plan = plan
     @recipient = recipient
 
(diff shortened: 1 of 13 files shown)


xsrust:
After getting the upgrade.rake syntax fixed up, it looks like the rubocop-ing has broken a large portion of the tests:
```
Finished in 19.81 seconds (files took 4.38 seconds to load)
1387 examples, 733 failures, 2 pending
```

app/mailers/user_mailer.rb
```diff
@@ -71,10 +71,9 @@ def plan_access_removed(user, plan, current_user)
   end
 
   def feedback_notification(recipient, plan, requestor)
-    return unless @user.org.present? && recipient.active?
+    return unless recipient.active?
 
     @user = requestor
-    @org = @user.org
```
xsrust, line 77: org was not used in the mail content

app/services/template/upgrade_customization_service.rb
```diff
@@ -76,19 +76,16 @@ def call
...
           # customization_phase
-          #
-          # a) If the Org's template ({#customized_template}) has the Phase...
-          if customized_phase == find_matching_record_in_collection(
+          customized_phase = find_matching_record_in_collection(
             record: funder_phase,
             collection: customized_template.phases
           )
+          # a) If the Org's template ({#customized_template}) has the Phase...
+          next unless customized_phase.present?
```
xsrust, line 84: Replaced the conditional with a guard-statement

PR #2620, opened 2020-07-23
fix up strong params etc for plan creation

raycarrick-ed:
Fixes # .

Changes proposed in this PR:

Create plan wasn't working. Not sure I haven't gone down a rabbit hole with this, so needs checking.

The strong params seemed to be a problem. The org params where at the same level as the plan attributes and so the plan_params stripped them out. As far as I can tell:
```
params.require(X).permit(Y,Z)
```
means that params need to be of the form
```
{X => {Y => foo, Z => bar}}
```
rather than
```
{
X => {Y => foo}
Z => bar
}
```
which is what was there.

So I changed the `plan` model to accept attributes for `org` and altered the org form to use `f.fields_for` rather than plain `fields_for` to embed the org attributes within the plan which gave params that looked like
```
{
plan => {
    ....
    org_attributes => {...}
}
```
which can then pass strong params.
I then reworked the other bits as needed to get it to work.
There was also an issue with the JS for plan creation where it was looking for `orgId.length` but `orgId` was undefined so I put in a check for that too.

### DIFF
--- a/app/controllers/orgs_controller.rb
+++ b/app/controllers/orgs_controller.rb
@@ -191,7 +191,7 @@ class OrgsController < ApplicationController
   end
 
   def search_params
-    params.require(:org).permit(:name, :type)
+    params.require(:plan).permit(org_attributes: [ :name, :type ])
   end
 
   def shib_login_url
--- a/app/controllers/orgs_controller.rb
+++ b/app/controllers/orgs_controller.rb
@@ -191,7 +191,7 @@ class OrgsController < ApplicationController
   end
 
   def search_params
-    params.require(:plan).permit(org_attributes: [ :name, :type ])
+    params.require(:plan).permit(org_attributes: %i[name type ])
   end
 
   def shib_login_url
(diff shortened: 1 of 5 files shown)


briri:
Hmm. The change fixes the create plan page but breaks the Org search on other pages like Create account and edit profile.

I had tried to originally make that controller action generic but maybe its best to refactor them.

briri:
The issue was indeed with the strong param changes, but also to do with the complexity of the new Org Selection / Typeahead code. The OrgsController#search is called by the AJAX Typeahead when the user is entering an Org name into the box. The PlansController#create was having an issue because the :org and :funder params were being passed outside of the :plan namespace.

I submitted a separate PR #2621 that addresses the create plan issue and adds a ton of commentary to the code to hopefully help clarify what's happening (or should be happening)

raycarrick-ed:
Not needed now - sorted by https://github.com/DMPRoadmap/roadmap/tree/fix-org-selectors

PR #2621, opened 2020-07-23
fixed issue with create plan

briri:
@raycarrick-ed this one fixes the issue with the Create Plan page. Thanks again for catching it. 

It wasn't actually an issue with the OrgsController#search action (which is used by the Typeahead AJAX logic). 

I also added a ton of comments to the OrgSelectable concern and the OrgsController#search action to hopefully help clarify what's going on. I have a feeling that it will be helpful in the future if we need to work on any of that code.

### DIFF
--- a/app/controllers/concerns/paginable.rb
+++ b/app/controllers/concerns/paginable.rb
@@ -186,13 +186,11 @@ module Paginable
     @args.except(*PAGINATION_QUERY_PARAMS).to_param
   end
 
-  # TODO: We would be better served here passing these arguments as a Ruby double splat
-  # rubocop:disable Lint/UnusedMethodArgument
   def stringify_query_params(page: 1, search: @args[:search],
                              sort_field: @args[:sort_field],
                              sort_direction: nil)
 
-    query_string = {}
+    query_string = { page: page }
     query_string["search"] = search if search.present?
     if sort_field.present?
       query_string["sort_field"] = sort_field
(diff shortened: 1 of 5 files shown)


app/controllers/concerns/paginable.rb
```diff
@@ -186,21 +186,18 @@ def stringify_nonpagination_query_params
...
   end
 
-  # TODO: We would be better served here passing these arguments as a Ruby double splat
-  # rubocop:disable Lint/UnusedMethodArgument
   def stringify_query_params(page: 1, search: @args[:search],
                              sort_field: @args[:sort_field],
                              sort_direction: nil)
 
-    query_string = {}
+    query_string = { page: page }
```
briri, line 193: this was a separate issue that I stumbled across while testing

app/views/plans/new.html.erb
```diff
@@ -50,7 +50,7 @@
         <div class="form-group col-md-6">
            <em class="sr-only"><%= research_org_tooltip %></em>
            <% dflt = @orgs.include?(current_user.org) ? current_user.org : nil %>
-           <%= fields_for :org, @plan.org do |org_fields| %>
+           <%= f.fields_for :org, @plan.org do |org_fields| %>
```
briri, line 53: I stupidly missed this when doing the strong params. `f.` ensures that the field id is `plan_org_org_name` instead of `org_org_name` 🙄

PR #2622, opened 2020-07-28
use new credentials.enc file and pull out constants

### LINKED ISSUE #2467: Move all constants out of the code and into the new initializer
**Constants:**
Remove any hard coded values into the new `config/initializers/dmproadmap.rb` file.

Use the `config.[category].[variable] = "[value]"` syntax. Be sure to add a comment explaining what the variable is used for.

Then update the place where the variable was removed from with `Rails.configuration.x.[category].[variable]`

If the variable is used often, consider adding a helper method to the `ApplicationService`. For example `Rails.configuration.x.application.name` is used all over the codebase so it was simplified to `ApplicationService.application_name`

**Credentials:**
Go through any config and config/initializer files we have that are samples and move their sensitive values into the new `config/credentials.yml.enc` file.

Also look into switching references to these values over to ENV variables so that installations can either use the `config/credentials.yml.enc` or environment variables.

Here are some articles to read before starting that work. I would suggest we err towards the Rails6 approach:
https://medium.com/cedarcode/rails-5-2-credentials-9b3324851336
https://blog.capsens.eu/rails-5-2-imperfect-credentials-4ef03934aea4
https://blog.saeloun.com/2019/10/10/rails-6-adds-support-for-multi-environment-credentials.html

raycarrick-ed:
Fixes #2467  .

Changes proposed in this PR:
- introduce use of config/credentials.yml.enc or storing values needing hidden
- adds a credentials.yml.example file which is a template for the file
- when then file is edited using rails credentials:edit it will create the encrypted file and the master.key file which can then be set up in your environment as RAILS_MASTER_KEY
- once we move to rails 6 it will be easier to differentiate between environments but for now they are all one.
- devise.rb and recaptcha.rb no longer contain confidential information so can be safely checked in.
- also pulls out a couple of other constants.


nicolasfranck:
I'm just wondering: rails 5.2 expects us to generate a `config/credentials.yml.enc`, and it put it in our git repository.

But what if you install the same software on different servers (development-server, production-server),
and each need to have their own credentials (e.g. different captcha account)?

Automatisation software (like puppet) cannot be expected to have a full ruby-on-rails installation
in order to generate a simple configuration file..

briri:
That's a great question @nicolasfranck and there's not a standard way to manage this since everyone's environment and deployment tools will differ.

This repo has the Dotenv gem installed and that is what I am using combined with some modifications to the [config/database.yml](https://github.com/CDLUC3/dmptool/blob/main/config/database.yml) file. I then use Capistrano to manage which `.env`, `config/master.key` and `config/credentials.yml.enc` files are deployed out to a particular environment. 

Whichever approach you take, keep in mind that Rails 6+ will have its own credentials and keys for each env (e.g. config/credentials/production.key).

PR #2623, opened 2020-07-30
fixed bug when contributor.name is blank for api submission

briri:
Fixes an issue with the new v1 API when contributor info has no name defined. The minimum contributor content is the email/mbox.

Will submit a corresponding PR to master as well


briri:
going to close this one out since the fix is already in master and will be addressed when we resync dev

Issue #2616, opened 2020-07-16
Adding 'last opened' plans to either usage statistics or to the table for plan review 


magdalenadrafiova:
In our DMPonline drop-in session there was a suggestion that with upgrading the plan review table #2365 it will be nice to also be able to see when was a plan last opened/accessed by an admin. 

It would be interesting to collect the data on this and Sam has suggested that this could feed more interesting statistics to the usage dashboard too.

briri:
@mariapraetzellis and I discovered that the upgrade to the latest version caused many of the plan records' `updated_at` date to be the date of the upgrade 👎 

This also got us thinking about the fact that if I go in and add a contributor to my plan or answer a question that the `updated_at` date on the `plans` table does not change and so the tables that display plans 'update' column would technically be inaccurate from the user's point of view

I think we should also consider adding a `changed_date` to the table (and possibly others) to track when a plan has changed.

xsrust:
I've hit that issue with rake:tasks before and found that `object.save(touch: false)` will prevent the timestamp changes (not sure if it's worth going back and changing that task though)

briri:
Yeah, we should probably update those Rake tasks to prevent others from encountering this issue

PR #2625, opened 2020-08-03
Fix upgrade tasks

briri:
Addresses issue with latest upgrade rake tasks referenced in #2616 .

Changes proposed in this PR:
- add `touch: false` to several of the tasks that update plans records so that the updated_at dates are not all set to the date of the upgrade


xsrust:
@briri I think this bugfix is to the old rake:tasks, so should likely go into Master first and then development?

briri:
closing this one will resubmit against master

PR #2626, opened 2020-08-03
Add org_id to api_clients

briri:
Adding a link between ApiClients and Orgs In preparation for some of the upcoming maDMP changes.
- Add org_id to api_clients table
- Add association to ApiClient model
- Update API v1 policies to allow ApiClient to access the Org's plans
- Updated SuperAdmin -> Api Clients pages to allow changes to Org
- Updated tests as needed


app/models/api_client.rb
```diff
@@ -29,6 +33,9 @@ class ApiClient < ActiveRecord::Base
   # = Associations =
   # ================
 
+  # TODO: Enable `optional: true` when merged into Rails 5 codebase
+  belongs_to :org # , optional: true
```
xsrust, line 37: :+1:

app/views/super_admin/api_clients/_form.html.erb
```diff
@@ -1,6 +1,7 @@
 <%
 url = @api_client.new_record? ? super_admin_api_clients_path : super_admin_api_client_path(@api_client)
 meth = @api_client.new_record? ? :post : :put
+orgs = Org.where(is_other: false)
```
xsrust, line 4: should this refer to the new `managed` flag rather than `is_other`?

app/controllers/api/v1/plans_controller.rb
```diff
@@ -12,30 +12,11 @@ class PlansController < BaseApiController
...
-          end
-
-          if plans.present? && plans.any?
-            @items = paginate_response(results: plans)
-            render "/api/v1/plans/index", status: :ok
-          else
-            render_error(errors: [_("Plan not found")], status: :not_found)
-          end
+        plans = Api::V1::PlansPolicy::Scope.new(client, Plan).resolve
+                                           .where(id: params[:id]).limit(1)
```
xsrust, line 16: It would be nice to re-use this pattern of using the policy-scopes to determine the viewership-rights across the app

xsrust:
Looks like the postgres/mySQL tests are failing because of pending migrations.
I think this is due to the `version` in the `schema.rb` file not having updated.

xsrust:
@briri It looks like a bunch of the development code leaked in with that Force-push yesterday.  Is it possible to pull out just the change on Master?

briri:
rebased off of dev by accident. Will resubmit to master

PR #2652, opened 2020-08-25
issue#2632: default prefs not being picked up. Hash structure changed.

### LINKED ISSUE #2632: Edit profile - notification preferences don't display
**Expected behaviour:**
Notification preferences offer me a list of tick boxes to select email notification preferences from. 

**Actual behaviour:**
Nothing is displayed there, but it shows an error " Error: No default preferences found (should be in dmproadmap.rb initializer)."
Edit profile notificstion issue

**Steps to reproduce:**
Go to "edit profile" and click on the notification preferences tab. I'm using Chrome as browser. The other tabs in that section are functional.

raycarrick-ed:
Fixes #2632 .

Changes proposed in this PR:
- prefs hash structure had changed slightly

### DIFF
--- a/app/controllers/registrations_controller.rb
+++ b/app/controllers/registrations_controller.rb
@@ -6,7 +6,7 @@ class RegistrationsController < Devise::RegistrationsController
 
   def edit
     @user = current_user
-    @prefs = @user.get_preferences(:email)[:prefs]
+    @prefs = @user.get_preferences(:email)
     @languages = Language.sorted_by_abbreviation
     @orgs = Org.order("name")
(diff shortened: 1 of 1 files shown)


xsrust (approved):

xsrust:
This should be ok to merge without the Brakeman passing, That's usually limited to PRs against Master.

We should re-enable and check why those cases are failing before bundling up a release.

xsrust:
Found another instance of `[:pref]` so I've updated the PR.
Looks like Brakeman's no longer running, so you can ignore the first half of the above

PR #2657, opened 2020-08-26
Add touch false to upgrade task that migrates Org/Funder/Grant ids on Plans

briri:
Prevents the `updated_at` from changing on `plans` records when migrating existing data to the new identifiers table


raycarrick-ed (approved):
looks fine

PR #2671, opened 2020-09-08
fix for plan deletion

### LINKED ISSUE #2643: Deleting old plan causes validation error

**Expected behaviour:**
Should be able to remove plan on dashboad

**Actual behaviour:**
if it's an old plan which doesn't have an org associated then it fails validation

**Steps to reproduce:**
On system with plan created prior to orgs being associated with plans, delete plan from dashboard.

briri:
Fixes #2643 .

Changes proposed in this PR:
- set the Plan's org_id to that of the Role record's user if it is not set

Ideally the `upgrade:v2_2_0_part1` upgrade task would have been run and none of the Plan records would have a nil org_id. This addresses the issue though for scenarios where the plans.org_id is nil for some reason.

### DIFF
--- a/app/models/role.rb
+++ b/app/models/role.rb
@@ -103,6 +103,10 @@ class Role < ApplicationRecord
   def deactivate!
     self.active = false
     if save!
+      # Set the org_id on the Plan before calling deactivate. The org_id should
+      # not be blank. This catches the scenario where the `upgrade:v2_2_0_part1`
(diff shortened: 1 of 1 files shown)


xsrust:
Thanks @briri   

We refreshed the data on the test server before deploying out the Rails5 work, but I don't think we ever ran through the migration tasks when upgrading that.

xsrust (approved):

PR #2683, opened 2020-09-15
Implimenting Translation.io

xsrust:
Starting a fresh PR here because I've been wrangling with Github actions with the old one.

Changes made:
replace `set_gettext_locale` -> `set_locale`
replace `FastGetText` -> `I18n`
add `en-CA` and `fr-CA` locales

Fix issues with seeding the database caused by new locale files and checking tables in initializer

Sync with translation.io for latest translations `rails translation:sync_readonly DOMAIN=app`


xsrust:
Just adding a note here that I've reverted the webdriver to the older version we used for Travis compatibility as I cannot get this new capybara headless one to not run sandboxed and thus it keeps failing locally.

@briri If you had a good reason to swap (you'd made the change with the maDMP work), just let me know and I'll make this configurable w/ a `Docker` env-variable to swap between the two

briri:
ok. I think it's fine to revert to the older version for now.

briri (approved):
looks good @xsrust. Thanks for working through the issues with the tests!

Issue #2638, opened 2020-08-17
Unable to create a plan w/ no research org and no funder boxes checked


mariapraetzellis:
Trying to create a new plan with the boxes ticked for both "No research organisation associated with this plan or my research organisation is not listed" and "No funder associated with this plan or my funder is not listed" this should give a plan based on the default template but gives the following error: 
"Error: Unable to create the plan.
Org translation missing: en-GB.activerecord.errors.models.plan.attributes.org.required"

mariapraetzellis:
I am now able to create a plan w/ no research org and no funder boxes checked.

However, selecting no research org but including a funder does not work. The Create Plan button is grayed out. I compared this with the Rails 4 version on https://dmptool-stg.cdlib.org/plans/new and I was able to do this in the prior version. 

![screenshot-dmponline-test dcc ac uk-2020 09 04-11_59_47](https://user-images.githubusercontent.com/27511575/92276607-2de22a80-eea6-11ea-8630-bc0262157d43.png)

briri:
@raycarrick-ed and @xsrust this may be due to a data issue. Make sure there is a default template defined and that it is 'published'

xsrust:
Does not look to be a data-issue.
with neither option selected, this works fine and returns the default template.

With only a funder specified, it's failing on `undefined method flatten for Template::ActiveRecord_Relation` on line 55.

mariapraetzellis:
Working now 👍

PR #2690, opened 2020-09-21
fix bug preventing user from selecting a non-customized funder-template

xsrust:
ensures the @templates variable is always cast to an array so the later call to `flatten` works

Addresses #2638


briri (approved):
thanks @xsrust

PR #2696, opened 2020-10-02
Fix for minor bugs with Rails5

briri:
- Found a bug saving an Org when  the shibboleth entityId field is enabled and visible on the form 
- Fixed issue with form id for the `togglizePasswords` call on the create account form's JS
- Fixes issue when translation file does not have a translation defined for an empty string. The admin - users page was displaying a nasty error if the user was not an admin. Removed the `_("")` entirely.

See https://github.com/CDLUC3/dmptool/issues/214 for details and screenshots

### DIFF
--- a/app/controllers/orgs_controller.rb
+++ b/app/controllers/orgs_controller.rb
@@ -47,8 +47,9 @@ class OrgsController < ApplicationController
       if Rails.configuration.x.shibboleth.use_filtered_discovery_service
         shib = IdentifierScheme.by_name("shibboleth").first
 
-        if shib.present? && attrs.fetch(:identifiers_attributes, []).any?
-          entity_id = attrs[:identifiers_attributes].first[:value]
+        if shib.present? && attrs[:identifiers_attributes].present?
+          key = attrs[:identifiers_attributes].keys.first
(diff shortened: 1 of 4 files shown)


app/controllers/orgs_controller.rb
```diff
@@ -47,8 +47,9 @@ def admin_update
       if Rails.configuration.x.shibboleth.use_filtered_discovery_service
         shib = IdentifierScheme.by_name("shibboleth").first
 
-        if shib.present? && attrs.fetch(:identifiers_attributes, []).any?
-          entity_id = attrs[:identifiers_attributes].first[:value]
+        if shib.present? && attrs[:identifiers_attributes].present?
+          key = attrs[:identifiers_attributes].keys.first
```
xsrust, line 51: The tests seem to be failing here:

```
     NoMethodError:
       undefined method `keys' for #<Array:0x000055d0585d22e8>
     # ./app/controllers/orgs_controller.rb:51:in `admin_update'
```

xsrust (approved):
These look good to me bar the array/hash issue the tests are throwing up

briri:
Updated the test to not pass those identifier attributes as an array so that they mimic what the controller is actually receiving from the form

PR #2700, opened 2020-10-05
update user merge function to work with new identifier model

### LINKED ISSUE #2693: Merging users leads to 404

**Expected behaviour:**
When merging 2 users, I would expect to see a successful user profile page that has the DMPs from the merged account added to it 

**Actual behaviour:**
After confirming the merge, I end up with a "page unavailable" message

**Steps to reproduce:**
I tried to merge https://dmponline-test.dcc.ac.uk/super_admin/users/40384/merge with the hotmail address account for Rosie.

**Update**
I do think the plan was moved across - the user now has 8 plans and the account that was merged into it doesn't have a record associated any longer.

xsrust:
Fixes #2693

looks like this reference was missed when refactoring the identifier model.

My fault for not adding tests to the merge workflow before:P

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -437,7 +437,7 @@ class User < ApplicationRecord
     # => auths -> map onto keep id only if keep does not have the identifier
     to_be_merged.identifiers
                 .where.not(identifier_scheme_id: scheme_ids)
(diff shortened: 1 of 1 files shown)


briri (approved):
good catch @xsrust

PR #2702, opened 2020-10-06
Rails5 fixes

briri:
- Tweak to a class on the Edit Profile page
- Fixed typo on the contributors page 'Principle' -> 'Principal'
- New `v3.rake` file (the `upgrade.rake` file is getting really long) that contains:
  - Task to set nil `orgs.language_id` and `users.language_id` to the default language (creates an 'en-GB' one if not defined)
  - Task to set the nil `orgs.feedback_email_subject` and `orgs.feedback_email_msg` to the ones defined in the FeedbacksHelper.rb


xsrust (approved):
Looks great,
Thanks for making a new upgrade rake:task file.
It's probably a good convention for us to atleast do that every time we release a major version.

PR #2706, opened 2020-10-12
update translation strings in codebase, re-sync translated strings fr…

xsrust:
…om Translation.io


briri (approved):
Thanks for remembering to do this Sam!

PR #2716, opened 2020-11-03
Patch for v3 - Fix bug preventing some identifiers from updating & fix bad fundref prefix

briri:
- Bug fix that was preventing ORCIDs, RORs and Fundrefs from being updated
- Update code to use/reference the correct Crossref funder id prefix: from 'https://api.crossref.org/funders/' to 'https://doi.org/10.13039/'
- Added v3 migration script that will update the identifier_scheme and fix any existing identifiers in the DB


xsrust (approved):

briri:
closing this one out in favor of the new one that includes some fixes for the credit taxonomy URLs

PR #2717, opened 2020-11-05
Bug fix for updating identifiers, update fundref DOI base URL and Credit taxonomy base URL

briri:
- Bug fix that was preventing ORCIDs, RORs and Fundrefs from being updated
- Update code to use/reference the correct Crossref funder id prefix: from 'https://api.crossref.org/funders/' to 'https://doi.org/10.13039/'
- Added v3 migration script that will update the identifier_scheme and fix any existing identifiers in the DB
- Discovered today that the Credit URLs we were using are no longer valid. This updates the code to work with the new URL structure.
- Remove ROR link from the affiliation column on the contributors table


briri:
@xsrust this is a combination of #2716 which you reviewed previously and an update to the URLs for the Credit taxonomy we use for contributor roles

xsrust (approved):

PR #2725, opened 2020-11-13
Add org_id to api_clients table and refactor of API v1 services

briri:
- Added org_id as an optional FKey on the api_clients table
- Updated UI to allow Org selection on 'Api Clients' screens
- Added a scope to the API's plan controller policy 
- Refactored the API deserialization services:
  - Extracted validation logic to a separate service that can be called before actually processing the incoming JSON. It checks for the metadata we require specifically and will return specific issues to the caller
  - Extracted several common functions from the old deserialization services into a shared DeserializationService class
  - Added a ContextualErrorService that passes through the a Plan and all of its dependencies checking for errors. The resulting errors provide essential context for the caller (e.g. instead of getting 'name can't be blank' the caller would receive 'Funder name can't be blank')
  - API deserialization services no longer create/save records
  - Added a PersistenceService that allows for the 'safe' saving of the Plan and all of its dependencies. For example, calling `plan.save` would result in errors if it had 2 contributors belonging to the same org. The first org would get saved but the subsequent attempt would result in a duplicate record error. The methods in this class essentially just do a `find_or_create`. The entire process is wrapped inside a transaction so that if any failures are encountered the entire transaction is aborted. 
  - Added logic to the PersistenceService that consolidates contributor/contact info so that `{ contact: { email: 'john.doe@uni.edu' }` and `{ contributor: { email: 'john.doe@uni.edu', name: 'John Doe' }` are combined into one single entry
  - Added tests for everything

I am considering adding a rake task that will take a json input and run it through these services to validate/provide feedback so that we can test the API more easily and also assist with debugging.


app/services/api/v1/contextual_error_service.rb
```diff
@@ -0,0 +1,66 @@
...
+          errs.map do |msg|
+            msg.gsub("org name", "affiliation name")
+               .gsub("identifiers value", "identifier values")
+               .gsub("Contributors", "Contact/Contributor")
+               .gsub(/^Identifiers/, "#{context.capitalize} identifier")
+               .gsub(/^Name/, "#{context.capitalize} name")
+               .gsub(/^Title/, "#{context.capitalize} title")
+               .gsub(/^Value/, "#{context.capitalize} value")
+          end
+        end
```
xsrust, line 51: Cool approach to making the ActiveModel errors make sense to the API users without needing to re-invent the wheel:)

xsrust (approved):
Changes look good.

Thanks for the big refactor and for updating all the testing

Issue #2724, opened 2020-11-13
Refactor method Org#plans?


nicolasfranck:
Please complete the following fields as applicable:

**Expected behaviour:**

The method `Org#plans` should only return plans of the current Org.

**Actual behaviour:**

It also returns plans of other orgs.

Besides, it generates very large queries, that are reexecuted every time someone uses `org.plans`
in his code. If I remove this method, and so use the more simple (and cached) relation `plans`,
the global page loading time goes from 4s to 40ms.

Reason: The relation `plans` in the model `Org` is defined,
but overridden a few lines further (https://github.com/DMPRoadmap/roadmap/blob/master/app/models/org.rb#L287). 
But that last method is slow and does not work as expected.

How it works now:

* Select all roles with `active`:`true`, access in set of `administrative`, and `user_id` in list of `org.users.pluck(:id)`
* Pluck attribute `plan_id`, make it unique, and store it in `plan_ids`
* Fetch all plans with id in list of `plan_ids`

Why it does not work: some users can have administrative rights to plans that actually belong to another organisation

**Steps to reproduce:**

briri:
Hi @nicolasfranck this behavior is desired as it allows for scenarios where a DMP is a collaboration between multiple organizations. For example if a user from Org A creates a DMP and then invites fellow collaborators from Org B. We have many areas in the code that rely on this.

We do however agree that overriding the default association `org.plans` may not have been the best choice as it can be misleading and return unexpected results. Perhaps a different name for the method like `related_plans` would have been better. 

We also agree that the query is inefficient and could use improvement.

nicolasfranck:
That explains why it is just a list of plans without adding links to the "show" or "edit" page.

So instead of being a list of plans belong to this current organisation, it is a list of plans
that members of this organisation have administrative rights to. Is it important for org-admins
to know that certain members have "private" access to plans outside of the organisation
(without being able to look more closely)?

briri:
Hi @nicolasfranck a cursory search for that method shows that its used on the statistics, policies (used in conjunction with the Devise and Pundit gems for authorization), the API, and several areas of page navigation. 
```shell
> grep -r 'rg\.plans' app 
app/models/plan.rb:    plan_ids = user.org.plans.where(complete: true).pluck(:id).uniq
app/policies/api/v1/plans_policy.rb:            ids += client.org.plans.organisationally_visible.pluck(:id)
[...]

PR #2726, opened 2020-11-16
use relation org.plans instead of method, which works faster and more correct

nicolasfranck:
Fixes issue #2724 .


nicolasfranck:
mm, there seems to be something wrong with the integration tests for postgresql and mysql..

briri:
it looks like there was a change to the way GitHub actions wants us to dynamically set env variables. I have submitted a fix for this in #2728. The dev team will be reviewing these PRs tomorrow.

briri:
Hi @nicolasfranck and thanks for sending these PRs over! We are a small team so your contributions and suggestions are appreciated.

The `org.plans` method is used fairly extensively across the site and so removing the above will have many unintended consequences.

A quick search shows it is used for statistics, security policies, UI navigation logic, and the API:
```shell
> grep -r 'rg\.plans' app 
app/models/plan.rb:    plan_ids = user.org.plans.where(complete: true).pluck(:id).uniq
app/policies/api/v1/plans_policy.rb:            ids += client.org.plans.organisationally_visible.pluck(:id)
app/policies/api/v1/plans_policy.rb:            ids += client.org.plans.pluck(:id) if client.can_org_admin?
app/policies/plan_policy.rb:       @user.org.plans.include?(@plan))
app/controllers/paginable/plans_controller.rb:    plans = @super_admin ? Plan.all : current_user.org.plans
app/controllers/api/v0/statistics_controller.rb:               @user.org.plans.where(complete: true)
app/controllers/api/v0/statistics_controller.rb:               @user.org.plans
app/controllers/api/v0/statistics_controller.rb:    scoped = @user.org.plans
app/controllers/api/v0/statistics_controller.rb:    @org_plans = @user.org.plans
app/controllers/api/v0/plans_controller.rb:    @plans = @user.org.plans.includes([{ roles: :user }, { answers: :question_options },
app/controllers/org_admin/plans_controller.rb:    @plans = @super_admin ? Plan.all.page(1) : current_user.org.plans.page(1)
app/controllers/org_admin/plans_controller.rb:      org.plans.includes(template: :org).order(updated_at: :desc).each do |plan|
app/views/plans/_navigation.html.erb:  <% if plan.administerable_by?(current_user) || (current_user.can_org_admin? && current_user.org.plans.include?(plan)) %>
app/views/plans/_navigation.html.erb:  <% if (plan.administerable_by?(current_user) || (current_user.can_org_admin? && current_user.org.plans.include?(plan))) && plan.owner_and_coowners.include?(current_user) && plan.owner.org.feedback_enabled? %>
```

We agree that `Org.plans` should default to the standard ActiveModel relationship functionality and just return any Plans where `org_id = ?`. We would still need to retain the other though and determine which one to use in each scenario. It will take some work for us to do the proper analysis.

xsrust:
Updating this PR as well, see https://github.com/DMPRoadmap/roadmap/pull/2741#issuecomment-732804820 for details

nicolasfranck:
Ok, now it works like before, but now `org.plans` is cached so the crazy query is not reexecuted every time someone uses that statement.

xsrust (approved):
Thanks @nicolasfranck 
I think the cache approach should alleviate some of the performance issues in the meantime, but longer-term we should definitely consider making this a custom-relationship and renaming to something like `org.plans_where_administrator` (or something more concise) and then defining `org.plans` through the association.

app/models/org.rb
```diff
@@ -285,11 +285,13 @@ def org_admins
   end
 
   def plans
-    plan_ids = Role.administrator
-                   .where(user_id: users.pluck(:id), active: true)
-                   .pluck(:plan_id).uniq
-    Plan.includes(:template, :phases, :roles, :users)
-        .where(id: plan_ids)
+    Rails.cache.fetch("org[#{id}].plans", expires_in: 2.seconds) do
```
briri, line 288: 👍 I think this approach can be useful elsewhere as well.

briri (approved):

nicolasfranck:
`org.plans` seems to be replaced by `org.org_admin_plans` on many places in release `3.0.3`, except for this [line](https://github.com/DMPRoadmap/roadmap/blob/master/app/views/plans/_navigation.html.erb#L25).
Is that line doing the right thing then?

PR #2727, opened 2020-11-17
Fix preload

nicolasfranck:
Fixes # .

Changes proposed in this PR:

* add more preload in the some controllers
* do not use typical activerecord methods like `count`, `find`, `find_by`, `where` or `order` on relations, as they remove the advantage given by relations by reexecuting the query (see logs). Ok, yes they hit the cache, but everything runs much faster without even trying to reexecute:
  * replace `count` by `size`
  * replace `where` and `find_by` by `select`
  * replace `order(:attr)` by `sort`
* apparently preloading through-relations does not work well (see /plans/:id/overview bellow), so I changed it to an explicit path.

There are probably other places too, but these queries were blocking for me.


nicolasfranck:
mm, I sincerely do not know what is supposed to be wrong with my code according to these tests..

nicolasfranck:
Remark: the "old" `Plan.load_for_phase` sometimes fails (with a null pointer exception because `plan` is null) when there are no sections attached to a phase, so that would fix this too. I know, that should not happen..

app/views/notes/_show.html.erb
```diff
@@ -1,6 +1,6 @@
 <%# locals: { note } %>
 <% if !note.nil? %>
-    <% user = User.find(note.user_id) %>
+    <% user = note.user %>
```
briri, line 3: good catch, no need to go back to the DB to find something that is already loaded

briri (changes requested):
Hi @nicolasfranck. Thanks for this PR. I think all of the changes look good. 

The errors look like they're failing on `/home/runner/work/roadmap/roadmap/config/initializers/devise.rb:7` which is likely due to Rails being unable to properly load/decipher the credentials file (e.g. `config/credentials.yml.enc` or `config/master.key`)
I pulled down your branch and am able to run `rails db:setup RAILS_ENV=test` without any issues. So it seems to be a strange thing happening in GitHub's environment and I don't see how it can be happening.

Can you please try: 
- closing this PR
- pull the latest changes from the development branch into your branch
- run `rubocop` to address the rubocop errors (or refer to the ones flagged in the Rubocop test Github ran)
- send a new PR

xsrust:
Updating this one as well.  See https://github.com/DMPRoadmap/roadmap/pull/2741#issuecomment-732804820 for details

briri:
it looks like some of the tests are failing due to the proposed changes. 

also please run Rubocop and clean up any issues it reports. You can run `rubocop -a` which will attempt to auto-correct as many as it can

nicolasfranck:
I fixed the last rubocop errors, but there is one error that is not related to my PR:

```
app/helpers/conditions_helper.rb:85:3: C: Metrics/AbcSize: Assignment Branch Condition size for num_section_questions is too high. [<6, 25, 10> 27.59/25] (http://c2.com/cgi/wiki?AbcMetric, https://en.wikipedia.org/wiki/ABC_Software_Metric)
  def num_section_questions(plan, section, phase = nil) ...
  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

nicolasfranck:
Crap, now it is failing because `spec/features/feedback_requests_spec.rb` cannot find a link `Request feedback`.
I had a look at `app/views/plans/_request_feedback_form.html.erb`, and it checks if the current user has [administrator rights](https://github.com/DMPRoadmap/roadmap/blob/development/app/views/plans/_request_feedback_form.html.erb#L3),
while the check in the [navigation](https://github.com/DMPRoadmap/roadmap/blob/development/app/views/plans/_navigation.html.erb#L25) is less strict.
And so you get an empty page..

Anyway, that failure is unrelated? No idea why it does succeed in the postgres test suite.

briri:
the mysql failure is related to a random bug we have in RSpec that is likely related to the Faker gem's data that is used. I just restarted the test

nicolasfranck:
@briri no, the problem is this:

https://github.com/DMPRoadmap/roadmap/blob/development/app/views/plans/_request_feedback_form.html.erb#L1 uses `@plan.administrable_by?(current_user)` but that method only accept a user identifier:

https://github.com/DMPRoadmap/roadmap/blob/development/app/models/plan.rb#L400

There are other places were this is misused (or maybe the method is wrong? I would prefer one with an object instead of an id):

* app/views/paginable/contributors/_index.html.erb
* app/views/plans/_share_form.html.erb
* app/views/plans/_navigation.html.erb

I probably missed some

nicolasfranck:
I've added a PR to fix that: https://github.com/DMPRoadmap/roadmap/pull/2768

nicolasfranck:
Now it failing because of an unrelated issue with`libffi`: `libffi.so.6: cannot open shared object file`

briri:
I'm not sure what's going on here. I just created #2810 from `development` made a small text change to see if it would have the same error with`libffi`.  That PR's tests pass.

nicolasfranck:
Is there a way to rerun the checks?

briri:
@nicolasfranck, for future reference, you can restart them by navigating through to the check/test that is failing and clicking 'Re-run jobs' 
Screen Shot 2021-03-08 at 7 58 59 AM

The dev team will be going through these PRs you have submitted this week and will get a new release out.

nicolasfranck:
@briri external users cannot do that, I suppose. I don't see it on my part

briri:
ah ok. I did not realize it was restricted

PR #2731, opened 2020-11-18
added version question to issue_template.md

briri:
Add a question to the ISSUE_TEMPLATE.md asking for the version number they are reporting the issue for.

Update Github workflows to fix issue with env variable assignments


raycarrick-ed (approved):
Looks good. Note if we ever get to double digits (2.10 or 10.2) the grep will produce the wrong answer. Unlikely but worth noting.

PR #2733, opened 2020-11-18
Minor change to external API services to use `active?` instead of `active`

briri:
Minor change to make the true/false `active` methods of the external API services to be `active?` instead

also (per Rubocop) updated a call to the `STDOUT` constant to use the`$stdout` class variable instead


app/services/external_apis/base_service.rb
```diff
@@ -114,7 +114,7 @@ def options(additional_headers: {}, debug: false)
           headers: headers.merge(additional_headers),
           follow_redirects: true
         }
-        hash[:debug_output] = STDOUT if debug
+        hash[:debug_output] = $stdout if debug
```
xsrust, line 117: Cool! I'd not seen the `STDOUT` `$stdout` difference before.

xsrust (approved):

app/services/external_apis/base_service.rb
briri: we either until Rubocop flagged it ;)

PR #2741, opened 2020-11-22
include gem uglifier for production

nicolasfranck:
If you run

```
bin/rails assets:precompile
```

in production mode, you'll get a stack trace

```
yarn install v1.22.5
[1/4] Resolving packages...
success Already up-to-date.
Done in 0.64s.
rails aborted!
LoadError: cannot load such file -- uglifier
/home/install/roadmap-ugent/vendor/bundle/ruby/2.6.0/gems/bootsnap-1.4.6/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:34:in `require'
/home/install/roadmap-ugent/vendor/bundle/ruby/2.6.0/gems/activesupport-5.2.4.3/lib/active_support/dependencies.rb:291:in `block in require'
/home/install/roadmap-ugent/vendor/bundle/ruby/2.6.0/gems/activesupport-5.2.4.3/lib/active_support/dependencies.rb:257:in `load_dependency'
/home/install/roadmap-ugent/vendor/bundle/ruby/2.6.0/gems/activesupport-5.2.4.3/lib/active_support/dependencies.rb:291:in `require'
/home/install/roadmap-ugent/vendor/bundle/ruby/2.6.0/gems/sprockets-4.0.2/lib/sprockets/autoload/uglifier.rb:2:in `<main>'
[..]
```

I'm not sure if this gem is actually needed (we have webpack now, not?),
but in case we do, I added it to the Gemfile, and now it stops complaining.


xsrust:
@nicolasfranck I've made some changes to the github-actions definitions in the `development` branch so I'm going to update your PRs  (I believe it will make changes to the branch on your fork).

Looks like the failures on MySQL and PostgreSQL are due to PRs from forks being unable to use github-secrets as part of the actions workflow.  I've updated this workflow so instead of storing the `master.key` in github-secrets, the credentials will just get re-generated each run.

briri:
thanks. yes, webpacker should be doing all of the minification. We will look into the bootsnap connection and see if we can update it to prevent the error/warning

nicolasfranck:
strange, the mysql tests fail here too with unrelated errors. Seems like some things were added before without running those tests?

xsrust (approved):
Thanks for updating this @nicolasfranck 

Rebased this off development but tests were passing before so should continue to pass.
The MySQL failure before seems to have been a one-off, as re-running resolved the issue.

xsrust:
Merging as all passing

PR #2743, opened 2020-11-23
Github-Actions fixes

xsrust:
Added punctiontion to README.md to make github-actions run.

We're experiencing issues across other PRs aimed at development so starting a new PR to troubleshoot the cause.


.github/workflows/postgres.yml
```diff
@@ -57,7 +56,8 @@ jobs:
         cp config/database.yml.sample config/database.yml
         cp config/initializers/contact_us.rb.example config/initializers/contact_us.rb
         cp config/initializers/wicked_pdf.rb.example config/initializers/wicked_pdf.rb
-        cp config/credentials.yml.enc.workflow config/credentials.yml.enc
+        # generate a default credential file and key
+        bundle exec EDITOR='echo "$(cat config/credentials.yml.example)" >' rails credentials:edit
```
xsrust, line 60: It looks like you can play with the `EDITOR` env-var when generating the credentials, so it should be possible to set these to the contents of the example file within the runtime instead of storing the master.key

briri, line 60: Nice. That should always guarantee that the key and credentials work for the test system

briri:
looks like a chicken vs egg issue. invoking bundler before its been installed. Try moving that bundler install so it runs before `credentials:edit`

xsrust:
I added the same changes to the MySQL version of the workflow and have removed the old workflow-credentials file, so this should be ready for review now

briri (approved):
thanks @xsrust. Looks good

PR #2751, opened 2020-12-01
Fix org links

### LINKED ISSUE #2737: Unable to add links on Org details page

v3.0.0

**Expected behaviour:**
Links are saved when you add them on the Org Details page

**Actual behaviour:**
You receive a 'Successfully saved the Organization' message but the link is not saved. 

I suspect that the params are coming into the controller differently after we upgraded to Rails 5 with strong parameters. It would be beneficial to add some tests around this to catch issues in the future

**Steps to reproduce:**
- Log in as an admin.
- Click 'Organisations' in the admin menu
- Edit an Org 
- Add a link (you may also need to add a contact email for unmanaged Orgs)
- Click save

briri:
Fixes #2737 .

Changes proposed in this PR:
- Links are not sent across within the Strong Parameter context. Updated code to reference `params[org_links]` instead. 

Ideally we would pass these through within the context of `'org': { ... }` but that would require a significant refactor of the views since the links view and JS are generic. This fix will work in the interim


raycarrick-ed (approved):
Looks fine to me.

PR #2753, opened 2020-12-02
@orgs should be orgs; use shib-ds[org-id]

### LINKED ISSUE #2686: orgs_controller#shibboleth_ds view expects list of Org, not Identifier

**Expected behaviour:**

**Actual behaviour:**

* `orgs_controller#shibboleth_ds` selects a list of `Identifier` and stores it in `@orgs`
    (which crashes btw because it cannot sort on `name`): https://github.com/DMPRoadmap/roadmap/blob/master/app/controllers/orgs_controller.rb#L92

* `app/views/orgs/shibboleth_ds.html.erb` uses `@orgs` as if they are organisations ("Select your organisation")

Unused code?

**Steps to reproduce:**

* checkout version 2.2.0
* logout
* make sure there are Identifier of type `Org` and scheme `shibboleth`
* go to `/orgs/shibboleth`

nicolasfranck:
Fixes #2686


nicolasfranck:
partial `shared/_accessible_combobox` seems to have disappeared too. Remove it?

app/controllers/orgs_controller.rb
```diff
@@ -113,8 +114,8 @@ def shibboleth_ds
   # POST /orgs/shibboleth_ds
   # rubocop:disable Metrics/AbcSize
   def shibboleth_ds_passthru
-    if !shib_params["shib-ds"][:org_name].blank?
-      session["org_id"] = shib_params["shib-ds"][:org_name]
+    if !shib_params["shib-ds"][:org_id].blank?
+      session["org_id"] = shib_params["shib-ds"][:org_id]
```
briri, line 118: Yes, storing the org_id in the session variable called 'org_id' makes sense. ha ha. good catch

briri (changes requested):
Looks good @nicolasfranck. You just need to update the `spec/controllers/orgs_controller_spec.rb` test to work with the change you made to the `@orgs` variable. It is looking for an Identifier object instead.

`expect(assigns(:orgs).include?(@identifier)).to eql(true)`

briri:
> partial `shared/_accessible_combobox` seems to have disappeared too. Remove it?

Yes. This is what I'm using in the DMPTool customization of that page:
```ruby
            <%= form_for Org.new, url: shibboleth_ds_path, method: :post do |f| %>
              <div id="shib-ds-org-controls" class="row">
                <div class="form-group col-md-12">
                  <%= render partial: "shared/org_selectors/local_only",
                             locals: {
                                 form: f,
                                 id_field: :id,
                                 default_org: nil,
                                 orgs: participating,
                                 required: false,
                                 label: _("Look up your institution here")
                              } %>
                </div>
              </div>
              <%= button_tag _('Go'), class: "btn btn-default", id: "org-select-go", disabled: "disabled" %>
            <% end %>
```

This ties into the newer org_selector partials (the Org typeahead boxes instead of select boxes. The local-only version ensures that it will not go out to the ROR API or allow the user to type whatever they want). The `participating` in my code example above is all of the Orgs that have a Shibboleth entityID defined in the identifiers table.

briri:
thanks for updating the tests @nicolasfranck. It looks like one is failing. I think you can address it by updating the controller from `if !shib_params["shib-ds"][:org_id].blank?` to `if !shib_params.fetch("shib-ds", {})[:org_id].blank?`

nicolasfranck:
So that partial should be removed then?

While updating the branch
I saw that two routes ..

* `post "/orgs/shibboleth"`
* `get "/orgs/shibboleth/:org_name"`

lead to `orgs#shibboleth_ds_passthru`,
but both get a different input (respectively `shib-ds[org_id]`
vs `org_name`)

While not simply replace both to `org_id`?

briri:
Hmm. We actually have a lot of cleanup to do in that routes.rb file. It looks like the GET path is not used, or at least I'm not seeing anywhere. 

I have some customizations to the baseline DMPRoadmap code for the Shib login, but here's a snippet of the logs when I login. It only appears to be using the POST method to initiate the Shib handshake with the IdP. The `users/omniauth_callbacks_controller.rb` handles the response back:
```shell
# User selects an Org from the list and clicks the button
Started POST "/orgs/shibboleth/15"
Processing by OrgsController#shibboleth_ds_passthru as HTML

# Shibd SP redirects to the Org's IdP based on the entityID
Redirected to https://my.site.org/Shibboleth.sso/Login?target=https://my.site.org/users/auth/shibboleth/callback&entityID=urn:mace:incommon:my.org
Completed 302 Found in 12ms (ActiveRecord: 4.4ms)

# User signs into IdP and gets redirected back
Started GET "/users/auth/shibboleth/callback"
omniauth: (shibboleth) Callback phase initiated.
Processing by Users::OmniauthCallbacksController#shibboleth as HTML

# Dashboard page gets rendered (assuming successful auth for an existing user)
Redirected to https://my.site.org/
Completed 302 Found in 60ms (ActiveRecord: 41.5ms)
Started GET "/"
Processing by HomeController#index as HTML
```

My guess is that the GET one can be removed.

nicolasfranck:
@briri the get-route is used: https://github.com/DMPRoadmap/roadmap/blob/development/app/views/orgs/shibboleth_ds.html.erb#L67

PR #2754, opened 2020-12-02
open omniauth orcid controller in same window

### LINKED ISSUE #2752: "Link credentials" on profile page opens a new window

3.0.0

**Actual behaviour:**

The links "link to your orcid" or "link your institutional credentials" are opened
in a new window where all authentication happens, and after a successfull login
you're redirected to your profile page.

But meanwhile there is still an outdated profile page open. Confusing?

I've seen this behaviour in applications that use this technique to shield
themselves from the authentication layer, but at the end of the authentication
a piece of javascript sends a signal to the opening page, and closes itself.
But that is not the case here.

nicolasfranck:
Fixes #2752


briri (approved):
I think users understand that any of these oauth links would send them to the relevant site to authenticate so removing the `target: '_blank'` makes sense. 

Although we would not be following the blanket guidance on accessibility we received about directing users off-site, have two browser tabs/windows open is likely a worse user experience ;)

What do you think @xsrust?

xsrust:
I lean towards User-experience over the accessibility.
We could always add a `aria-hidden` or `sr-only` to warn screen-reader users that these links will take them off-site?

pherterich:
Things open in the same tab - not sure it's currently linking to the right place, the institutional login points me back to my dashboard and ORCID sends me to ORCID (sandbox?) with an error message which probably is the right thing coming from test.

briri:
yes, that's probably alright. It should go to the ORCID sandbox when testing and the Shibboleth issue is likely just a configuration problem. @xsrust can you verify?

PR #2759, opened 2020-12-07
Bugfix for issue with setting plan visibility

### LINKED ISSUE #2745: Unable to set plan visibility to organization
An admin user reported that plans that they have set to be visible on the organization dashboard do not appear. I was able to recreate this issue in my own account.

For example, this plan is set to be available for the UCOP org: https://dmptool.org/plans/56493

But does not appear on University of California, Office of the President (UCOP) Plans table.

briri:
Fixes #2745.

Changes proposed in this PR:
- Replaced old `no_questions_matches_no_answers?` with `percent_answered` on the Plan model which returns the percentage of answered questions rather than a boolean which was doing an equality check between questions and answers
- Updated the Answer model to use the new `percent_answered` method and then compare the returned percentage to the one defined in `config/initializers/_dmproadmap.rb`
- Updated tests

### DIFF
--- a/app/models/answer.rb
+++ b/app/models/answer.rb
@@ -142,9 +142,11 @@ class Answer < ApplicationRecord
     # Remove guard? this is an after-save so unreachable if there is no plan
     return unless plan_id?
 
-    complete = plan.no_questions_matches_no_answers?
-    if plan.complete != complete
-      plan.update!(complete: complete)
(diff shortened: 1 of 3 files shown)


xsrust (approved):
Thanks, the updates look more robust and I appreciate the guard-statements being added to those methods.

PR #2768, opened 2020-12-16
fix incorrect use of Plan#administerable_by?

nicolasfranck:
Changes proposed in this PR:

While `Plan#administerable_by?` request a user identifier,
some views provide a user record.


nicolasfranck:
not sure why `spec/features/questions/textarea_questions_spec.rb:20` fails. Maybe `sleep(0.2)` is too low?

nicolasfranck:
I did some tests again that `spec/features/questions/textarea_questions_spec.rb`:

* it also fails in branch development
* changing the sleep command does not help
* I dumped the page html, and found this in `#answer-status-1` (where "Answered just now" is expected):
```
<div id="answer-status-1" class="mt-10">
  <!--
    Partial for handling the answer status (e.g. saving, error-saving, data-status)
  -->
  <span class="label label-info status" style="display:none;" data-status="saving">Saving...</span>
  <span class="label label-warning status" style="display: none;" data-status="error-saving"></span>
</div>
```

So it seems like the post-request was sent, and an answer is to be expected.
Increasing the timeout did not help, so maybe the javascript call fails somewhere
else and therefore cannot update the status?

nicolasfranck:
pfew, I merged in the development branch, and it started working. I believe that javascript error with tinymce_content css not being loaded was fixed there, so maybe that was the cause?

briri:
thanks @nicolasfranck these changes look good. We will need to schedule some thorough UAT for this in January though. So far the areas concerned (e.g. providing feedback, adding contributors, etc.) have all been tested and were providing the expected functionality (even though this was failing). So we will need to ensure that the fix continues to provide the expected results.

nicolasfranck:
The reason why `Plan#administerable_by?(user_id)`  "worked" in the past was this:

The argument `user_id` that is inserted into `Role.administrator.where(plan_id: id, user_id: user_id, active: true).any?`
is interpreted when it is not a regular integer/string. Apparently this only happens if you use a where-clause.

For example, these calls generate the same query:

```
plan.administerable_by?(current_user)
plan.administerable_by?(current_user.id)
```

which maybe handy at first, but conceals incorrect use (as shown in the examples above)
(and it is also very confusing).

BTW: I made these changes in the first place to fix my previous PR https://github.com/DMPRoadmap/roadmap/pull/2727
where a `where` clause is not used

briri (approved):
looks good and just verified that its working as expected.

PR #2771, opened 2020-12-18
return from redirect_to; fix typo

### LINKED ISSUE #2684: orgs_controller#shibboleth_ds: redirect should return

**Expected behaviour:**

**Actual behaviour:**

Rails complains that rendering happens twice.

Reason: https://github.com/DMPRoadmap/roadmap/blob/master/app/controllers/orgs_controller.rb#L88

The `redirect` does not issue a return from the function `shibboleth_ds` and therefore
https://github.com/DMPRoadmap/roadmap/blob/master/app/controllers/orgs_controller.rb#L96 is executed also.

**Steps to reproduce:**

* checkout version 2.2.0
* login
* make sure there are no organisations with a shibboleth scheme identifier
* go to `/orgs/shibboleth`

nicolasfranck:
Fixes #2684

### DIFF
--- a/app/controllers/super_admin/themes_controller.rb
+++ b/app/controllers/super_admin/themes_controller.rb
@@ -51,7 +51,7 @@ module SuperAdmin
         redirect_to super_admin_themes_path, notice: msg
       else
         flash.now[:alert] = failure_message(@theme, _("delete"))
-        redner :edit
+        render :edit
       end
     end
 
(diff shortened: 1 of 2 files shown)


app/controllers/super_admin/themes_controller.rb
```diff
@@ -51,7 +51,7 @@ def destroy
         redirect_to super_admin_themes_path, notice: msg
       else
         flash.now[:alert] = failure_message(@theme, _("delete"))
-        redner :edit
+        render :edit
```
briri, line 54: good catch

briri (approved):
this looks good. thanks @nicolasfranck

PR #2772, opened 2020-12-18
sync translations from translation.io

xsrust:
Sync translations to/from Translation.io in preparation for release


briri (approved):
Thanks for syncing these @xsrust

PR #2775, opened 2020-12-21
compare time string against time for the same time zone

### LINKED ISSUE #2774: spec test for Api::V1::DeserializationService#safe_date expects wrong results

https://github.com/DMPRoadmap/roadmap/tree/v3.0.1

**Expected behaviour:**

This spec https://github.com/DMPRoadmap/roadmap/blob/v3.0.1/spec/services/api/v1/deserialization_service_spec.rb
should succeed, no matter the time zone it runs in.

**Actual behaviour:**

If you run this spec in time zone that is higher than UTC (e.g. Brussels),
then some these tests fail:

https://github.com/DMPRoadmap/roadmap/blob/v3.0.1/spec/services/api/v1/deserialization_service_spec.rb#L188

**Steps to reproduce:**

* Run this from a different timezone higher than UTC

What happens in `safe_date`:

* The variable `expected` with value `2020-11-11`  is parsed by `safe_date` as a local time into an appropriate Time object with time zone support: `2020-11-11T00:00:00+01:00` 
* That result is then converted to UTC: `2020-11-10T23:00:00Z` 
* the spec expects the string to start with `2020-11-11`, which fails in this timezone

What should happen:

* compare `expected` to `[Time object].localtime.to_s` instead of to the stringified version of UTC.

nicolasfranck:
Fixes #2774 .

### DIFF
--- a/spec/services/api/v1/deserialization_service_spec.rb
+++ b/spec/services/api/v1/deserialization_service_spec.rb
@@ -196,21 +196,21 @@ RSpec.describe Api::V1::DeserializationService do
     it "converts :value to a UTC format" do
       now = Time.now
       expected = "2020-11-11"
-      rslt = described_class.safe_date(value: "2020-11-11").to_s
-      expect(rslt.start_with?(expected)).to eql(true)
-      expect(rslt.end_with?("UTC")).to eql(true)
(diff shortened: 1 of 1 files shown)


briri (approved):
thanks for addressing this

PR #2792, opened 2021-01-27
see #2782: remove git depended code

### LINKED ISSUE #2782: VersionHelper does not work outside of git

3.0.1

**Expected behaviour:**

**Actual behaviour:**

When you use this application,
without the full git repository (`.git`),
then the startup crashes:

```
app/helpers/version_helper.rb:7:in `<module:VersionHelper>': undefined method `strip' for nil:NilClass (NoMethodError)
```

Reason: https://github.com/DMPRoadmap/roadmap/blob/master/app/helpers/version_helper.rb#L7

**Steps to reproduce:**

* remove the git folder
* `bin/rails server`

Why I want to do this: I want to prepackage this application (including ruby dependencies with bundler)
into an rpm. The git repository does not belong in that rpm, so that
is not included as well. Is this version helper even used?
What about prepackaged releases?

nicolasfranck:
Fixes #2782 .

Changes proposed in this PR:
- removes `VersionHelper` that depends on the command `git`
- removes lines from views that use that helpers methods


briri:
this looks good. Thanks @nicolasfranck

briri (approved):

PR #2802, opened 2021-02-02
Fix issue with one click plan creation

### LINKED ISSUE #2800: One click plan creation is not working
When clicking on the button to copy a link to create a new plan (from here: https://dmptool.org/public_templates) I am directed to an empty link box. Brian was able to reproduce the issue on his end. Looks like searching the plans table is breaking the link creation, as if you don't run a search first you do get the link. In my case I was searching for NIH templates and did not get links to copy. 

screenshot-dmptool org-2021 01 29-15_22_12

briri:
Fixes #2800 .

Changes proposed in this PR:
- Switch JS for the one click plan creation controls to use event bubbling. Search and pagination controls for the table are making AJAX calls so the re-rendered content was not wired up to the JS.


raycarrick-ed (approved):
Looks functionally equivalent. Makes sense using child selector.

PR #2803, opened 2021-02-03
current page works with multi tenancy where active page doesnt

raycarrick-ed:
Fixes #https://github.com/DigitalCurationCentre/DMPonline-Service/issues/365 .

Changes proposed in this PR:
- use current_page? rather than active_page? to determine if you are on the home page. Works with multi tenancy where active doesn't.


briri (approved):
this looks good @raycarrick-ed I addressed the issue with the failing mySQL test in #2808

briri:
@raycarrick-ed I'm not sure if this one is still relevant. If so, I think you just need to merge in the latest from master and resolve the conflict.

raycarrick-ed:
No longer relevant.

PR #2808, opened 2021-02-04
Bump wkhtmltopdf-binary to fix GitHub actions

briri:
Trying to fix broken Github action


briri:
Merging this to fix the Github actions which are now causing all PRs to fail

PR #2828, opened 2021-03-08
fixed bug with funder_only org selection. Added a 'Searching ...' mes…

briri:
Reported in [DMPTool#246](https://github.com/CDLUC3/dmptool/issues/246).

The funder_only restriction for some of the Org search boxes was not limiting the results to Orgs that have a Crossref funder id. 
- Fixed that bug in the controller.
- Updated the JS logic to remove the current search results and add a 'Searching ...' message that appears until the Ajax response if returned. This should help prevent users from adding their own text in scenarios where the results take too much time.


raycarrick-ed (approved):
Code looks good to me. Haven't looked into the mysql failure but assume it's not really to do with this.

PR #2836, opened 2021-03-16
Orgs Autocomplete

raycarrick-ed:
The line used to be:
    if orgs.present? && params.fetch(:funder_only, "false") == true
but that didn't work because the param is a string so it was changed to
    if orgs.present? && params.fetch(:funder_only, false)
which looks better but doesn't work either for the same reason
(try something like
     puts "hello" if "false"
)
so it needs to be
if orgs.present? && params.fetch(:funder_only, "false") == "true"


briri (approved):
looks good @raycarrick

PR #2837, opened 2021-03-17
Update RDA metadata standard URL

briri:
The Rake task used to initialize the RDA metadata standard question format used an old dmponline-test URL. This patch updates it to its current home

### DIFF
--- a/lib/tasks/initialize_data.rake
+++ b/lib/tasks/initialize_data.rake
@@ -13,7 +13,7 @@ namespace :initialize_data do
     end
     rda_q.option_based = false # keeping this false as options not stored locally
     rda_q.formattype = QuestionFormat.formattypes[:rda_metadata]
-    rda_q.description = "https://dmponline-test.dcc.ac.uk/rda/api/" # TODO: Update to permanant API address once HTTPS is added
+    rda_q.description = "https://rdamsc.bath.ac.uk/api/"
     if rda_q.save
       puts 'Sucessfully added/updated'
     else


nicolasfranck:
The new api url works partially. See my browser console output:

```
XHR finished loading: GET "https://dmpplay.dmponline.be/question_formats/rda_api_address".
XHR finished loading: GET "https://rdamsc.bath.ac.uk/api/subject-index".
11:30:46.680 edit:1 Access to XMLHttpRequest at 'https://rdamsc.bath.ac.uk/query/schemes' from origin 'https://dmpplay.dmponline.be' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource.
11:30:46.681 jquery.js:8817 POST https://rdamsc.bath.ac.uk/query/schemes net::ERR_FAILED
```

Some CORS related problem. 

So now some select boxes appear, and then .. nothing. It says "Please wait, Standards are loading".

Is this expected? Are you required to run your own rda server instance, or can you use the "main" one?
I haven't found any documentation about this here, except the github of https://github.com/rd-alliance/metadata-catalog-v2 itself.

nicolasfranck:
Oops, you're using v1: https://github.com/rd-alliance/metadata-catalog-dev.

nicolasfranck:
Is there is way to "fix" this (both the CORS problem and the missing route), or are we expected to run our own server for that?

briri:
Hi @nicolasfranck 

DCC no longer hosts that API. Its not located at the University of Bath in the UK. Their implementation seems to be quite different though.

We have a new customization in DMPTool that pulls down the metadata standards from the RDA site, sticks the records into a DB table. We did not have anyone using the template question type for the RDA metadata standard though so I have not yet refactored that part of the code to use the new data structure.

I will be submitting a PR later this month for the new research outputs tab that will contain this code along with a re3data repository selector, spdx license selector, and a field of science selector.

If you want to update/patch your own instance before I get around to submitting the PR you can refer to the following:
- Add the `metadata_standards` table: https://github.com/CDLUC3/dmptool/blob/0a906ffb1e3e936ecbe8ae68038dd22298f61c4b/db/migrate/20210317202247_create_metadata_standard_categories.rb#L12 (note that you would just want the single table there if you use this)
- Add a corresponding model: https://github.com/CDLUC3/dmptool/blob/main/app/models/metadata_standard.rb
- Add a new external_api service that fetches the standards from the RDA API and loads them into the table: https://github.com/CDLUC3/dmptool/blob/main/app/services/external_apis/rdamsc_service.rb
- Add a rake task that calls the API to populate the table: https://github.com/CDLUC3/dmptool/blob/0a906ffb1e3e936ecbe8ae68038dd22298f61c4b/lib/tasks/utils/external_api.rake#L107

nicolasfranck:
@briri Ok, thanks, I'll wait then ;-)

PR #2851, opened 2021-03-30
Org selection options

raycarrick-ed:
Introduce switch in config (in _dmproadmap.rb) which uses the local_only partial for org selection and doesn't allow on the fly account creation.


app/presenters/org_selection_presenter.rb
```diff
@@ -27,7 +27,11 @@ def crosswalk
   end
 
   def select_list
-    @crosswalk.map { |rec| rec[:name] }.to_json
+    begin
+      @crosswalk.map { |rec| rec[:name] }.to_json
+    rescue
```
briri, line 32: No need for the begin here and RuboCop prefers specifying an error type. Use:
``` ruby
def select_list
  @crosswalk.map { |rec| rec[:name] }.to_json
rescue StandardError => e
  nil
end
```

briri (approved):
looks good @raycarrick. Just some minor Rubocop cleanup. Feel free to merge once its happy.

PR #2858, opened 2021-04-01
Org selector refactor

briri:
@raycarrick-ed I noticed that the 'Add contributor' page was failing because it was not setting the `org_partial`. I also noticed that the 'Sign in' menu was still trying to open the modal sign in/create account form that you removed.

This PR refactors those recent changes a bit.
- Move setting of `@all_orgs` and `@org_partial` to the OrgSelectable concern which sets them in a `before_action` hook
- Update 'Sign in' link in nav menu to send user to the `root_path` instead of trying to open the modal popup that was removed


raycarrick-ed (approved):
That is a lot neater. Excellent.

PR #2861, opened 2021-04-02
plan/edit: produce less queries

nicolasfranck:
Changes proposed in this PR:
- add relations `question_format question_options annotations themes` to list of `include` in `PlansController`
- remove `include` in `Section#answered_questions`, because already done by `PlansController` and because it breaks the chain of relations
- replace `where` and `order` by `select` and `sort` in model `Question` because it reexecutes the queries instead of using the relations
- Optimize `GuidancePresenter#guidance_groups_by_theme`
- Optimize `ConditionsHelper`

You should see less queries in the log when opening the tab "Write plan"

P.S. Is `ConditionsHelper#num_section_questions` still expected to receive `section` and `phase` as hashes, or is this kept for backward compatibility for those who are branding their views?

### DIFF
--- a/app/controllers/plans_controller.rb
+++ b/app/controllers/plans_controller.rb
@@ -228,7 +228,16 @@ class PlansController < ApplicationController
   #       doing this when we refactor the Plan editing UI
   # GET /plans/:plan_id/phases/:id/edit
   def edit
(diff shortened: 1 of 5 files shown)


briri (approved):
thanks for this refactor @nicolasfranck

PR #2862, opened 2021-04-06
redirect to edit page of notification after successfull creation/update

nicolasfranck:
Changes proposed in this PR:
- redirect to edit page of notifications controller after create
- redirect to edit page of notifications controller after update

Just trying to prevent a user from trying to reload the page by
hitting the url as presented in the url bar.


briri (approved):
changes look good and functionality will be consistent with other pages. Thanks @nicolasfranck

PR #2863, opened 2021-04-09
Share plan org

raycarrick-ed:
When sharing a plan with a new user, when the new user followed the link in the invitation email and filled in the new user form, they got an "org needs set" error because the org params coming back to devise where not as needed.

This adds a before-action to fix up the params for devise.


briri (approved):
Looks good @raycarrick-ed thanks for catching/patching this

briri:
Looks like a minor rubocop issue with extra spaces. Once that's updated feel free to merge yourself 👍🏻

PR #2865, opened 2021-04-12
removed check for @public_plan on plan export. It was preventing the …

briri:
Fixes https://github.com/DigitalCurationCentre/DMPonline-Service/issues/491.

Hey @raycarrick-ed, I'm not sure why, but the logic was set so that it would not show the question text if the plan was publicly visible. I looked back through the git history and see that it was in place back in 2018. Perhaps `@public_plan` was not being properly set before and this was not an issue?

Anyway I removed that check so that all public plans now show the question text.

### DIFF
--- a/app/views/shared/export/_plan.erb
+++ b/app/views/shared/export/_plan.erb
@@ -35,7 +35,7 @@
                 <% next %>
               <% end %>
               <div class="question">
-                <% if !@public_plan && @show_sections_questions%>
+                <% if @show_sections_questions%>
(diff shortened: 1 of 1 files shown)


magdalenadrafiova:
I tried with the following plan: https://dmponline-dev.dcc.ac.uk/plans/73839/export.pdf?export%5Bquestion_headings%5D=true and it worked well: 
Screenshot 2021-04-13 at 11 17 52

PR #2867, opened 2021-04-12
Prevent the 'remove' link on Orgs page if it has contributors or plans

briri:
Fixes https://github.com/DigitalCurationCentre/DMPonline-Service/issues/498

@raycarrick-ed I think this will address the issue you were seeing with this one. I think its worth considering though if that is the behavior we want. If its actually ok to delete an Org despite it having those associated records, we could add a callback that nils the association first (on the plan and contributor records) before deleting the Org record.

Also note that it doesn't check for guidance which is also a relation but I think the cascading delete logic handles those.

### DIFF
--- a/app/controllers/super_admin/orgs_controller.rb
+++ b/app/controllers/super_admin/orgs_controller.rb
@@ -12,7 +12,7 @@ module SuperAdmin
     def index
       authorize Org
       render "index", locals: {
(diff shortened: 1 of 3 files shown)


raycarrick-ed:
Agree. We can think through the logic again later but for now this looks like the safest solution.

PR #2868, opened 2021-04-12
Fixed issue with enabling/disabling notifications from table

briri:
Fixes https://github.com/DigitalCurationCentre/DMPonline-Service/issues/480

Updated code to use Rails 5+ new data remote logic for input elements. Also needed to update controller to grab the enabled flag value from `params`


magdalenadrafiova:
notification removing the tick - works now on the dev

PR #2874, opened 2021-04-15
Fix share required params

nicolasfranck:
Changes proposed in this PR:
- check on existance of `:email` in hash `:user`. If you do not supply an email address now, the controller crashes
- make fields email and access required in html


briri (approved):
this looks good. Thank you @nicolasfranck!

PR #2883, opened 2021-04-20
pdf export filename qith quotes fails in chrome

raycarrick-ed:
Filename for pdf export is derived from plan title. If it has quotes in it, this fails in chrome. This removes the quotes.


briri (approved):

briri:
looks good @raycarrick-ed just a very minor rubocop complaint then go ahead and merge after that

PR #2892, opened 2021-05-05
fix issue with investigator and data_Contact info for API v0

### LINKED ISSUE #2870: API - issue reported by the user for API v0 and V1
@briri When trying to get the full list of plans through API V0 or V1, I do not seem to be able to collect the most recent plans, even when processing through all pages as suggested in the documentation. V1 provides the most comprehensive list of plans (330 entries) but the latest plan I seem to be able to read was created back in mid-February, and the number of items in this list has been stable since I first tried the API in February.

briri:
Fixes #2870.

Changes proposed in this PR:
- We moved the investigator and data_contact info from the plans table to the contributors table. The JSON was trying to display the investigator and data_contact info even if none had been defined for the plan which was resulting in an error due to undefined method on a nil object

@raycarrick-ed this was the error reported for me in the DMPTool (ours has the same issue but we have no one actively using v0). I am unsure if it is the same for your installations.

### DIFF
--- a/app/views/api/v0/plans/index.json.jbuilder
+++ b/app/views/api/v0/plans/index.json.jbuilder
@@ -33,9 +33,7 @@ json.array! @plans.each do |plan|
   json.data_contact do
     json.name         data_contact.is_a?(Contributor) ? data_contact.name : data_contact.name(false)
     json.email        data_contact.email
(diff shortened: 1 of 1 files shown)


raycarrick-ed (approved):
I'm not aware of us having issues with this but it looks safer.

PR #2914, opened 2021-05-14
addressed a few minor bugs found while working on the dmptool

briri:
Discovered a few bugs reported in our logs

- Move the call to the Rails `sanitize` method from the Mailer to the View. That method is not available on a Mailer
- Removed the `@` from the `@help_url` in the admin_privileges email so that it uses the Rails route instead
- The Org Admin Edit User page had a typo (probably from a copy/paste). Swapped `presenter.identifiers` out with `@user.identifiers`


raycarrick-ed (approved):
all looks good

PR #2918, opened 2021-05-19
Bug 428 changing org admin rights remain

johnpinto1:
Fixes #428 .

Changes proposed in this PR:
The attribute_changed? callback has been removed in Rails 5.2, so we
    need to use saved_change_to_attribute?.
    
    So in our case, we replaced :org_id_changed? with :saved_change_to_org_id? in the after_update callbacks in the User model:
    
    after_update :delete_perms!, if: :saved_change_to_org_id?, unless: :can_super_admin?
    
      after_update :remove_token!, if: :saved_change_to_org_id?, unless: :can_super_admin?


raycarrick-ed:
Need to check those 2 tests and see what's going on with them.

briri (changes requested):
I think these tests are accurately failing. I think the tests themselves may be inaccurate. 

I just took a look at the `spec/models/user_spec.rb` and it sets the test user up with the following Perms: 'use_api', 'modify_guidance', 'modify_templates'. Those Perms make the user an org_admin not a super_admin, so it technically should clear their Perms and API token.

I think the spec tests need to be updated to make the user a super admin. I would change this line: https://github.com/DMPRoadmap/roadmap/blob/4ca2dc848dbe6a87a5cb25d9443855a86a269776/spec/models/user_spec.rb#L132
to: `create(:user, :super_admin, org: @org, api_token: "barfoo")` to make use of the user factory helper trait

johnpinto1:
@briri Thanks for comments. I looks like I still need to look at why tests still failing after change. Will need to do that next week @raycarrick-ed .

johnpinto1:
@briri & @raycarrick-ed This has been changed to pass tests. I had to change the tests pointed out by @raycarrick-ed

briri (approved):
thanks for fixing the tests @johnpinto1 this looks good to me now

briri:
@raycarrick-ed is out this week, so I'll go ahead and merge

PR #2944, opened 2021-06-11
bug fix: customization upgrade #2898

### LINKED ISSUE #2898: Versionable id not always set on Phases, Sections and Questions
We are seeing records with NULL values for their `versionable_id`

The Versionable concern defines a default for this value but for some reason we are seeing records with a NULL. This could be due to the use explicitly setting it to `nil` somewhere in the code or the ActiveRecord ORM is not using the default in all situations: https://github.com/DMPRoadmap/roadmap/blob/bc32927483d4186daca1cc187bb4b7d5a130442a/app/models/concerns/versionable_model.rb#L14

raycarrick-ed:
Fixes #2898  .

Changes proposed in this PR:
- restructure upgrade_customization_service to remove confusing use of variable and method names.


raycarrick-ed:
didn't include in tests as couldn't get them to run properly. Will come back to that.

briri (approved):

PR #2950, opened 2021-06-18
Versionable id fix 

briri:
Back fills versionable_ids for records affected by #2898.

Changes proposed in this PR:
- mod to the transfer customizations controller so that the template is published, versioned and then unpublished before transferring the changes. Previously it appeared as though any unpublished changes could be lost.
- new `lib/tasks/utils/templates.rake` file with tasks to repair Phase, Section, Question and QuestionOption versionable_ids and one to renumber/reorder them
- removed a few obsolete rake files
- moved a few rake file to the new `lib/tasks/utils/` subdir

The original task that seeded the versionable_ids could not be used for this task since it would overwrite the ids on any existing records.

I have run this against a copy of the DMPTool production data and a cursory examination of the records and how the versionable_ids relate across versions, customizations (and their versions) looks correct. Will do some further investigation to verify next week.


briri:
it takes a long time to run since its instantiating Templates and all of its related child objects. I may spend some time updating so that it just fetches ids next week.

raycarrick-ed:
Ran this on almost current data and had to make 2 changes to get it to run.
1. templates.rake:237 downcase -> underscore to take account QuestionOption needing to be question_options
2. versionable_model:10 needed to comment out attr_readonly or it fails when it tries to update.
AFter that it ran. Didn't take too long - about 10-15 mins but didn't improve the versionable_ids == nil much. Here is a table of them:
< Number Phase nil = 5500/8066
> Number Phase nil = 5708/8066
< Number Section nil = 33291/47924
> Number Section nil = 33590/47924
< Number Question nil = 80005/125494
> Number Question nil = 80597/125494
< Number Question option nil = 50978/93795
> Number Question option nil = 51230/93795

< is after running > is before number is number that are nil / total number. So an improvement but not huge.
Haven't had time to dig into why this is but will do that next.

PR #2954, opened 2021-06-24
only include active roles in plan share tab; don't take inactive role…

### LINKED ISSUE #2951: Update role permissions after deactivation of role creates confusion

`2.0.3`

**Expected behaviour:**

If a user "removes" a plan, the plan isn't really removed.
Merely the associated "role" is internally flagged as "active=false".

The user no longer has access to the plan, but..

..other users that still have access still see this "role" in the share tab,
and they have no way of knowing that the role has been deactivated.
So they may alter the permissions, and unintentionally send
notifications to a user that lost interest?

Not sure about this

nicolasfranck:
Fixes #2951 

Comments:

* I also updated the query for `@plan_roles` in `PlansController#request_feedback`, but that variables isn't used anywhere?

### DIFF
--- a/app/views/plans/_share_form.html.erb
+++ b/app/views/plans/_share_form.html.erb
@@ -41,7 +41,7 @@
 
 <h2><%= _('Manage collaborators')%></h2>
 <p><%= _('Invite specific people to read, edit, or administer your plan. Invitees will receive an email notification that they have access to this plan.') %></p>
-<% if @plan.roles.any? then %>
+<% if @plan_roles.any? then %>
   <table class="table table-hover table-bordered" id="collaborator-table">
     <thead>
       <tr>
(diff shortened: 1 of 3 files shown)


app/views/plans/_share_form.html.erb
```diff
@@ -41,7 +41,7 @@
 
 <h2><%= _('Manage collaborators')%></h2>
 <p><%= _('Invite specific people to read, edit, or administer your plan. Invitees will receive an email notification that they have access to this plan.') %></p>
-<% if @plan.roles.any? then %>
+<% if @plan_roles.any? then %>
```
briri, line 44: great catch!

briri (approved):
This looks good to me. Thanks @nicolasfranck

raycarrick-ed (approved):

Issue #2930, opened 2021-06-03
"view less/view all" should show activity to end user


nicolasfranck:
Please complete the following fields as applicable:

**What version of the DMPRoadmap code are you running? (e.g. v2.2.0)**

`3.0.2`

**What happens:**

Some pages that show multiple records (like `/super_admin/orgs`),
show a few records, and the rest can be loaded by pressing "view more"
or a link to a page.

This has two problems:

* nothing is reflected in the url. So no way to copy and and paste the url to the current view.
  this can easily be done by using the javascript history api, like turbolinks does.
* the loading happens in the background, but there is no indication to the end user. For some
   views like /super_admin/orgs (that joins with templates) it can take a long time,
   and end users are not aware why "it is not working".

PR #2959, opened 2021-06-25
Add pagination/sort/search args to the query string 

briri:
Part of #2930 

- Updates the public plans and public templates pages so that the pagination/sort/search functionality performs a normal http GET instead of an ajax request and ensures that the page, search sort_field and sort_direction are conveyed in the query string.


briri:
@raycarrick-ed please review when you have a chance

PR #2960, opened 2021-06-25
Adds a spinner for all UJS ajax calls

### LINKED ISSUE #1310: Display a spinner when waiting for AJAX requests to be completed
Users have reported confusion when they click, e.g., to expand all accordions on the write plan page or edit a template, and the action takes a long time to be completed. Currently there is no indication that anything is happening. On Tues call we discussed adding a spinner to notify users that the AJAX request is being processed.

briri:
Fixes #1310 and also Part of #2930

- Add a spinner gif
- Add JS logic to display the spinner when a UJS Ajax call begins and then hide it once the request is complete.

Note that in some scenarios (e.g. editing a plan answer) we are not using UJS and are using JQuery's `$.ajax` call directly, so the spinner does not appear. We could update the JS here to allow it to be exported and then have the direct AJAX calls invoke `toggleSpinner();` directly but I don't think its necessary since the longer running ones have visual cues like "Saving ..."


raycarrick-ed (approved):
Looks good.

PR #2961, opened 2021-06-29
Adds an additional "other" role to the contributors page

briri:
We received feedback that the current list of roles was too limited. So adding an "other" category since the user is required to select one but the contributor may not be a Project Admin, Investigator or Data Curator.

This role is carried through into the JSON output as `"role": "other"` without any Credit taxonomy URL since Credit does not have an equivalent


raycarrick-ed (approved):
Looks fine to me.

PR #2963, opened 2021-07-01
could still create new orgs when adding contributor

raycarrick-ed:
Fixes https://github.com/DigitalCurationCentre/DMPonline-Service/issues/534 .

Changes proposed in this PR:
- disallow new org creation if instance has restrict_orgs configured.


briri (approved):
looks good @raycarrick-ed thanks for patching this

Issue #462, opened 2017-06-25
Unpublished guidance is visible to users


sjDCC:
Inist CNRS reported this issue and on investigation it seems to be correct. All individual pieces of guidance seem to be shown on front end regardless of whether it’s published or not.
![guidance-publishing](https://user-images.githubusercontent.com/6169940/27515927-b391f590-59a6-11e7-81c3-725d7272540e.PNG)
![guidance-publishing-2](https://user-images.githubusercontent.com/6169940/27515929-bf361340-59a6-11e7-8150-5c297a880ab0.PNG)

Overall, the guidance group choice display seems to be pretty accurate. I did tests of adding groups and the main part this filtered out ones that were unpublished or didn’t have relevant pieces of guidance e.g. for themes not used in the template. 

I did a test though of unpublishing the Inist guidance group though. This didn’t remove it from the front end display, perhaps because the individual pieces of guidance are published? Logic would suggest that unpublishing the higher-level guidance group would override any individual pieces of guidance being published.

![guidance-group-choices](https://user-images.githubusercontent.com/6169940/27515933-cce2945a-59a6-11e7-99b5-c4b0964b04b5.PNG)
![guidance-group-choices-admin](https://user-images.githubusercontent.com/6169940/27515938-d0835ce8-59a6-11e7-9d63-8dc2f3f6c5c5.PNG)

xsrust:
I've gone through the plan details and plan display pages and ensured unpublished guidance will no longer display, and un-published guidance groups will be hidden from the selection menu.

They do stay selected for the plans, so unpublishing followed by publishing will result in no change for the end-users.
While unpublished, the guidances will not display on their plan.

sjDCC:
This is working now. Closing out

PR #2970, opened 2021-07-14
Issue #462 - Fix for error messaging when you attempt Merge accounts

raycarrick-ed:
without realising you are attemping to merge the same account.

The failure did not display a customised error message.

Changes:
- Added a check before merging that the Account id and Merge Account id
where different.
- If the check above failed you now get the error message:
"You attempted to merge 2 accounts with the same email address. Please merge with a different email address."
- Created a private method merge_accounts with merge step in merge()
method to avoid Rubocop offense "Metrics/AbcSize: Assignment Branch
Condition size for merge is too high."

Fixes # .

Changes proposed in this PR:
-


raycarrick-ed:
@johnpinto1 submitted the PR to DCC/roadmap/development resubmitting the PR here as I think that's where it should go.

briri (approved):
looks good @raycarrick-ed

PR #2974, opened 2021-07-19
Remove lowercase from application name

martaribeiro:
Changes proposed in this PR:
- Remove lowercase from application name (/roadmap/app/services/application_service.rb)

I couldn't find a reason for this. Is there one?


briri:
I did  quick scan of the can and can't see a reason for it to be using `downcase`. If an area of the code needs to do that then it should handle it, not the service.

briri:
Need to fix the two tests.

briri:
Ahh. Ok, the tests are showing that the V1 API deserialization service wants the name in lower case. I think just adding the `.downcase` to this line will fix those tests: https://github.com/DMPRoadmap/roadmap/blob/8e05bcc86aa1aa4ace57b1b4d93a55f281fc0896/app/services/api/v1/deserialization_service.rb#L64

PR #2981, opened 2021-07-26
Merging latest development into master

briri:
Hey @raycarrick-ed here's the PR to merge dev into master. See if you can resolve those other 2 open PRs though before this one.

Note that the JS spinner was added to the app/views/layouts/application.html.erb which is typically branded.

- New tr-TR translations
- Fix for github actions
- new JS spinner that displays during ajax calls
- bug fix for API v0
- bug fix so that contributors page now obeys the `restrict_orgs` flag
- refactor of `org.plans` so that we're allowing Rails' baseline functionality to work as expected
- update to paginable logic for the public plans and funder templates pages so that the sort/page/search params are carried along in the query string so that the URL can be copied and shared. (Not done on other paginable pages as they require login and are often contextual based on the user)
- bug fix for users who create an account instead of accepting the invitation
- bug fix for the collaborators table that prevents inactive roles from being displayed. 
- added new "Merge" tab to the Org page to allow Super Admins to merge Orgs
- bug fix for merging user accounts
- bug fix for the counting conditional questions as answered
- added a new "other" contributor type
- bug fix for displaying contributor/contact info in API v0 json


raycarrick-ed (approved):
As far as I can tell this looks good to go. I've had a brief skim through the changes and it all looks ok, i think. So let's try it and see.

Issue #2920, opened 2021-05-20
Add new field for Funding Status


mariapraetzellis:
We need a field that tracks the status of funding. 

- We can put this in the Project Details page
-  titled "Funding Status" 
- with the drop down options of: Granted, Planned, Rejected

briri:
Run through user group to see what to name/call the values perhaps "not funded" instead of "rejected"

raycarrick-ed:
see

https://github.com/DMPRoadmap/roadmap/issues/2922

which suggests the funding status may be more complex

Issue #2816, opened 2021-02-18
Remove mime type field from the Research Outputs tab


mariapraetzellis:
Following feedback from FAIR Island, we've decided that the mime type field was too detailed and would not be useful for researchers at the planning stage. So, we're removing this field from the Outputs tab.

Issue #2673, opened 2020-09-08
Add new field for "Field of Science" classification system


mariapraetzellis:
As recommended and implemented by DataCite (described here: https://blog.datacite.org/power-of-pids/) we should add a new field to define the "field of science" for which a DMP is being created. This will allow for more granular search queries and allow us to use filters for content search.

I think this would be best on the Project Details page but we can discuss.

The Field of Science and Technology classification is described here: https://www.oecd.org/science/inno/38235147.pdf

briri:
We have added a subset of this into the re3data repository search. It would be better to add it as described above to help pre-filter searches.

Issue #2432, opened 2020-03-16
Add ethical issues section to project details


briri:
The RDA common metadata standard for DMPs provides the following fields new fields:
```json
{ 
  "dmp": {
    "ethical_issues_exist": "yes",
    "ethical_issues_description": "We will collect personal information from human test subjects. This data should be anonymized prior to publication and preservation.",
    "ethical_issues_report": "https://dmproadmap.org/path/to/my/ethical_compliance_report.pdf"
  }
}
```

We need to determine how best to capture this information within a DMP. I can see 2 possible solutions:
- As fields on the Project Details page (see screenshot). 
  - The data in the common standard is at a high level so I think this may be a good place to record the information 
- As a question type/format that can be added to a template. 
  - The common standard only requires the `ethical_issues_exist` flag to be present, and the default if 'unknown', so any templates that do not use this new question type would still be compliant.
  - We could make use of the conditional question logic: user tics the `ethical_issues_exist` box and the description and report fields get displayed

<img width="610" alt="Screen Shot 2020-03-16 at 2 25 37 PM" src="https://user-images.githubusercontent.com/1204467/76801019-07aebe00-6792-11ea-9b0c-599d397162a9.png">

Regardless of which route we take, we will need to come up with appropriate labels and tooltip text that will help researchers understand what type of information they should provide.

briri:
will drop the new columns from the plans table. Placing this back into Pending was it will rely on the conditional question code.

PR #2986, opened 2021-07-30
New maDMP additions for the project details page

briri:
**maDMP features**
- Adds a "Funding status" selection box #2920
- Adds a "Research domain" selection box #2673
- Adds the "Ethical Issues" checkbox and corresponding "Description" and "Report URL" #2432

**Other features/fixes**
- Fixed bug that was preventing the `plan.funder_id` from being set to nil
- Added a new `app/javascript/src/utils/conditionalFields.js` file that allows you to define sets of fields that are hidden/shown. Currently using it to hides/show the Description and Report URL if the ethical issues checkbox is not checked and the Grant and Funding Status fields if no Funder is selected.
- Added `grant` getter and setter overrides to the Plan model to allow the grant to be loaded and set properly from incoming UI params.
- Added new config settings to allow the Research Domain and Ethical Issues to be enabled/disabled
- Update API V1 to:
  - Use the `plan.owner` as the data contact if available (otherwise the contributor who is the :data_curator)
  - Include the new `ethical_issues` fields
  - Include the `research_domain` as a :keyword
  - include the `plan.org` as a :dmproadmap_funded_affilaition to deal with scenarios where the `plan.org` does not match the orgs defined for the owner or contributors
  - include the user eneterd `plan.identifier` as :dmproadmap_funding_opportunity_number
- Added lots of tests

**DB Changes**
- Removed old `data_contact`, `principal_investigator`, `grant_number` and `funder_name` fields from the `plans` table since they have been replaced with associations to contributors, identifiers and orgs
- Removes the mime_types table which we decided not to use #2816 
- Add a Rake task `rails external_apis:add_field_of_science_to_research_domains` that populates the new `research_domains` table


raycarrick-ed (approved):
Looks fine. Like the fact that it can be switched on/off so we can make it live once it suits.

PR #2987, opened 2021-08-03
2 small tweaks to get fix to run on dcc live

raycarrick-ed:
Changes proposed in this PR:
- add question options to modifiable attrs
- protect against cases where a template doesn't have any question options by throwing and catching an exception.


briri (approved):
good catch. thanks @raycarrick-ed

PR #2990, opened 2021-08-04
Fix migration classname

raycarrick-ed:
Fixes bug in migration class name which cause db:migrate to fail with NameError looking for CreateResearchDomains.

Changes proposed in this PR:
- change class name from CreateFos to CreateResearchDomains


raycarrick-ed:
No - not master - developoment - AAAhhhhhh

PR #2991, opened 2021-08-04
change migration classname

raycarrick-ed:
Fixes bug in db:migrate caused by classname/filename mismatch.

Changes proposed in this PR:
- change classname to match filename


briri:
Ah thanks @raycarrick-ed we have been using 'Field of Science' designations and when I was creating the PR realized it should perhaps be more generic. Thought I had caught them all.

PR #2996, opened 2021-08-10
exclude private and test plans if read_all not set

raycarrick-ed:
Fixes https://github.com/DigitalCurationCentre/DMPonline-Service/issues/594

Changes proposed in this PR:
- there was already a config setting org_admins_read_all but it wasn't being used in the org_admin plans page
- this PR fixes that by removing private and test plans from the orgadmin view if org_admin_read_all is false.

We use this in our DMPTuuli instance.

### DIFF
--- a/app/models/org.rb
+++ b/app/models/org.rb
@@ -292,7 +292,14 @@ class Org < ApplicationRecord
   # This replaces the old plans method. We now use the native plans method and this.
   def org_admin_plans
     combined_plan_ids = (native_plan_ids + affiliated_plan_ids).flatten.uniq
-    Plan.includes(:template, :phases, :roles, :users).where(id: combined_plan_ids)
+    
+    if Rails.configuration.x.plans.org_admins_read_all
+      Plan.includes(:template, :phases, :roles, :users).where(id: combined_plan_ids)
+    else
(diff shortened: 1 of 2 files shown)


briri (approved):
This change looks good to me @raycarrick-ed.

Can you also check the logic in the API as well. I think they should probably also adhere to that config setting as well:
- For API V1: https://github.com/DMPRoadmap/roadmap/blob/241f537673f1fee602f1fb318f7a95b388b32909/app/policies/api/v1/plans_policy.rb#L32
- For API V0: https://github.com/DMPRoadmap/roadmap/blob/241f537673f1fee602f1fb318f7a95b388b32909/app/controllers/api/v0/plans_controller.rb#L71

pherterich:
@weiweishi @lagoan - highlighting this option to restrict visibility of plans for admins as I think you were interested in that and I didn't know we were working on this for some of our instances.

weiweishi:
> @weiweishi @lagoan - highlighting this option to restrict visibility of plans for admins as I think you were interested in that and I didn't know we were working on this for some of our instances.

Thank you @pherterich this is great to see! thanks for the update. I will let the team know.

PR #3015, opened 2021-08-19
Issue #561 - Fix for ensuring Json format rendering errors no longer

### LINKED ISSUE #561: Refactor CSS/JS for Download Plan page
Refer to the work done in #555 for tabs and #554 for forms

johnpinto1:
return the standard html error page.

Fixes #561

Changes in ApplicationController:
- so that rescue_from errors are rendered appropriately formatted messages for html and json with a private method render_respond_to_format_with_error_message()
- added a rescue_from StandardError with a method handle_server_method()
- updated existing user_not_authorized
- updated rescue_from ActiveRecord::RecordNotFound by removing
if Production condition
- added missing method render_not_found()


johnpinto1:
This PR was created to start a conversation on changes I made based on @briri's comments https://github.com/DigitalCurationCentre/DMPonline-Service/issues/561#issuecomment-872306836

Attached is image of an error with Json that previously returned html 
Selection_017

app/controllers/application_controller.rb
```diff
@@ -30,9 +32,13 @@ def current_org
 
   def user_not_authorized
     if user_signed_in?
-      redirect_to plans_url, alert: _("You are not authorized to perform this action.")
+      # redirect_to plans_url, alert: _("You are not authorized to perform this action.")
+      msg = _("You are not authorized to perform this action.")
+      render_respond_to_format_with_error_message(msg, plans_url, 404, nil)
```
briri, line 37: Should this be a 403 forbidden?

app/controllers/application_controller.rb
```diff
@@ -30,9 +32,13 @@ def current_org
...
     else
-      redirect_to root_url, alert: _("You need to sign in or sign up before continuing.")
+      # redirect_to root_url, alert: _("You need to sign in or sign up before continuing.")
+      msg = "You need to sign in or sign up before continuing."
+      render_respond_to_format_with_error_message(msg, root_url, 404, nil)
```
briri, line 41: Should this be a 401 - unauthorized?

briri (approved):
Thanks, this looks good @johnpinto1 

I just question the use of 404 in the user_not_authorized section. @raycarrick-ed what do you think?

It also looks like a test broke due to the change. It may be that the test just needs to be updated to accommodate this change in error handling rather than there being something wrong with your changes.

app/controllers/application_controller.rb
raycarrick-ed: Agreed.
404 = resource not found which isn't the case here.
if repeating the request after authenticating would work then should be a 401, I think.
if even after authenticating you still wouldn't be allowed the it's 403
That's my understanding anyway.

johnpinto1:
@briri @raycarrick-ed Updated code as suggested. Thanks for feedback.

briri:
@johnpinto1 thanks for making those changes.

It looks like the test is failing because it was expecting to be redirected to the "My dashboard" page when the user is unauthorized. I think you can just update the test itself to expect `root_path` instead of `plans_path`. The HomeController (the one that handles `root_path`) checks to see if the user is signed in. If they are it redirects them to `plans_path`. The RSpec controller test though is just looking at the response from that controller so it would never hit that redirect on the HomeController.

johnpinto1:
Thanks @briri my bad. As usual I never looked at tests, assumed error were db related. Will update tomorrow.

johnpinto1:
@briri & @raycarrick-ed  To fix tests I had to make the following changes (not sure about whether removing "format: :js" is acceptable:

      In spec/controllers/super_admin/orgs_controller_spec.rb failing tests
       (1) merge_analyze: test "fails if user is not a super admin"
       (2) merge_commit: test "fails if user is not a super admin"
       - changed redirect_to(plans_path) -> redirect_to(root_path)
       - removed "format: :js" because of ActionController::UnknownFormatError
           post :merge_analyze, params: @params. format: :js
           and
           post :merge_commit, params: @params. format: :js

briri:
looks good. Thanks @johnpinto1!

PR #3016, opened 2021-08-19
Db cleanup

### LINKED ISSUE #2818: DB Cleanup
While writing up some [DB schema info](https://github.com/DMPRoadmap/roadmap/wiki/DB-Schema) I realized we have some cleanup to do:

Be sure to update the wiki doc after these changes have been made!

Remove the following columns (once confirmed that they are no longer used):
- answers.label_id
- identifier_schemes.logo_url - logos are stored in the assets/images dir
- **RENAME** orgs.feedback_email_msg to orgs.feedback_msg - Its no longer emailed!
- orgs.feedback_email_subject - This is no longer an email the message is displayed on the page
- orgs.is_other - Replaced by the managed flag and ROR API functionality
- orgs.sort_name
- plans.data_contact These are now stored in `contributors`
- plans.data_contact_phone 
- plans.data_contact_email
- plans.grant_number` - This is now recorded via the `grant_id` fkey
- plans.funder_name - This is now a recorded via the `funder_id` fkey
- plans.principal_investigator - These are now stored in `contributors`
- plans.principal_investigator_identifier
- plans.principal_investigator_phone
- plans.principal_investigator_email
- research_outputs.coverage_region - Not going to be necessary
- research_outputs.coverage_start - Not going to be necessary
- research_outputs.coverage_end - Not going to be necessary
- research_outputs.mandatory_attribution - Not going to be necessary
- research_outputs.mime_type_id - Not going to be necessary
- users.other_organization - Replaced by actual Orgs

Drop the following tables:
- mime_types - It was meant for the ResearchOutputs functionality but is no longer necessary
- org_identifiers - Those values are now stored in the `identifiers` polymorphic table
- user_identifiers - Those values are now stored in the `identifiers` polymorphic table

Things to do after some research and refactor:
- See if we need the `settings` table. It stores the font and margin settings a user specifies when exporting a plan. That info is carried through in the controller though and available there. There doesn't seem to be a need to store it.

briri:
Fixes #2818

Changes proposed in this PR:
- See ticket for specific columns and tables that were removed.
- Updated code to remove references

Exceptions:
- Did not remove the `settings` table. We fixed a bug with the `exported_plans` table so I think it is still used. Not sure why we would want to retain the selected font and margins but 🤷🏻‍♂️ 
-  See the [maDMP PR migration](https://github.com/DMPRoadmap/roadmap/blob/madmp-research-outputs/db/migrate/20210729204611_madmp_cleanup.rb) for the removal of columns from `research_outputs` and removing the `mime_types` table


raycarrick-ed:
@briri I think we need to discuss this. Maybe on Monday?

Most of this looks fine from what I can tell but I am concerned about the use of "is_other" and "other_organisation" because I think we may be using these differently from you. At least, specifically:
Org.where(is_other: true)
and
Org.where(managed: false)
are not the same things.
We have a single "Other" org still which is managed and we have users who have org = other and then some value in the other_organisation field.

Let's talk about it Monday. Cheers.

briri:
ok @raycarrick-ed if those two fields are still used heavily in your installations then I can revert those. The switch over from is_other to managed was part of the move to use the ROR API which I know has been problematic.

Happy to chat about it more on Monday.

PR #3021, opened 2021-09-02
Remove p tag from help page

martaribeiro:
Fixes # .
The 'p' tag was being closed inside the rails. This was causing problems on the html display
Changes proposed in this PR:
- remove the p tag from inside rails and add it to the out side as it should be


martaribeiro:
Please ignore. There are old merges  attached with this PR. I'll redo it

PR #3022, opened 2021-09-02
Remove p tag from help page text changes

martaribeiro:
Fixes # .
The 'p' tag was being closed inside the rails. This was causing problems on the html display
Changes proposed in this PR:

Remove the p tag from inside rails and add it to the out side as it should be.


briri (approved):
looks good. thanks @martaribeiro

PR #3026, opened 2021-09-08
Issue2989

### LINKED ISSUE #2989: Guidance not displaying correctly
When a user from an organization that has customized guidance creates a plan, the guidance that appears within the plan should include the organizational guidance AND the generic DMPOnline/DMPTool guidance. Currently, only the institutional guidance is displaying. @briri checked, and this is due to a bug in Roadmap.

briri:
Fixes #2989.

Changes proposed in this PR:
- Prevent the project details page from displaying all guidance groups associated with an Org whose org_type is 'organization'.

It Iooks like I added this line over a year ago and don't recall why. Updated it to only use the default org's guidance.

### DIFF
--- a/app/controllers/plans_controller.rb
+++ b/app/controllers/plans_controller.rb
@@ -192,7 +192,7 @@ class PlansController < ApplicationController
       @important_ggs << [current_user.org, @all_ggs_grouped_by_org[current_user.org]]
     end
     @all_ggs_grouped_by_org.each do |org, ggs|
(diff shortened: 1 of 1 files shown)


raycarrick-ed (approved):
Sounds reasonable.

nicolasfranck:
why is something like `Org.default_orgs` equal to `Org.where(abbreviation: Rails.configuration.x.organisation.abbreviation)`? It is not documented
in the configuration and it surely has some unintended side effects?

PR #3030, opened 2021-09-19
Remove vulnerability xss hack

martaribeiro:
Fixes # https://github.com/DigitalCurationCentre/DMPonline-Service/issues/603.

Changes proposed in this PR:
Callback was added to user and plan models. 'before_save' was added with a function sanitise a number of field before data was saved into the db
-


briri (changes requested):
Thanks for tackling this one @martaribeiro.

I think it would good to move this into the base application_record.rb file that all of the models inherit from. Something like:
```ruby
class ApplicationRecord < ActiveRecord::Base

   #  other existing methods

   def sanitize_fields(*attrs)
     attrs.each do |attr| 
       self.send("#{attr.to_s}=", ActionController::Base.helpers.sanitize(self.send(attr)))
     end
   end
```
Then, in the models you can do something like:
`before_save sanitize_fields(:title, :funder_name, :grant_number, :identifier, :description)`

briri (approved):

briri:
was out of sync with development so rebased via GitHub

Karimblack:
">] ">]

PR #3031, opened 2021-09-21
Bug fix for contributors new.html.erb

raycarrick-ed:
fixes: https://github.com/DigitalCurationCentre/DMPonline-Service/issues/622
Org select not working when creating a contributor.
view refernces @orgs when it should be @all_orgs


briri:
looks good @raycarrick-ed

PR #3032, opened 2021-09-21
Tinymce label fix

briri:
While working on something else, I noticed that the existing Tinymce JS was connecting the editor to the incorrect labels. This can cause problems for screen reader accessibility.

The old code was changing the `for` attribute of ALL labels on the form. This change ensures that it updates only the related label.


raycarrick-ed (approved):
Looks good to me. Not an easy one to catch.

PR #3036, opened 2021-09-27
added handler for Pundit errors on ApplicationController

briri:
Just building off of the pattern @johnpinto1 established for us for handling errors.

This ensures that any Pundit authorization errors result in an Unauthorized error message and brings the user back to their Dashboard or the homepage if they're not logged in


raycarrick-ed (approved):
Looks an improvement.

raycarrick-ed:
The test failures look like a consequence of the exception handling change so just eed changed to be new results. I think.

briri:
fixed up the tests

PR #3044, opened 2021-10-01
New maDMP feature - Related Identifiers

### LINKED ISSUE #2815: Add a related identifiers field to project details page
Following feedback from the FAIR Island Project testing, we need to add a new "Related Identifiers" field to the project details tab. This will allow users to connect specific outputs to the record and gives us another route to connect planned outputs to their completed/published versions

- this should apply to all different output types
- will require a tool tip to define the field 
- we will need talk w/ DataCite about how we can make assertions to the identifiers included here

briri:
Fixes #2815.

Changes proposed in this PR:
- Add new `related_identifiers` table and corresponding model and tests
- Add link to [uc3-citation gem](https://github.com/CDLUC3/uc3-citation)
- Add 2 new flags to the config/initializers/_dmproadmap.yml file to:
  - enable related identifiers - determine whether or not the user can add related identifiers on the Project Details page
  - enable citation lookups - whether or not the system will use the uc3-citation gem to fetch citations for DOIs when a related identifier is saved
  - updated read-only view of the Project Details page to display related identifiers (if applicable)
  - updated pdf download options to allow user to select whether or not to display related works as an appendix of the PDF (if applicable)
  - updated API V1 to display related identifiers (if applicable)


briri:
NOTE THAT THIS PR IS LIKELY OUT OF DATE WITH DMPTOOL

The DMPTool has had many updates since this PR was first created so an eval of these changes should be done to ensure that any bug fixes are included

PR #3047, opened 2021-10-06
IssueID #629 - Fix to allow API Access tab visible on the users's "Edit 

johnpinto1:
Fixes #629 .

Changes proposed in this PR:
-  Added **can_use_a**pi? to the User model method **can_org_admin.**
    
    (Noticed, that if only **can_use_api?** is set, then the user sees the **Admin**
    **Plans** and **Usage**. Advice needed if this is ok, or a different check is required around **"API Access**" fragment in **"Edit Profile"**)

Selection_057


briri:
I would think that a user who only has `can_use_api` access should NOT be able to see those other Admin menu items. 

Yes, the rubocop errors are unrelated to your changes. We are in the process of upgrading to the latest version.

PR #3060, opened 2021-11-03
Issue #DCC649 - Fix for the contributor to fail to save on Contributor

johnpinto1:
page if not required Affiliation field not set.

Fix for DCC bug https://github.com/DigitalCurationCentre/DMPonline-Service/issues/649

Changes:
 - In ContributorsController an if condition has been changed in
process_org() method
Replaced  'return nil if org.blank? && !allow' by
 'return hash if org.blank? && !allow' where 'hash remove_org_selection_params(params_in: hash)'

Based on @briri  comment https://github.com/DigitalCurationCentre/DMPonline-Service/issues/649
I had to remove org from hash. Also combining the returns in process_org() defied my efforts so have

   `
    hash = remove_org_selection_params(params_in: hash)

    return hash if org.blank? && !allow
    return hash unless org.present?
`


briri (approved):
looks good John. Thanks!

Issue #2983, opened 2021-07-28
possible to add contributor without a name - even though it appears as compulsory field


magdalenadrafiova:
@raycarrick @briri 
I am testing on DMPonline-dev the latest 3.3 release. Not sure 
The contributor name appears as compulsory field but when I left it empty it was alright to save: 
<img width="989" alt="Screenshot 2021-07-28 at 14 53 06" src="https://user-images.githubusercontent.com/47178106/127334418-123747a1-8956-48e2-b9fa-f57bd2a58579.png">

See contributor without a name: 
<img width="1009" alt="Screenshot 2021-07-28 at 14 53 42" src="https://user-images.githubusercontent.com/47178106/127334501-95de505f-ee37-4922-bf9e-94a602975c31.png">

briri:
The logic in the code requires one or the other (name or email). Should we actually require both or is one more important?

Whoever addresses this ticket should also remove the phone number per #2941

johnpinto1:
Will look at https://github.com/DMPRoadmap/roadmap/issues/2941 next. Have a PR see above.

magdalenadrafiova:
@johnpinto1 both still appears as compulsory fields on dev follow up with @dsisu once fixed so she can re-test and close 
<img width="853" alt="Screenshot 2022-03-01 at 14 54 45" src="https://user-images.githubusercontent.com/47178106/156191897-e07bdbfb-63d1-4b47-8f3d-1467224d4fb5.png">

johnpinto1:
The view is displaying wrong info. The code saves either name or email, but view shows incorrectly that both name and email are required.

johnpinto1:
Trivial fix done and PR created https://github.com/DMPRoadmap/roadmap/pull/3120
![Selection_058](https://user-images.githubusercontent.com/8876215/156759056-e8cb2eef-6367-4af7-9271-290fb374c8c9.png)

pherterich:
Just catching up with things as part of trying to understand what the latest release covered: how is it obvious to a user in this new version that either name or email needs to be provided? Do we provide a meaningful error message if it isn't? Should we add a sentence saying that one of the two fields needs to be filled?

raycarrick-ed:
This is a good point. @johnpinto how easy would it be to add a message to that effect? Fill in at least one of these fields?

johnpinto1:
@raycarrick-ed & @pherterich It should be trivial. Famous last words. Its just adding an text.

[...]

PR #3062, opened 2021-11-04
Issue#dcc2983 - Fix for preventing adding a Contributor via Contributor

johnpinto1:
page if both Name and Email are not present.

Fixes DCC bug https://github.com/DMPRoadmap/roadmap/issues/2983
spec/services/api/v1/contextual_error_service_spec.rb
Changes:
  - Changes in Contributor model:
       - Renamed validation method name_or_email_presence() -> name_and_email_presence()
       - Updated conditions in name_and_email_presence() method to return errors if Name or Email missing.
  - Updated tests in spec/models/contributor_spec.rb and spec/services/api/v1/contextual_error_service_spec.r to reflect change in Contributor model.

@briri Not sure if this conditionality is different for UC3.
(Away for a week so won't pick up until Nov 15th.)

Scrrenshots of different error scenarios:
Selection_031
Selection_030
Selection_029

### DIFF
--- a/app/models/contributor.rb
+++ b/app/models/contributor.rb
@@ -55,7 +55,7 @@ class Contributor < ApplicationRecord
   validates :roles, numericality: { greater_than: 0,
                                     message: _("You must specify at least one role.") }
 
-  validate :name_or_email_presence
+  validate :name_and_email_presence
 
   ONTOLOGY_NAME = "CRediT - Contributor Roles Taxonomy"
   ONTOLOGY_LANDING_PAGE = "https://credit.niso.org/"
@@ -130,11 +130,11 @@ class Contributor < ApplicationRecord
 
   private
 
-  def name_or_email_presence
-    return true unless name.blank? && email.blank?
+  def name_and_email_presence
+    return true unless name.blank? || email.blank?
 
-    errors.add(:name, _("can't be blank if no email is provided"))
-    errors.add(:email, _("can't be blank if no name is provided"))
+    errors.add(:name, _("can't be blank.")) if name.blank?
+    errors.add(:email, _("can't be blank.")) if email.blank?
   end
 
 end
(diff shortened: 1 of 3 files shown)


johnpinto1:
@briri & @raycarrick-ed  Just saw test breaks not sure in **spec/services/api/v1/contextual_error_service_spec.rb**

1) Api::V1::ContextualErrorService contextualize(errors:) returns errors if a Contributor is invalid
     Failure/Error: expect(result.first.include?(" can't be blank if no ")).to eql(true)

       expected: true
            got: false

       (compared using eql?)

       Diff:
       @@ -1 +1 @@
       -true
       +false
     # ./spec/services/api/v1/contextual_error_service_spec.rb:83:in `block (3 levels) in <top (required)>'


I could fix it by  changing expect to false in test below **expect(result.first.include?(" can't be blank if no ")).to eql(true)** by changing  eql(false). Not sure if this has other implications

  **it "returns errors if a Contributor is invalid"**
      @plan.contributors.first.name = nil
      @plan.contributors.first.email = nil
      @plan.valid?
      result = described_class.contextualize(errors: @plan.errors)
      expect(result.length).to eql(2)
      expect(result.first.start_with?("Contact/Contributor ")).to eql(true)
      **expect(result.first.include?(" can't be blank if no ")).to eql(true)**  **<--- BROKEN**
    end
Will look to fix after I get back.

johnpinto1:
Ignore my previous comment on tests. I deleted comment and fixed broken test.

raycarrick-ed (approved):
Looks fine to me.

briri:
Yes, this will be problematic for us as we have many records where email was not provided. 

We only have a handful of records with a nil name, so I can manually address that one fairly easily. Maybe set this as an additional config flag like: `x.application.require_contributor_emails = false`

johnpinto1:
@raycarrick-ed @briri  Will close issue and re-open after finding a way of reconciling with @briri comment.

PR #3065, opened 2021-11-16
protect against nil owner after user merge

raycarrick-ed:
Fixes https://github.com/DigitalCurationCentre/DMPonline-Service/issues/656 .

Changes proposed in this PR:
- after a user merge in some edge cases you can end up with a nil plan owner which causes an exception. This protects it.


briri (approved):
looks good @raycarrick-ed

PR #3071, opened 2021-11-23
Issue#dcc2983 - Fix for preventing adding a Contributor via Contributor

johnpinto1:
form for Plan based on whether email and/or name are required.

Fix for DCC bug https://github.com/DMPRoadmap/roadmap/issues/2983

Changes:
 - Added two new properties to config/initializers/_dmproadmap.rb
   (both devault to false):
   config.x.application.require_contributor_name = false
   config.x.application.require_contributor_email = false
 -In Contributor model the private method name_or_email_presence()
  has been updated to take account of the above mentioned property
settings.

Fixes # .

Changes proposed in this PR:
-

### DIFF
--- a/app/controllers/paginable/plans_controller.rb
+++ b/app/controllers/paginable/plans_controller.rb
@@ -58,16 +58,17 @@ module Paginable
     end
     # rubocop:enable Metrics/AbcSize
 
-    # GET /paginable/plans/org_admin/:page
-    def org_admin_other_user
+    # GET /paginable/users/:id/plans
+    def index
       @user = User.find(params[:id])
       authorize @user
       raise Pundit::NotAuthorizedError unless current_user.present? && current_user.can_org_admin? && @user.present?
 
       paginable_renderise(
-        partial: 'org_admin_other_user',
+        partial: 'index',
         scope: Plan.active(@user),
-        query_params: { sort_field: 'plans.updated_at', sort_direction: :desc }
+        query_params: { sort_field: 'plans.updated_at', sort_direction: :desc },
+        format: :json
       )
     end
   end
--- a/app/controllers/super_admin/orgs_controller.rb
+++ b/app/controllers/super_admin/orgs_controller.rb
@@ -153,7 +153,8 @@ module SuperAdmin
       params.require(:org).permit(:name, :abbreviation, :logo, :managed,
                                   :contact_email, :contact_name,
                                   :remove_logo, :feedback_enabled, :feedback_msg,
-                                  :org_id, :org_name, :org_crosswalk)
+                                  :org_id, :org_name, :org_crosswalk,
+                                  :funder, :institution, :organisation)
     end
 
     def merge_params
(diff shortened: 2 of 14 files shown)


johnpinto1:
@briri & @raycarrick-ed  This is attemot to address @briri comment https://github.com/DMPRoadmap/roadmap/pull/3062#issuecomment-971766758
The default leaves the current contributor form validation unchanged.
@raycarrick-ed DCC will need to set properties below to address dcc bug ihttps://github.com/DMPRoadmap/roadmap/issues/2983
```
# Setting require contributor requirement of contributor name and email
      config.x.application.require_contributor_name = true
      config.x.application.require_contributor_email = true
```

briri (approved):
thanks @johnpinto1!

johnpinto1:
@briri I like you to double-check the logic as it took me trial and error to get it to what seemed like "work". Thanks.

briri:
will do @johnpinto1 and will report back here

raycarrick-ed:
Looks good overall.
Agree with @briri 's comments above.
I **think** the code can be simplified where you have:

Rails.configuration.x.application.foobar.present? && Rails.configuration.x.application.foobar

I think that is exactly the same as just

Rails.configuration.x.application.foobar

if it's not present it will return nil which is equivalent to false. That would cut out quite a few lines of code.

johnpinto1:
@raycarrick-ed Adding the properties now seems a good idea. I have done something similar in another project to prevent forgetting about it in future.

johnpinto1:
@raycarrick-ed & @briri I will look at it again. I had issues with Rails.configuration.x.application.foobar
when property not in config. I play around to establish if I was seeing another issue.

johnpinto1:
@briri & @raycarrick-ed  I have updated code based on your comments and my experimenting with nil (when a property is missing by usiing ! or !! (if I want current value or nil -> false)
irb(main):001:0> nil
=> nil
irb(main):002:0> !nil
=> true
irb(main):003:0> nil == true
=> false
Method now 
  
  ```
def name_or_email_presence
    return true unless (name.blank? && email.blank?) ||
                        (!!Rails.configuration.x.application.require_contributor_name &&
                        name.blank?) ||
                        (!!Rails.configuration.x.application.require_contributor_email &&
                        email.blank?)

    # Error messages for case where name.blank? and email.blank? and
    # properties Rails.configuration.x.application.require_contributor_name and
    # Rails.configuration.x.application.require_contributor_email both not set (i.e, nil) or false.
    if (name.blank? && email.blank?) &&
       !Rails.configuration.x.application.require_contributor_name &&
       !Rails.configuration.x.application.require_contributor_email
      errors.add(:name, _("can't be blank if no email is provided."))
      errors.add(:email, _("can't be blank if no name is provided."))
    else

      # Error messages cases where enither ame.blank? or email.blank?
      if name.blank? && !!Rails.configuration.x.application.require_contributor_name
        errors.add(:name, _("can't be blank."))
      end

      if email.blank? && !!Rails.configuration.x.application.require_contributor_email
        errors.add(:email, _("can't be blank."))
      end
    end
  end
 
```

PR #3077, opened 2021-12-09
use current user rather than form user id which can be spoofed

raycarrick-ed:
Fixes #[603](https://github.com/DigitalCurationCentre/DMPonline-Service/issues/603) .

Changes proposed in this PR:
- use current user id for comments rather than the user id on the form
- the above ticket listed a couple of "security issues" which came from Erasmus. This is the second.
- if someone edits the html and replaces the userid with one for someone who has permissions to comment they can spoof the user. Only works with people who have permission to comment so not a big deal but, still, might as well plug it.


briri (approved):
Looks good @raycarrick-ed. Thanks for patching this

Issue #3064, opened 2021-11-12
Configuration option restrict_orgs does not fully restrict assignable organisations/funders in plan update controller


nicolasfranck:
Please complete the following fields as applicable:

**What version of the DMPRoadmap code are you running? (e.g. v2.2.0)**

`3.0.4`

**Expected behaviour:**

If you set the configuration option `config.x.application.restrict_orgs = true`,
one should not be able to create an unmanaged organisation, not?

**Actual behaviour:**

[This line](https://github.com/DMPRoadmap/roadmap/blob/master/app/controllers/plans_controller.rb#L270)
allows you to create a new (unmanaged) organisation.

Now we have a long list of unmanaged organisations of type "organisation",
because people tend to fill in a custom name in the field "funder" on the
"project details page". Afterwards they cannot find their name in the lookup
because that custom organisation was not marked as funder (and also because
it is unmanaged).

Not sure if that is intended behaviour?

These unmanaged organisations tend to pollute the list of organisations.
Is it worthwhile to add an organisation that is only used in one plan?

briri:
Hi @nicolasfranck we were just discussing this issue last week. You are correct, the code on this page should be using the 'restrict_orgs' flag. I think we just missed it when implementing.

We have not yet issued a patch for this. 

If you need to clean up any orgs that were created due to this issue, you can run a query like this to identify any Orgs created via the Funder field on the Project Details page 
```SQL
SELECT * FROM orgs 
WHERE managed = false AND id IN (SELECT DISTINCT funder_id FROM plans) 
  AND id NOT IN (SELECT DISTINCT org_id FROM plans WHERE org_id IS NOT NULL) 
  AND id NOT IN (SELECT DISTINCT org_id FROM users WHERE org_id IS NOT NULL) 
  AND id NOT IN (SELECT DISTINCT org_id FROM contributors WHERE org_id IS NOT NULL)
  AND id NOT IN (SELECT DISTINCT org_id FROM templates WHERE org_id IS NOT NULL) ;
```

raycarrick-ed:
this is known and is unintended behaviour. In dmponline we are patching in the meantime to stop it. @briri is currently working on a more comprehensive solution to this issue as there are multiple options for orgs and funders related to user and plan creation.

raycarrick-ed:
Possibly connected to: https://github.com/DigitalCurationCentre/DMPonline-Service/issues/650

mariapraetzellis:
This has been fixed and in latest release.

PR #650, opened 2017-09-18
Minor bugfix

briri:
Added some missing close parens in the password reset page (1st file listed here).

The edit profile and create plan commits are already a part of the CDL-MVP branch on DMPRoadmap. Wondering why they appear below as changes. Could possibly be due to performing a 'Squash and Merge' here in Github when pulling the latest from DMPRoadmap -> CDLUC3

PR #3078, opened 2021-12-09
get org creation to follow restrict_orgs

raycarrick-ed:
Fixes [#3064](https://github.com/DMPRoadmap/roadmap/issues/3064) and [#650](https://github.com/DigitalCurationCentre/DMPonline-Service/issues/650) .

Changes proposed in this PR:
- change default allow creation to trigger off the config setting restrict_orgs
- remove various explicit allow_creates so that they just go to the default


briri (approved):
👍🏻 makes sense to update the default on the concern

PR #3085, opened 2022-01-10
Issue DCC-674: Fix for broken CSV Download noticed because a nil

johnpinto1:
exception occuring for an answer with question options selected but text
nil. Link to issue
https://github.com/DigitalCurationCentre/DMPonline-Service/issues/674

Change to app/models/concerns/exportable_plan.rb,
we replaced line
   answer_text += answer.text if answer.answered?
with answer.text present check
   answer_text += answer.text if answer.answered? && answer.text.present?


raycarrick-ed:
Happy to have a counter argument put but isn't this the wrong answer?
An answer is answered if it is option based and there is an option selected or
if it is text based and there's some text there.
So rather than adding "answer.text.present?" shouldn't he change be to answer#answered? 
if it's text based and there's no text, it isn't answered.
No?

johnpinto1:
> Happy to have a counter argument put but isn't this the wrong answer? An answer is answered if it is option based and there is an option selected or if it is text based and there's some text there. So rather than adding "answer.text.present?" shouldn't he change be to answer#answered? if it's text based and there's no text, it isn't answered. No?

I did wonder why the answer_text was outside the previous if for answer_options. So I have changed it as suggested by @raycarrick-ed 

  ```
if answer.present?
        if answer.question_options.any?
          answer_text += answer.question_options.pluck(:text).join(", ")
        elsif answer.answered?
          answer_text += answer.text
        end  
      elsif unanswered
        answer_text += _("Not Answered")
      end
```

briri:
I believe the answer text was outside of the if else block because some option based questions allow for an 'additional comments' text area and those comments get stuffed into the answer's text field.

I haven't verified this though, so please check. You'd have to find a question (or add your own)  on a template and make sure the 'allow comments' checkbox is ticked for the option based question. Then publish the changes and make a new plan from that template to test it out.

johnpinto1:
@briri and @raycarrick-ed Then it safer to change it to how I had it, So @raycarrick-ed I will change it tomorrow.

johnpinto1:
Reverted to change to take account of case suggested by @briri above@

Replaced line
   answer_text += answer.text if answer.answered?
with answer.text present check
   answer_text += answer.text if answer.answered? && answer.text.present?

nicolasfranck:
Just a thought, but these rubocop fixes generate a lot "clutter" in the log. It is hard to see in this pull request what was changed "code wise".

johnpinto1:
@nicolasfranck Thanks. Also for some ideas for fixes you suggested and I have used in past.

PR #3091, opened 2022-01-13
DCC Issue 679 - Fix for issue that Super Admins unable to create new

johnpinto1:
Orgs.

Fix for DCC issue https://github.com/DigitalCurationCentre/DMPonline-Service/issues/679

Changes:
- in **app/controllers/super_admin/orgs_controller.rb** added missing params
to **orgs_params** method
     **:funder, :institution, :organisation**
- in **app/models/org.rb** removed (as it is never set on Org creation)
      **validates :abbreviation, presence: { message: PRESENCE_MESSAGE }**
- in **app/views/orgs/_profile_form.html.erb** removed :abbreviation required
      **"aria-required": true**
- in **app/views/shared/org_selectors/_external_only.html.erb** renamed wrongly named text field **:org_name** to **:name**
      **<%= form.label :name, label %>
      <%= form.text_field :name, class: "form-control autocomplete",**

### DIFF
--- a/app/controllers/super_admin/orgs_controller.rb
+++ b/app/controllers/super_admin/orgs_controller.rb
@@ -153,7 +153,8 @@ module SuperAdmin
       params.require(:org).permit(:name, :abbreviation, :logo, :managed,
                                   :contact_email, :contact_name,
                                   :remove_logo, :feedback_enabled, :feedback_msg,
-                                  :org_id, :org_name, :org_crosswalk)
+                                  :org_id, :org_name, :org_crosswalk,
+                                  :funder, :institution, :organisation)
     end
 
     def merge_params
--- a/app/models/org.rb
+++ b/app/models/org.rb
@@ -94,8 +94,6 @@ class Org < ApplicationRecord
   validates :name, presence: { message: PRESENCE_MESSAGE },
                    uniqueness: { message: UNIQUENESS_MESSAGE }
 
-  validates :abbreviation, presence: { message: PRESENCE_MESSAGE }
-
   validates :is_other, inclusion: { in: BOOLEAN_VALUES,
                                     message: PRESENCE_MESSAGE }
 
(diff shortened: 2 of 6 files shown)


raycarrick-ed:
You have changes to _external_only which are not reflected in _combined or _local_only.
Is that right?
Without thinking about it a lot so I might be wrong, I'd have thought that they are interchangeable and so need to be equivalent.
So all returning org_name or all returning name.
Is that not the case?

johnpinto1:
@raycarrick-ed will check based on above comment.

johnpinto1:
@raycarrick-ed  looks like change in app/views/shared/org_selectors/_external_only.html.erb is not quite right. As lookup of orgs fail. Need to look into it as _combined or _local_only versions work correctly.

johnpinto1:
@raycarrick-ed Switching RoR on affects following:
**app/views/devise/registrations/_personal_details.html.erb**
82:        <%= render partial: "external_identifier",

**app/views/orgs/_profile_form.html.erb**
12:    <%= render partial: "orgs/external_identifiers",
158:    <%= render partial: "orgs/external_identifiers",

The **app/views/orgs/_profile_form.html.erb** is the used in:

**app/views/super_admin/orgs/new.html.erb**
26:              <%= render partial: 'orgs/profile_form',

**app/views/orgs/_profile_form.html.erb**
9:                                    id: "edit_org_profile_form" } ) do |f| %>

**app/views/orgs/admin_edit.html.erb**
42:              <%= render partial: 'orgs/profile_form',

**app/views/devise/registrations/_personal_details.html.erb** is used by:
**app/views/devise/registrations/edit.html.erb**
35:              <%= render partial: 'devise/registrations/personal_details', locals: {orgs: @orgs, org_partial: @org_partial} %>

briri (approved):
looks ok to me @johnpinto1 that 'external_only' partial is only used on that one page so changing the param names will not negatively impact other portions of the site.

Issue #693, opened 2017-10-11
Add search box to Public DMPs page


stephaniesimms:
Add search box to Public DMPs page 
Use same box as My Dashboard, DMP Templates, other lists

briri:
The table has a search box, but it only becomes available if there are more than 10 public plans.
The 10 row limit is set globally. We can reduce it to 5 (or 1 even) if we want it to always be visible, but it would end up working that way for all tables across the site.

@stephaniesimms let me know if the 10 row limit is alright

stephaniesimms:
thanks for the reminder. 10 rows still sounds sensible to me. closing issue.

PR #3105, opened 2022-02-10
DCC Bug #693 - Fix for pagination of plans retreived using the REST API V0

johnpinto1:
Fix for DCC Issue https://github.com/DigitalCurationCentre/DMPonline-Service/issues/693

Changes:
- replaced wrongly named param :page_size with :per_page that is used in
pagination in app/controllers/api/v0/base_controller.rb
- added a before_action :page_params to app/controllers/api/v0/plans_controller.rb
- added code to retreive :per_page param to scope in
app/controllers/concerns/paginable.rb.


johnpinto1:
@briri  & @raycarrick-ed The test error is (nothing to do with changes):

Failed examples:
rspec ./spec/features/locales_spec.rb:51 # Locales when new locale has no region user changes their locale

briri:
can confirm the error isn't related to the changes @johnpinto1 made.

I fixed it by removing the trailing 's' on the token in that string. I can't wrap my head around why it wants a trailing 's' in some places and not others. The examples in thee rubocop docs appear to be contradictory: https://www.rubydoc.info/github/bbatsov/RuboCop/RuboCop/Cop/Style/FormatStringToken

PR #3119, opened 2022-03-03
DCC bug 697 - Fix for a roadmap-wide bug where the following were

johnpinto1:
failing in the REST API V1 call "GET 0/api/v1/plans"
for following reasons:
(1) Plans policy initialize method was broken
(2) Each plan in the json array was using the email for the user name.

Changes made:
(1) The initialize method in /policies/api/v1/plans super call requires
no arguments, it was taking the ApiClient as an argument.
(2) In /views/api/v1/contributors/_show.json.jbuilder we changed
json.name as follows as we were previously using the 'User' rather than
'Contributor' object
 json.name contributor.is_a?(User) ? contributor.name(false) : contributor.name

Changes proposed in this PR:
-


johnpinto1:
@briri & @raycarrick-ed  I don't think Tests error are due to changes I made.

johnpinto1:
@briri & @raycarrick-ed  I have added a commit that comments out the Karma tests as suggested.

PR #3128, opened 2022-03-14
Revert string format token changes made for Rubocop

briri:
We switched the way we handle string formatting tokens when we upgraded Rubocop. Rubocop's new default is to use `%<variable>s` instead of `%{variable}`.

This has caused quite a bit of confusion for our translators, and I discovered today that Translation.io is not properly auto-translating these. It takes `%<variable>s some text` and translates it to `%<variable> s un peu de texte`. Note the space between the closing bracket '>' and the 's' character.

This update reverts the changes we made to use `%<variable>s` and updates the `.rubocop.yml` file to enforce use of the `template` style, `%{variable}` going forward.

Once merged, translations will need to be resynced!

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -147,6 +147,9 @@ Style/DocumentDynamicEvalDefinition: # new in 1.1
   Enabled: true
 Style/EndlessMethod: # new in 1.8
   Enabled: true
+Style/FormatStringToken:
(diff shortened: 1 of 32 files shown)


briri:
FYI @benjaminfaure and @hbayindir

raycarrick-ed:
Changes all look fine.

PR #3134, opened 2022-03-29
add validation to new madmp models

nicolasfranck:
Adds validation to the new madmp models `Repository`, `License` and `MetadataStandard`


briri:
Thanks for adding these validations @nicolasfranck

PR #3137, opened 2022-03-31
DCC bug 711 - Fix for the Rest API call POST /api/v1/plans for Plan

johnpinto1:
creation. The previous call based on instructions in wiki was always
defaulting to using the default template regardless of template id sent.

Fix for DCC bug 711 https://github.com/DigitalCurationCentre/DMPonline-Service/issues/711

The API Documentation V1 - Create Plan page has been changed
"dmptool" replaced by "dmproadmap" in fragment of Json sent for plan
creation.

Changes:
- We use :dmproadmap which is the symbol of the dmproadmap json object
  nested in extensions which is the container for the json template object, etc.
- In app/services/api/v1/deserialization_service.rb the method
app_extensions() has been changed to use the symbol :dmproadmap.
- In app/views/api/v1/plans/_show.json.jbuilder the block that builds
the template json object now uses the symbol :dmproadmap
- fixed tests in spec/services/api/v1/contextual_error_service_spec.rb
and spec/services/api/v1/deserialization_service_spec.rb.

The Extensions fragment in POST
Selection_001

Extension fragment in Response 
Selection_002

### DIFF
--- a/app/services/api/v1/deserialization_service.rb
+++ b/app/services/api/v1/deserialization_service.rb
@@ -57,15 +57,14 @@ module Api
         end
 
         # Retrieve any JSON schema extensions for this application
-        # rubocop:disable Metrics/AbcSize
         def app_extensions(json: {})
           return {} unless json.present? && json[:extension].present?
 
-          app = ::ApplicationService.application_name.split('-').first.downcase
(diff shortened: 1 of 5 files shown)


briri (approved):
looks good @johnpinto1

PR #3141, opened 2022-04-05
Tinymce size fix

### LINKED ISSUE #3138: Tinymce initialization inconsistent

3.0.6

**Expected behaviour:**
tinymce boxes should open in a cinsistent size

**Actual behaviour:**
tinymce boxes open with different sizes. Some too large, Some too small. When clicked on they correct to the proper size

**Steps to reproduce:**
open or create a a plan with multiple sections which have tinymce boxes in. Click to open and close them and tinymce boxes will be diferent sizes. Behaviour is also browser dependent e.g. Firefox and Chrome behave differently.

image

briri:
Fixes #3138.

Hey @raycarrick-ed, I did a quick diff of the TinyMCE initialization files between DMPTool and DMPRoadmap and noticed that I have the following change. When you have a chance can you see if this fixes the issue you've been seeing with the TinyMCE editors?

### DIFF
--- a/app/javascript/src/utils/tinymce.js.erb
+++ b/app/javascript/src/utils/tinymce.js.erb
@@ -29,7 +29,7 @@ export const defaultOptions = {
   target_list: false,
   elementpath: false,
   resize: true,
-  min_height: 230,
+  autoresize_min_height: 230,
   autoresize_bottom_margin: 10,
   branding: false,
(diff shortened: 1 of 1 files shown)


raycarrick-ed (approved):
Tried it out locally and it looks like it fixes the problem.

PR #3143, opened 2022-04-05
proposal to avoid used unpublished templates

### LINKED ISSUE #2860: old unpublished template should not be editable

`v3.0.2`

**Expected behaviour:**

* create a template, and publish it
* create a plan based on that template
* manually unpublish it by clicking "unpublish"
* add question (or make any change) to that template
* new version should be created, because there are plans attached to it

**Actual behaviour:**

* I can freely add changes to that template, without a new version being created. This way the changes are visible in the new plan..

**Steps to reproduce:**

* create a template, and publish it
* create a plan based on that template
* manually unpublish it by clicking "unpublish"
* add question (or make any change) to that template
* go to your plan: the changes are visible. 

B.T.W: the same happens when you publish a new version, and the old published one becomes unpublished again

I guess the problem here is that published=false looks like the first (draft) state of the template,
when no plans can be attached, so the interface has no clue it was used before. Maybe set `archived=true`?
I thought that `archived=true` was meant as the last stage in the lifecycle of a template; at least for migrated
templates, from dmponline_v4..

nicolasfranck:
Fixes #2860 

Changes proposed in this PR:
- when button "unpublish" is clicked, and that template is published and that template is used by one or more plans, a new version should be created, instead of risking to alter the structure of existing plans
- select and update should happen within the same transaction


briri (approved):
This looks reasonable to me. Thanks @nicolasfranck

app/controllers/org_admin/templates_controller.rb
```diff
@@ -292,11 +292,13 @@ def publish
     def unpublish
       template = Template.find(params[:id])
       authorize template
-      versions = Template.where(family_id: template.family_id)
-      versions.each do |version|
-        unless version.update_attributes!(published: false)
-          flash[:alert] = _("Unable to unpublish your #{template_type(template)}.")
+      Template.transaction do
+        # expected: template is latest
+        if template.published? && template.plans.any?
```
briri, line 297: Change this to a one-liner `template.generate_version! if template.published? && template.plans.any?` to make Rubocop happy

briri (changes requested):
Thanks for patching this @nicolasfranck. Rubocop has one complaint. If you can switch that if statement to a one-liner it should make Rubocop happy

PR #3150, opened 2022-04-12
fix issue with string token for model save errors

briri:
Fixes an issue with the string token for model errors which was causing messages for the user to appear as 'Unable to save your changes: {errors}'


raycarrick-ed (approved):
Good catch! Small but annoying.

PR #3155, opened 2022-04-13
Referenced in DCC Bug 698 - The School departments functionality broken.

johnpinto1:
Referenced in a comment in DCC bug 698 https://github.com/DigitalCurationCentre/DMPonline-Service/issues/698#issuecomment-1096577463

Fix involved following change:
 - in app/policies/department_policy the index?() policy method
 reference @department (which does not exist). Even changing @department
to @record fails as @record is the Department class (not an instance of
Department). Inspect of @record in index?() shows
 "Department(id: integer, name: string, code: string, org_id: integer,
created_at: datetime, updated_at: datetime)".
So replaced
    (@user.can_org_admin? && @user.org.id == @department.org_id) ||
     @user.can_super_admin?
by    @user.can_org_admin? || @user.can_super_admin?
as @department is nil and @record.org_id id not defined for a Class
object Department.

This fixes issue. But to see results you will need to change

> 
>     The largest page size allowed in requests to the API (all versions)
>     config.x.application.api_max_page_size = 100

which I have done
**config.x.application.api_max_page_size = 10**

Selection_035

Selection_037

**"View all"** shows
Selection_036

Reverting back to with
**config.x.application.api_max_page_size = 100**
you see the same set of Departments for **"View less"** and **"View all"**
Selection_039
Selection_038

### DIFF
--- a/app/policies/department_policy.rb
+++ b/app/policies/department_policy.rb
@@ -6,8 +6,7 @@ class DepartmentPolicy < ApplicationPolicy
   # NOTE: @user is the signed_in_user and @record is an instance of Department
 
   def index?
-    (@user.can_org_admin? && @user.org.id == @department.org_id) ||
-      @user.can_super_admin?
+    @user.can_org_admin? || @user.can_super_admin?
   end
 
   def new?


briri (approved):
Makes sense @johnpinto1. Thanks!

Issue #3163, opened 2022-04-25
Conditional questions causing plans to disapear 


mariapraetzellis:
Plans that include a conditional question are disappearing. For example, a user can successfully create a plan from a template including a conditional question, but when they go back to view the plan, the plan will disappear from the page. 

We temporarily fixed this issue by removing the conditional question from the template, but I believe this bug will be affecting all plans that include conditional questions.

johnpinto1:
@raycarrick-ed This seems to be related to issue I noticed when dealing with a bug for Maastricht. The plan sections container continues to have style="display:none". Will investigate.

johnpinto1:
@mariapraetzellis, @briri Not yet identified cause, but working on it. As I have an example which fails.

johnpinto1:
Just proved in broken case cause this rendering cuases everything in **content_for** to be surrouned by a display:none  block. 
  ```
<%= render partial: 'phases/edit_plan_answers',
               locals: {
                 plan: plan,
                 phase: phase,
                 answers: answers,
                 readonly: readonly,
                 base_template_org: base_template_org,
                 guidance_presenter: guidance_presenter,
               } %>
```

johnpinto1:
Thanks @briri. A cursory test shows commenting out hide questions fixes issue.

Will tomorrow dig deeper as suggested in your comment.

pherterich:
I think @raycarrick-ed added this to the latest release, so I will close out here.

johnpinto1:
@pherterich I can't see it in https://github.com/DigitalCurationCentre/roadmap/commits/deploy/dmponline?before=99d780716fca202f19a6dcb7da40ce62edb52e76+35&branch=deploy%2Fdmponline&qualified_name=refs%2Fheads%2Fdeploy%2Fdmponline 
It is possible has been patched by @raycarrick

pherterich:
It is patched and part of release 3.1.1 proper

PR #3164, opened 2022-04-27
Roadmap bug #3163 -  Fix for bug "Conditional question causing plans …

johnpinto1:
…to disappear."

Cause is commit bb5941e5d8998ba142be014c35bdb964f4fd86c5
"Added toggleable guidance/comments section"  the following changes were made
to file  app/javascript/src/utils/sectionUpdate.js:
the question containers classes were changed as follows .row -> .question-body & .col-md-8 -> .question-section:

```
-                <div class="row">
-                  <div class="col-md-8">
+                <div class="question-body">
+                  <div class="question-section">
```

This broke the function getQuestionDiv() in app/javascript/src/utils/sectionUpdate.js.
Fix for issue #3163.

Changes:
- in  app/javascript/src/utils/sectionUpdate.js the function was updated
as follows:
```
-export const getQuestionDiv = (id) => $(`#answer-form-${id}`).closest('.row');
+export const getQuestionDiv = (id) => $(`#answer-form-${id}`).closest('.question-body');
```

I used an example we had for issue to debug and fix.

**Before:**

- Write Plan tab displayed:
Selection_004
The whole Plan panel container was hidden because it was the closest div with class .row.

Selection_005

**After**
The plan container is visible with the removed question with id 6177 (as required with style="display:none").

Selection_009
Selection_010

Section removed would have displayed
Selection_007

### DIFF
--- a/app/views/plans/_download_form.html.erb
+++ b/app/views/plans/_download_form.html.erb
@@ -119,7 +119,7 @@
             options_for_select(Settings::Template::VALID_MARGIN_RANGE.to_a,
                                @export_settings.formatting[:margin][:right]),
             class: 'form-control',
-            "data-default": @plan.template.settings(:export).formatting[:margin][:rigth] %>
+            "data-default": @plan.template.settings(:export).formatting[:margin][:right] %>
       </div>
     </div>
   </div>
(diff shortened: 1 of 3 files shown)


briri (approved):
thanks for digging into this one @johnpinto1

PR #3167, opened 2022-05-02
Add array judgement for creators in text and csv download 

pengyin-shan:
The `join` method was used to cause errors for `@hash[:attribution]` after it was changed to string. Add judgment in case some organizations want to include more than one creator.


briri (approved):
Thanks @pengyin-shan this looks like a great approach

briri:
Looks like the change made Rubocop unhappy. I think it is fine to just add `, Metrics/PerceivedComplexity` to the rubocop exception comments that wrap that method.

pengyin-shan:
> Looks like the change made Rubocop unhappy. I think it is fine to just add `, Metrics/PerceivedComplexity` to the rubocop exception comments that wrap that method.

I see. Yeah, I also feel there's no easy way to get around it. I removed the corresponding check

PR #3169, opened 2022-05-04
Upgraded TinyMCE to V5

benjaminfaure:
Upgraded to the latest release of TinyMCE V5
(V6 just came out, it might be a bit soon)

Hopefully everything works !


briri:
Not sure why these 2 tests would be failing. The page doesn't even have a TinyMCE editor. Switching the  target branch to rails_6 instead since this can be part of the larger upgrade and we will need to dig in and fix other tests anyway

PR #3171, opened 2022-05-10
metadata standard does not always have locations

nicolasfranck:
Changes proposed in this PR:
- metadata.locations is sometimes `nil`, which makes this view crash


briri (approved):
thanks for patching this @nicolasfranck

PR #3184, opened 2022-06-07
do not show black left border on guidance-section when comments are not toggleable

### LINKED ISSUE #3183: Do not show black left border when comments are not toggleable

`3.1.1.`

**Expected behaviour:**

When configuration option `config.x.application.guidance_comments_toggleable` is set to `false`
that black left border of 5px on the div  with class`guidance-section` should not be there, as it
is serving no purpose.

https://github.com/DMPRoadmap/roadmap/blob/development/app/views/phases/_edit_plan_answers.html.erb#L100

nicolasfranck:
Fixes #3183 

Not sure if this is the way to do it, as the css style for the 5px border is in the css,
and this code probably will not work anymore if that moves..


app/views/phases/_edit_plan_answers.html.erb
```diff
@@ -97,6 +97,7 @@
                     </div>
                   </div>
                   <% style = guidance_comments_opened_by_default ? '' : 'display: none' %>
+                  <% style = Rails.configuration.x.application.guidance_comments_toggleable ? style : style + ';border-left:none' %>
```
briri, line 100: Preference would be to handle that in the Sass file:`app/assets/stylesheets/blocks/_question_form.scss`.

The initial 'style' assignment above probably shouldn't even run if that `guidance_comments_toggleable` flag is false.

Maybe a new class in the Sass file like this:
```
  .no-toggle-section {
      border-left: none;
    }
```
And then in the view have:
```
<% if Rails.configuration.x.application.guidance_comments_toggleable %>
  <% style = guidance_comments_opened_by_default ? '' : 'display: none' %>
  <div class="guidance-section" style="<%= style %>">
<% else %>
  <div class="guidance-section no-toggle-section"> 
<% end %>
```

Ideally that `display: none` would also be handled by the `hide` css class, but that would require additional changes to the JS probably to remove/add the class instead of changing the display style.

nicolasfranck: Isn't it a bit odd to add a new css style, and keep the `display:none` for this case only?
`.no-toggle-section` is not really reusable

briri:
this can be managed locally for each installation

Issue #3200, opened 2022-08-08
The V1 API call GET /api/v1/plans returns duplicate plans in some cases.


johnpinto1:
Please complete the following fields as applicable:

**Expected behaviour:**
The V1 API call GET /api/v1/plans should return unique plans.

**Actual behaviour:**
Noticed that some returned plans are duplicates.

**Steps to reproduce:**
Retrieve all plans by using pagination, e.g.
curl -k -v -X GET <DMP base url> /api/v1/plans\?page\=1\&per_page\=100 -H "Authorization: Bearer <YOUR V1 AUTHENTICATION TOKEN>" -H "Content-Type: application/json"
Continue by changing "page=" value each time until all plans retrieved.
Should see some duplicate plans.

The reason for this is because we do not remove duplicate in the following methods
https://github.com/DMPRoadmap/roadmap/blob/382e14f01a020c0fdc9009b8948fed83a0cddc02/app/policies/api/v1/plans_policy.rb#L25-L40

PR #3202, opened 2022-08-10
Bug #3200 - Fix for the V1 API call GET /api/v1/plans returns duplica…

johnpinto1:
…te plans in some cases.

Fix for issue #3200.

Changes in app/policies/api/v1/plans_policy.rb methods
plans_for_client() and plans_for_user() returns a unique array of plan
ids by finally calling the method uniq() on the ids the methods
previously returned.


cdluc3-admin (approved):
thanks for patching this @johnpinto1

PR #3208, opened 2022-08-15
update to prevent users from being able to make all of their guidance…

### LINKED ISSUE #3199: Guidance page returns a 500 error in specific scenario

v3.1.1

**Expected behaviour:**
The Admin -> Guidances loads

**Actual behaviour:**
A 500 error is thrown when I have no Guidance Groups that are NOT 'optional subsets'

**Steps to reproduce:**
- Login as an Org Admin. 
- Go to the Admin -> Guidance page
- Make sure there is a Guidance Group whose name matches the Org abbreviation
- Update all Guidance Groups so that the 'optional subset' flag is true
- Navigate back to the Admin -> Guidance page if necessary

**Diagnosis**
The controller attempts to create the default GuidanceGroup for the Org when none exists. This code though uses the Org abbreviation as the guidance group name. The guidance group model though requires that the name be unique. This cause the save/create to fail resulting in a 500 error
https://github.com/DMPRoadmap/roadmap/blob/382e14f01a020c0fdc9009b8948fed83a0cddc02/app/controllers/guidances_controller.rb#L19

It would be better to prevent users from making all of the guidance groups 'optional subsets'. Change the view logic to NOT enable/display this flag if there is only one guidance group, or if there are no other guidance groups that are not 'optional subsets'

briri:
Fixes #3199.

Changes proposed in this PR:
- Add logic to prevent an org admin from making all of their guidance groups 'optional subsets'


briri:
Looks like these were already merged into the development branch at some point. @benjaminfaure and @martaribeiro if you could both review anyway it would be appreciated

briri:
Closing as this was already fixed and merged elsewhere

PR #3209, opened 2022-08-30
pluck with ampersand returns not id

nicolasfranck:
Changes proposed in this PR:
- `plans.pluck(&:id)` returns array of array values containing the plan attributes values. Probably typo. Should be `plans.pluck(:id)`


briri (approved):
I think you're right @nicolasfranck probably a typo

briri:
Tests are failing on the development branch (Rails 6) and we are aware, so merging this one even though the tests fail

Issue #3187, opened 2022-06-13
Rails 6 Tasks (to-do list)


gjacob24:
Creating this ticket to add tasks to it that need to be completed before the release with the upgrade to 6.1 
(including any manual testing required, issues found while testing etc.)

briri:
Note that the new underlying AWS environment for https://dmproadmap.uc3dev.cdlib.org/  is not sending out emails. I need to add AWS Simple Email Service (SES) to the application stack.

briri:
Receive an error when trying to change a plan's visibility. The error from the logs is:
```
I, [2022-07-06T12:45:42.494139 #158]  INFO -- : [30320f2d-4da4-409e-afdc-0ff4faae80e8] Completed 500 Internal Server Error in 75ms (ActiveRecord: 29.7ms | Allocations: 12075)
F, [2022-07-06T12:45:42.495232 #158] FATAL -- : [30320f2d-4da4-409e-afdc-0ff4faae80e8]   
[30320f2d-4da4-409e-afdc-0ff4faae80e8] NoMethodError (undefined method `to_sym' for 0:Integer Did you mean?  to_s):
[30320f2d-4da4-409e-afdc-0ff4faae80e8] app/mailers/user_mailer.rb:148:in `plan_visibility'
[30320f2d-4da4-409e-afdc-0ff4faae80e8] app/controllers/plans_controller.rb:412:in `block in visibility'
```

briri:
test server is reporting an invalid wkpdftohtml path

gjacob24:
> Receive an error when trying to change a plan's visibility. The error from the logs is:
> 
> ```
> I, [2022-07-06T12:45:42.494139 #158]  INFO -- : [30320f2d-4da4-409e-afdc-0ff4faae80e8] Completed 500 Internal Server Error in 75ms (ActiveRecord: 29.7ms | Allocations: 12075)
> F, [2022-07-06T12:45:42.495232 #158] FATAL -- : [30320f2d-4da4-409e-afdc-0ff4faae80e8]   
> [30320f2d-4da4-409e-afdc-0ff4faae80e8] NoMethodError (undefined method `to_sym' for 0:Integer Did you mean?  to_s):
> [30320f2d-4da4-409e-afdc-0ff4faae80e8] app/mailers/user_mailer.rb:148:in `plan_visibility'
[...]

PR #3223, opened 2022-10-24
Fix/plan visibility

benjaminfaure:
Fixes template & plan visibility issues, see : https://github.com/DMPRoadmap/roadmap/issues/3187#issuecomment-1176634328
Updates to gems & JS packages


briri:
going to merge this. will investigate the failing mysql test further. I am not having any issues creating a plan, so I think the failure has to do with the test itself.

Issue #3217, opened 2022-10-03
Template Search issue


gjacob24:
Please complete the following fields as applicable:

**What version of the DMPRoadmap code are you running? (e.g. v2.2.0)** v3.1.1

**Expected behaviour:** The search feature for templates should return the templates with titles that contain the keyword used in the search

**Actual behaviour:** nothing is returned i.e. no templates
<img width="1152" alt="Screenshot 2022-10-03 at 14 04 28" src="https://user-images.githubusercontent.com/97518971/193583457-e5115265-1ea2-4664-9efa-f1fa4aff13ef.png">

**Steps to reproduce:** 
Log in (as admin/super admin) -> Select Templates (from the dropdown menu) -> Type a keyword and click on 'Search'

gjacob24:
@johnpinto1 has found a way to fix this. It has been tested with Rails 5 and 6 locally. I will create a PR with the code changes against the development branch so that it can be included in the upcoming Rails 6 release.

Issue #3139, opened 2022-04-04
Upgrade Ruby version


briri:
Upgrade Ruby from 2.6.3 to 2.7 or 3.x

See the release notes at: https://www.ruby-lang.org/en/

Related to #3103

briri:
The big question here is whether or not we should consider upgrading to 3.0 or 3.1.

benjaminfaure:
Seems like Rails 5 is not compatible with Ruby 2.7 according to these sources : 
https://www.fastruby.io/blog/ruby/rails/versions/compatibility-table.html
https://github.com/rails/rails/issues/38426

The official guide says 2.2.2 or newer : https://guides.rubyonrails.org/upgrading_ruby_on_rails.html

I tried switching from Ruby 2.6 to 2.7 on the roadmap code, basic features seems to be working fine, but I guess we can expect some side effects

briri:
Upgraded to 2.7. Will open a new ticket for 3.0

Issue #3104, opened 2022-02-09
In downloaded template pre-ticked answers do not download 


magdalenadrafiova:
@briri could you check - this is def not working for DMPonline (@raycarrick )

**Expected behaviour:**]
When you download a template and there are pre-ticked answers one would expect to see these in the download form. 

**Actual behaviour:**
I have created a template with pre-ticked answer but when I go to download this is missing

**Steps to reproduce:**
Ensure that if there is a pre-ticked answer in template this will get downloaded

johnpinto1:
@raycarrick-ed  I will look at this next.

johnpinto1:
I have added a PR for this. The changes can be seen in screenshots
Example screenshots of template:
**![Selection_017](https://user-images.githubusercontent.com/8876215/166442839-b8fdbc23-e670-4fcb-b3c5-d77b28410589.png)**
**Preview**
![Selection_018](https://user-images.githubusercontent.com/8876215/166442877-1496a34a-cba4-456b-a3e8-d719a0a11cd5.png)

Issue #3103, opened 2022-02-03
Rails 6 


benjaminfaure:
Hello everyone,

According to Rails maintenance policy, Rails 5.2 we'll reach its end of life in June 2022
https://guides.rubyonrails.org/maintenance_policy.html

Seems like yet another Rails upgrade is on the horizon.

Rails 7 is already out, I guess they switched to yearly release for major versions of the framework. Hopefully upgrading won't be such a big step anymore

briri:
Just discovered today that Rails is [no longer supporting the Webpacker gem](https://github.com/rails/webpacker). I learned this after running `yarn upgrade` and noticing that all of the Karma tests would no longer run due to a mismatch in dependencies for webpack.

Rails 7 moves away from Webpacker and provides [3 new options for how to manage JS](https://world.hey.com/dhh/rails-7-will-have-three-great-answers-to-javascript-in-2021-8d68191b)

We should also consider upgrading:
- Bootstrap (we are 2 major versions behind) - this will require significant work
- TinyMCE - we tried upgrading but the editors stopped loading
- JQuery - we might want to consider moving away from JQuery and just using vanilla JS (depending on how much work will be involved to upgrade)

benjaminfaure:
I took some time and tried to upgrade the roadmap code to Rails 6.1.

It went fairly well. I tested some of the feature and with some minor corrections I managed to get the roadmap code to run.
I'm not sure how we could start to share the work done on the migration. Should we create a new branch and commit it ?

pherterich:
Tagging @gjacob24 here as she wanted to give this an initial look for the DMPonline team

briri:
Added a new wiki page to record upgrade notes: https://github.com/DMPRoadmap/roadmap/wiki/Rails-Upgrade-Notes

The intended audience would be developers who manage their own instances and may need to upgrade their own cutomizations

briri:
I was just doing some reading on Rails 7 and discovered that they're moving away from Webpacker and UJS (which they introduced in Rails 5/6). An apparent side effect is that there is no longer support for `js.erb` (would have been nice if they deprecated instead and made this backwards compatible). 

This is going to have a potentially significant impact on us: https://guides.rubyonrails.org/working_with_javascript_in_rails.html

pengyin-shan:
[...]

PR #3226, opened 2022-10-24
Upgrade to v4.0

briri:
## Researchers
- Updated the PDF coversheet to include all Contributors
- Fixed an issue that was causing the link to organizational guidance to repeat 

## Administrators
- Fixed an issue that was impacting API v0 and preventing the batch assignment of users to departments
- Fixed an issue that was preventing admins from performing a search on the Templates page #3217
- Removed case sensitivity from the validation rules for Org and Department names.
- Fixed issues with API v1 
- Fixed an issue that was causing plans that were based off of templates with conditional questions to disappear when the user triggered the condition #3163
- Your admin can now set a flag that makes the 'coversheet' checkbox on the Plan Downloads page to be checked by default  #3063 #3104 
- Fixed an issue that was causing plans to not appear in the publicly visible if 100% of their questions had not been answered #3086

## Developers
This release includes a major upgrade of Rails (5.x to 6.1) and an upgrade from Ruby 2.6 to 2.7.6. Please review the following release notes as well as our [upgrade guide](https://github.com/DMPRoadmap/roadmap/wiki/Upgrading-from-v3.x-to-v4.x) and be aware that you will most likely need to adjust/modify any local customizations you may have made within your local fork of the code.

**This release includes an upgrade from Ruby 2.6 to Ruby 2.7.6 and Rails 5.x to Rails 6.1!**

You will need to do the following:
- Review the new default configuration options for Rails 6.1 and adjust as necessary for your implementation. The config settings can be found in `config/initializers/new_framework_defaults_6_1.rb`
- Set a `DMPROADMAP_HOST` environment variable. Rails 6+ introduces new middleware that helps mitigate DNS rebinding attacks. This includes a hostname whitelist. The value you assign to this environment variable will be added to the whitelist.
- Update the `SUPPORTED_LOCALES` `CLIENT_LOCALES` and `DEFAULT_LOCALE` in the `config/initializers/translation.rb` for your instance
- If you have any raw SQL queries were deprecated in Rails 5.2. You may need to convert these over. [See this article](https://medium.com/@mitsun.chieh/activerecord-relation-with-raw-sql-argument-returns-a-warning-exception-raising-8999f1b9898a)

Updates:
- Updated Ruby from 2.6.3 to 2.7.6 #3139
- Updated Rails from 5.2 to 6.1 #3103 #3187
- Updated the TinyMCE editor to v5 and installed the 'oxide' skin #2942
- Switched from Turbolinks to Turbo 
- Updated all JS and gem dependencies
- Fixed an issue that allowed researchers to add an ORCID for a contributor that included spaces resulting in URLs that looked like `https://orcid.org//  0000-0000-0000-0000. ` #3212
- Made some minor modifications to improve query responsiveness on the Contributors, Guidances, Plans, Orgs, Users and Templates pages
- Replaced all occurrences of `update_attributes` with `updates` since `update_attributes` was deprecated
- Upgraded chart.js which is used on the Usage page to render the bar and pie charts
- Made changes to accommodate the new Zeitwerk autoloader (which has stricter convention-over-configuration rules):
  - Changed the name of the `JSONLinkValidator` class file to `JsonLinkValidator`
  - Updated models so that any JSON fields have their default values defined prior to serialization
  - Updated deprecated occurrences of `record.errors[:model] << 'foo'` to `record.errors.add(:model, 'foo')`
  - Updated JS babel configuration
  - Updated the `config/environments/*.rb` files to use the new `DMPROADMAP_HOST` environment variable to defined the `Rails.application.routes.default_url_options` and new `config.hosts` setting.
  - Added a new `download_coversheet_tickbox_checked` flag to the `config/initializers/_dmproadmap.rb` initializer that will default the Plans Download page to default that checkbox to be true
  - Removed the out-of-date eslint package from webpacker. We will investigate re-introducing a JS linter in a future release
  - Fix to the `db/seeds.rb` file that is used for new installations so that it uses the `find_or_create_by` instead of `create!` to prevent duplicates when the DB already exists.
  - Updated and fixed tests as needed


pengyin-shan (approved):
Looks good to me!

benjaminfaure:
It might be good to add a warning for people using query methods with raw SQL. It's been deprecated since Rails 5.2 and removed in Rails 6 
You can see an example here : https://medium.com/@mitsun.chieh/activerecord-relation-with-raw-sql-argument-returns-a-warning-exception-raising-8999f1b9898a

DMP OPIDoR is impacted because we're using JSON fields with queries on the JSON data.
For example : 
`.pluck("data->>'keyword'")`
`.order("data->>'lastName', data->>'firstName'")`

benjaminfaure (approved):
looking good so far !

It might be good to add a warning for people using query methods with raw SQL. It's been deprecated since Rails 5.2 and removed in Rails 6 
You can see an example (and a solution) here : https://medium.com/@mitsun.chieh/activerecord-relation-with-raw-sql-argument-returns-a-warning-exception-raising-8999f1b9898a

DMP OPIDoR is impacted because we're using JSON fields with queries on the JSON data.
For example : 
`.pluck("data->>'keyword'")`
`.order("data->>'lastName', data->>'firstName'")`

briri:
thanks @benjaminfaure and @pengyin-shan 

I added a note to the `You will need to do the following` section for Developers that mentions the raw SQL change

Issue #3216, opened 2022-09-30
bin/rails db:schema:load issue


StCyr:
Please complete the following fields as applicable:

**What version of the DMPRoadmap code are you running? (e.g. v2.2.0)**
master branch

**Expected behaviour:**
bin/rails db:schema:load should just work

**Actual behaviour:**
```
root@dmponline-rails-uk:/home/roadmap# bin/rails db:create
Created database 'roadmap_development'
Database 'roadmap_development' already exists
root@dmponline-rails-uk:/home/roadmap# bin/rails db:schema:load
-- create_table("annotations", {:id=>:integer, :force=>:cascade})
   -> 0.0332s
-- create_table("answers", {:id=>:integer, :force=>:cascade})
   -> 0.0388s
-- create_table("answers_question_options", {:id=>false, :force=>:cascade})
   -> 0.0237s
-- create_table("api_clients", {:id=>:integer, :force=>:cascade})
   -> 0.0256s
-- create_table("conditions", {:id=>:integer, :force=>:cascade})
   -> 0.0228s
-- create_table("contributors", {:id=>:integer, :force=>:cascade})
   -> 0.0344s
-- create_table("departments", {:id=>:integer, :force=>:cascade})
   -> 0.0228s
-- create_table("exported_plans", {:id=>:integer, :force=>:cascade})
   -> 0.0163s
-- create_table("external_api_access_tokens", {:force=>:cascade})
   -> 0.0393s
-- create_table("guidance_groups", {:id=>:integer, :force=>:cascade})
   -> 0.0377s
-- create_table("guidances", {:id=>:integer, :force=>:cascade})
   -> 0.0364s
-- create_table("identifier_schemes", {:id=>:integer, :force=>:cascade})
   -> 0.0317s
-- create_table("identifiers", {:id=>:integer, :force=>:cascade})
   -> 0.0566s
-- create_table("languages", {:id=>:integer, :force=>:cascade})
   -> 0.0297s
-- create_table("licenses", {:force=>:cascade})
   -> 0.0501s
-- create_table("metadata_standards", {:force=>:cascade})
   -> 0.0296s
-- create_table("metadata_standards_research_outputs", {:force=>:cascade})
   -> 0.0478s
-- create_table("notes", {:id=>:integer, :force=>:cascade})
   -> 0.0551s
-- create_table("notification_acknowledgements", {:id=>:integer, :force=>:cascade})
   -> 0.0479s
[...]

PR #3229, opened 2022-10-27
Updates to bin/setup to make installation easier

briri:
Updated the `ruby bin/setup` script so that it:
- Runs `bundle install` and `yarn install`
- Creates the `.env`, `config/database.yml` and `config/initializers/wicked_pdf.rb` files
- Opens `vim` to allow user to edit the information in the new `.env` file 
- Creates the initial `config/credentials.yml.enc` and `master.key` files and opens it for editing
- Builds the DB
- Populates the `licenses`, `metadata_standards`, and `research_domains` tables
- Clears out any existing Rails cache
- Provides instructions for populating the `repositories` table ... didn't include that one since it can take over 10 minutes

Will update the wiki installation page once this is merged into `main` in the next release.

Should address the issue outlined here: #3216


benjaminfaure (approved):
Looking good :)

johnpinto1:
@briri @benjaminfaure @pengyin-shan @martaribeiro Not sure why the setup fails for me. 

I am on an Ubuntu PC and I am using RVM to set the Ruby to version 2.7.2.
I have run the setup several times and it fails at db setup:

> ▶ ruby bin/setup     

> == Installing dependencies ==
> The Gemfile's dependencies are satisfied
> 
> == Copying sample files ==
> 
> == Preparing database ==
> /mnt/DataDrive1/DMP-WS/without-docker-roadmap/config/initializers/recaptcha.rb:8:in `block in <main>': undefined method `[]' for nil:NilClass (NoMethodError)
> 	from /home/jpinto/.rvm/gems/ruby-2.7.2/gems/recaptcha-5.12.3/lib/recaptcha.rb:37:in `configure'
> 	from /mnt/DataDrive1/DMP-WS/without-docker-roadmap/config/initializers/reca

briri:
did the `.env` and `credentials` files ever open up in the editor for you @johnpinto1? If you look at the 2 new `config/credentials.yaml.[db]` files you'll see the recaptcha settings. It should be creating those files and opening them for you to edit.

I haven't used a PC though in over 15 years and don't a machine available to test with. 

@johnpinto1 I just pushed a change that attempts to set the `EDITOR` env variable based on whether the platform is Windows. I suspect that it may fail though when it tries to seed the credentials file. Can you have a look and try it out?

If it doesn't work, can you use that `Gem.win_platform?` check and create versions of the following lines that will work for Windows? Feel free to make your changes to this branch (`git fetch origin install-update && git checkout install-update`)
- Seed the credentials file with the correct sample based on the version of the DB the user specified: `system! "EDITOR='echo \"$(cat config/credentials.yml.#{db_adapter})\" >' bin/rails credentials:edit"`
- Open the .env in an editor `system! '$EDITOR .env'` and wait for the user to save before continuing
- Open the credentials in an editor `system! 'bin/rails credentials:edit'` and wait for the user to save before continuing

johnpinto1:
@briri That sorted issue of credentials. I just need to sort issue with duplicate tables after running db:migrate.
So I guess sorted. Thanks.

>  ▶ rails db:migrate
> Running via Spring preloader in process 3895
> == 20131118094629 ChangeVersionsPublished: migrating ==========================
> -- add_column(:versions, :published_tmp, :boolean, {})
>    -> 0.0014s
> -- table_exists?("versions")
>    -> 0.0004s

johnpinto1:
@briri I am testing the setup with Postgres. It is failing because the **annotations** table is never created in the migrations. Am I misunderstanding something. I don't see a migration file for creating this table.

Also with Postgres one has to run migrations twice at least  because of the following error within a **db:migrate** because you can't change a type within a single transaction several times:

> Caused by:

> PG::FeatureNotSupported: ERROR:  cached plan must not change result type

briri:
Thanks for trying it out @johnpinto1. I did some more reading on the new `db:prepare` and it looks like it runs: `rails db:create && rails db:migrate && rails db:seed` which is not what we want. I just switched it to use `db:setup` which runs `rails db:create && rails db:schema:load && rails db:seed`.

There are some very old migrations in the `db/migrations/` directory that attempt to manipulate data using ActiveRecord models that no longer exist, so `db:migrate` is impossible to run from scratch.

johnpinto1:
@briri Phew that was the quickest fix. Yippee! up and running with Postgres using setup. Thanks.

PR #3231, opened 2022-11-04
fix YAML safe load capatibility potential error

pengyin-shan:
Fixes a possible error caused by a bug in Rails. Some users see this error when moving the 'Download' tab:
image

Rails proposed the solution: https://discuss.rubyonrails.org/t/cve-2022-32224-possible-rce-escalation-bug-with-serialized-columns-in-active-record/81017

Thus, made the following changes:
- Make sure to upgrade to Rails 7.0.3.1, 6.1.6.1, 6.0.5.1, 5.2.8.1 (needs to be done separately by each team)
- In the configuration file, add trusted resources (DMP Assistant has seen a bug from `Symbol`. `HashwithDifferentAccess`, `Date` and `Time` are also trusted and should be added)
- I'm using an M2 Mac so `Gemfile.lock` got updated with the arm64 version there


briri (approved):
Looks good @pengyin-shan 

What do you mean by "when moving the 'Download' tab"? I haven't seen any issues with this in DMPTool Rails 6.1.

pengyin-shan:
> Looks good @pengyin-shan
> 
> What do you mean by "when moving the 'Download' tab"? I haven't seen any issues with this in DMPTool Rails 6.1.

So this occurred for some user groups (so far DMP Assisant has only seen one user has this problem). I meant to use click 'Download' tab. i.e. User tried to download the plan, but the error message will show and bring the user back to their dashboard

pengyin-shan:
> > Looks good @pengyin-shan
> > What do you mean by "when moving the 'Download' tab"? I haven't seen any issues with this in DMPTool Rails 6.1.
> 
> So this occurred for some user groups (so far DMP Assisant has only seen one user has this problem). I meant to use click 'Download' tab. i.e. User tried to download the plan, but the error message will show and bring the user back to their dashboard

As an update, I moved separate settings from the configuration file to `application.rb`, since we need all environment following this rule

PR #3232, opened 2022-11-08
Add error message and update saving message for plan writing

pengyin-shan:
This ticket is created as an improvement for a user's plan-saving problem. The user so far only sees 'Saving...' and 'Answered by...' messages after saving the plan successfully.

However, a user of DMP Assistant was stuck on the 'Saving' message and no explicit guidance showed that if the user turns to other tabs now, the content in the rich text box will be lost and unrecovered. *So she lost all her inputs of plan content*.

Thus:
1) **We need to add an error message in the `error-saving` span. We used to have nothing in it.**
2) We need a more explicit expression of telling the user to save the plan to a separate file if an error occurred in this process. So the user won't feel misled and fresh this page then lose all input.

Changes proposed in this PR:
1) For error message:
Screenshot 2022-11-08 at 2 31 13 PM
2) For saving... message (very long, but users normally skim through it if their plan is saved successfully, so it won't affect regular user experience)
Screenshot 2022-11-08 at 3 12 25 PM

Please let me know if you have any suggestions for these messages!


app/views/answers/_status.html.erb
```diff
@@ -2,8 +2,24 @@
...
-<span class="label label-info status" style="display:none;" data-status="saving"><%= _('Saving...') %></span>
-<span class="label label-warning status" style="display: none;" data-status="error-saving"></span>
+<%
+  helpdesk_email = Rails.configuration.x.organisation.helpdesk_email
+  email_subject = _('Plan writing error related to %{tool_name}') %{ :tool_name => tool_name }
+%>
+<span class="label label-warning status" style="display:none;" data-status="saving">
+  <%= _("Saving...Please do not refresh or go to another page!") %>
+  <br><br>
+  <%= sanitize(_("If you stuck on this message, please save your plan to a separate file and contact us at %{helpdesk_email} .") % {
```
briri, line 12: I think "If you are stuck on this message" might be confusing since it doesn't let them know what is supposed to happen (e.g. how long is 'stuck'?). 

Suggest we either change the messaging (maybe ask the PMs)

briri (approved):
looks good to me @pengyin-shan I just have a small comment on the wording of one of the messages. I think it might be useful to run it by the PMs

app/views/answers/_status.html.erb
pengyin-shan: > I think "If you are stuck on this message" might be confusing since it doesn't let them know what is supposed to happen (e.g. how long is 'stuck'?).
> 
> Suggest we change the messaging (maybe ask the PMs)

How about "If this message doesn't disappear in 1 minute?"  @mariapraetzellis how do you feel?

pengyin-shan: @mariapraetzellis @dsisu : one more UI message suggestion is needed. For anyone sticking on the 'Saving...' message because of a system problem, we need them to remember not to fresh the page and to contact the helpdesk immediately. 

Instead of saying `If you are stuck on this message, please save your plan to a separate file and contact us at %{helpdesk_email} `, any suggestions you have?

briri:
@mariapraetzellis @dsisu can you have a look at the wording for these messages and let us know what would be appropriate?

dsisu:
Apologies for not responding sooner. Am I too late?

Which screens are we talking about exactly, at which point in the process of writing a plan are these messages meant to appear? Thanks.

PR #3236, opened 2022-11-14
Webdriver update

pengyin-shan:
Web driver version 4 (we currently have) is not capable of Mac M-powered Chip:

Screenshot 2022-11-14 at 6 29 36 PM

Fixes are done in version 5.2 (https://github.com/titusfortner/webdrivers/pull/239). Thus we should think about upgrading it.

My changes are done by adding `~5.2` in Gemfile then doing a `bundle install`


Gemfile.lock
```diff
@@ -104,13 +104,7 @@ GEM
       rack-test (>= 0.6.3)
       regexp_parser (>= 1.5, < 3.0)
       xpath (~> 3.2)
-    capybara-webmock (0.6.0)
```
briri, line 107: weird that this downgraded 5 minor versions. I'm theoretically ok with it though since it's just for running tests and not something that would be present when deployed to a production server

briri (approved):

Gemfile.lock
pengyin-shan: Might it be just my local environment changes? Can someone do the same webdriver upgrade and see if `Gemfile.lock` get the same changes as I do?

benjaminfaure (approved):

PR #3237, opened 2022-11-15
Add translation ability to funding status dropdown by creating corresponding enum - Ready

pengyin-shan:
This change will enable the dropdown of 'funding status' to be translatable:

Screenshot 2022-11-15 at 1 40 22 PM

Screenshot 2022-11-15 at 1 39 44 PM

I created a translation.io repo  (https://translation.io/pengyin-shan/dmproadmap-development/fr-CA) and tested it. It will work after syncing translations for 'Funded', 'Planned' and 'Denied'


briri (changes requested):
this will be a great enhancement for the site. I think there are probably other enums that would benefit from being translated!

Instead of setting up a separate const that contains the translatable labels, can you instead try this:
```
enum funding_status: {
    "#{_('Planned')}": 0,
    "#{_('Funded')}": 1,
    "#{_('Denied')}": 2
  }
```
That way they won't be at risk of getting out of sync if someone needs to add a new enum value.

I'm not sure what it will do if you try to save with different languages. I suspect it will work though since it should be saving the numerical value in the DB 🤷🏻‍♂️

pengyin-shan:
> @pengyin-shan can you verify that it saves properly when the user selects an item in a different language?

I tested and unfortunately, the new method won't work when selecting a different language. It only works for the defaulted language (English or French). What I think is that the translation won't work when it is used directly as the attribute.

briri:
verified that the information is still saving properly

PR #3239, opened 2022-11-16
Force local time in UI - DO NOT MERGE FOR PR

briri:
In the event that the database stores date/time entries as UTC, this will ensure that the UI displays the information in Local time.

Most of the dates in the UI that display created_at and updated_at only display the date without time (e.g. 11/15/22, 2022-11-15, etc.) so it's not probably an issue for most users. DMPTool is 8 hours behind UTC though, so when a user does something like creating a plan in the afternoon, the My Dashboard page shows the modification date as in the future.

This PR just adds a conversion to local server time.

**Do not merge this one yet. Lets discuss in the next dev meeting**

### DIFF
--- a/app/controllers/api/v0/plans_controller.rb
+++ b/app/controllers/api/v0/plans_controller.rb
@@ -117,8 +117,8 @@ module Api
       # takes in the params hash and converts to a date-range
       def dates_to_range(hash, start, stop)
         today = Date.today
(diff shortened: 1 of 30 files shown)


briri:
see if there is a configuration setting that can accomplish this instead

benjaminfaure:
Is something like that would be useful for you @briri  ? https://stackoverflow.com/questions/6118779/how-to-change-default-timezone-for-active-record-in-rails

It's quite old, I'm not sure it still working

PR #3243, opened 2022-11-21
Added EditorConfig file

benjaminfaure:
EditorConfig is a config file that helps to maintain constant coding style. 

I created the file and added some minimal rules : 
- auto insert of a final new line
- lines are idented with 2 spaces for JS, RB & ERB files

A lot of IDEs have a plugins that supports EditorConfig. Some doc & examples here : https://editorconfig.org/


pengyin-shan (approved):
Should be good to go!

PR #3255, opened 2022-12-06
Froze mail gem version

### LINKED ISSUE #3254: Latest 'mail' gem version triggers "Uninitialized constant Mail::TestMailer (NameError)"
Hello, 

After merge the latest roadmap release, I encountered the following error : 
`uninitialized constant Mail::TestMailer (NameError)`

After some investigation, the error comes from the `mail` gem that was updated 2 days ago to V2.8.0. It added compatibility to Ruby 3.1.0 but seems to have broken compatibility with other versions.

The roadmap code is not impacted because the version in Gemfile.lock is set to V2.7.1, but I think it would be good to freeze the version in our Gemfile while we are using Ruby 2.7.

benjaminfaure:
Fixes #3254


pengyin-shan (approved):
Looks good to me! Just make sure to fix the rubocop : )

briri (approved):
looks good Benjamin. Thanks for adding the comment so that we remember to come back to it.

PR #3256, opened 2022-12-06
Added scss files to editorconfig

benjaminfaure:
I forgot to add scss files to the editorconfig file. 

Should I review every scss files in the PR too ?


pengyin-shan (approved):
Looks good to me. Don't know if there's any auto-correct tool to just fix our existing scss files?

PR #3263, opened 2022-12-09
[#trivial] Translation sync and Dangerfile update

pengyin-shan:
Changes proposed in this PR:
1. Do another translation.io sync (shouldn't affect anything)
2. Change the `raise` in Dangerfile to `failure`. It used to be `fail`, but rubocop auto-corrected it to `raise`. As a note, this is the rule for rubocop:
Screenshot 2022-12-09 at 1 00 26 PM

However, observed from DMP Assistant that Dangerfile will complain about `raise`, so I changed it to `failure` to make both rubocop and danger happy.


briri (approved):
looks good. Thanks @pengyin-shan

PR #3279, opened 2023-01-11
fix to properly use max per page settings

briri:
- Update API v1 to obey they `api_max_page_size` configuration param
- Update `paginable.rb` to use `results_per_page` instead of `api_max_page_size`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,11 @@
 # Changelog
 
-### Added 
+### Added
 
 - Added CHANGELOG.md and Danger Github Action [#3257](https://github.com/DMPRoadmap/roadmap/issues/3257)
 
 ### Fixed
+
+- Fixed discrepencies with default/max per_page values for API and UI pagination
 
 ### Changed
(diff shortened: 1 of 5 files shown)


pengyin-shan (approved):
Tested your branch and it works. Thank you!

johnpinto1:
@briri can I suggest you add logic to the Rest API version of V0  **app/controllers/api/v0/base_api_controller.rb**
as you have done for **app/controllers/api/v1/base_api_controller.rb** so that **Rails.configuration.x.application.api_max_page_size** value is taken into account in V0 too now that it has been removed from **app/controllers/concerns/paginable.rb**.

johnpinto1:
@briri The API V1 works fine and UI pagination.

Found issue with API V0. It is broken. Trying to debug to get further info. Will add here.

briri:
thanks @johnpinto1. I had originally added the function there to mimic v1 but then realized that plans was the only paginated endpoint and it needed `@args` instead of the individual `@per_page` and `@page`.

PR #3292, opened 2023-02-17
disable debugger check for rubocop >= 1.45.0

### LINKED ISSUE #3291: Disable debug checking for rubocop >=1.45.0
This issue will appear for Rubocop >= 1.45.0. I tested in roadmap >=3.1.0 version.

Noted in https://github.com/portagenetwork/roadmap/issues/316, rubocop starts to complain about debugging messages after upgrading to 1.45.1:

image

It is because rubocop 1.45.0 (https://github.com/rubocop/rubocop/issues/11517) added debugger detection that will affect most of our rake tasks.

Will push the new rubocop.yml soon.

pengyin-shan:
Fixes #3291  and DMP Assisant reported common Github Action error: https://github.com/portagenetwork/roadmap/issues/319

Changes proposed in this PR:
- after DMP Assistant upgrades rubocop to version >= 1.45, the rubocop debugging detection feature (https://github.com/rubocop/rubocop/issues/11517) is added and rubocop complains all terminal messages, including those ones in rake tasks, like this:
image

- Added node version 16 specification to Github Actions

I configure rubocop to disable debugging message checking because we have a lot of necessary messages that need to show in the terminal even if in the production environment.


benjaminfaure (approved):

briri:
Thanks for this @pengyin-shan. I actually think that's a useful check for files in the /app and /spec directories. Can you modify this PR so that it only excludes rake tasks? 

Instead of 'Enabled: false' use:
```
Exclude:
    - 'lib/tasks/**/*'
```

pengyin-shan:
UPDATE on Feb 21, 2023: Github Action was forced to upgrade to node 18. Thus we see the same error as the DMP Assisant bug https://github.com/portagenetwork/roadmap/issues/319

I'll include `setup-node` Github action in this PR to assign a specific node version (16 will resolve this problem).

pengyin-shan:
> Thanks for this @pengyin-shan. I actually think that's a useful check for files in the /app and /spec directories. Can you modify this PR so that it only excludes rake tasks?
> 
> Instead of 'Enabled: false' use:
> 
> ```
> Exclude:
>     - 'lib/tasks/**/*'
> ```

Finally, all tests passed lol. I added exclusive to rubocop and updated Github actions to use node 16.

benjaminfaure:
@pengyin-shan you can solve this problem by adding `NODE_OPTIONS=--openssl-legacy-provider` to your environment 

I ran into it recently when upgrading our config to Node 18

pengyin-shan:
> @pengyin-shan you can solve this problem by adding `NODE_OPTIONS=--openssl-legacy-provider` to your environment
> 
> I ran into it recently when upgrading our config to Node 18

Unfortunately, I keep getting the `--openssl-legacy-provider is not allowed in NODE_OPTIONS`, I doubt it is because of Github action (I referred https://github.com/actions/setup-node/issues/356 but it seems not to work). Currently, I'm going to assign the node 16 version so that our test can be passed. 

@briri how do you feel about this?

briri:
@pengyin-shan was correct, Github Actions would not allow the `--openssl-legacy-provider` to be set on NODE_OPTIONS. I reverted our actions to use node 16.6.0 for now in the main branch.

Issue #3161, opened 2022-04-20
Not catching data size too large.


raycarrick-ed:
Please complete the following fields as applicable:

**What version of the DMPRoadmap code are you running? (e.g. v2.2.0)**
v3.1

**Expected behaviour:**
Fill in Research Outputs tab, create a dataset, put in ridiculously large number for the size like 10,000,000 PB and it'll crash trying to save the integer. Shold give some sort of sensible error message

**Actual behaviour:**
Generic failure 500 message.

**Steps to reproduce:**
above

PR #3294, opened 2023-02-21
update to allow the research_outputs.byte_size field to be blank

briri:
- fixes an issue related to #3161 so that you can now leave the `byte_size` field blank and output types that do not have a byte size can be saved (e.g. collection and physical object)


gjacob24 (approved):
Thanks for catching this Brian. I was so focused on testing it with different file sizes as input that I missed testing it without.

PR #3299, opened 2023-03-23
Patch issue causing template visibility to get set to organizationally visible

briri:
When an admin saves changes to their template on the template details page, the visibility os getting automatically set to 'organizationally_visible' this patch carries forward whatever the current value is.


briri:
ok. I'm going to see if I can fix the failing Github action. Looks like node might be out of date on the CI builds

briri:
GitHub actions have been patched thanks to @pengyin-shan recommendation on PR #3288 to lock node at v16.6 (for now anyway)

PR #3405, opened 2024-04-09
Fix for failing Rubocop tests.

johnpinto1:
Changes:
 - To deal with missing favicon.ico error ( ActionController::RoutingError: No route matches [GET] "/favicon.ico") : added config/routes_test.rb to add a route will return an empty response with a 200 OK status code when the browser requests the favicon.ico file. Then for test purpopses we add the lines to config/encvironments/test.rb: # Add config/routes_test.rb to routes config.paths['config/routes.rb'] << Rails.root.join('config/routes_test.rb')

  - The other outstanding errors arose because in the test environment we sometimes encounter errors like (Selenium::WebDriver::Error::ElementClickInterceptedError: element click intercepted: # Element <input type="submit" name="commit" value="Change affiliation" class="btn btn-secondary" data-disable-with="Change affiliation"> # is not clickable at point (101, 203). Other element would receive the click: <div id="ui-id-2" tabindex="-1"
class="ui-menu-item-wrapper">...</div>).
    To get round this we use JS to click on element using code like
this:
  change_affiliation_input_button = find('input[value="Change affiliation"]')
   execute_script('arguments[0].click();', change_affiliation_input_button)


benjaminfaure (approved):

johnpinto1:
@aaronskiba You fix was neater than mine for the Selenium::WebDriver::Error::ElementClickInterceptedError. Yours is future proof.

gjacob24 (approved):

PR #3406, opened 2024-04-09
TEMP Test rspec fix

aaronskiba:
Testing if RSpec tests will pass before adding to branch.


aaronskiba:
I just wanted to verify the MySQL and PostgreSQL workflows would pass.

PR #3407, opened 2024-04-10
Fix for improving user search.

johnpinto1:
Currently Searches of users using search "firstname_<empty_spaces>_surname" will only get valid matches if the _<empty_spaces>_ is a single empty spaces.

Change:
The search term string is squished to remove extra empty spaces. As a
result search terms like "Jill Bloggs" (one space between) and "Jill    Bloggs" (more than one space between) will both return the same results.

Reference:
https://api.rubyonrails.org/classes/String.html#method-i-squish


aaronskiba (commented):
Looks like it's working as it should. I wonder if we should apply this to other searches? For example, the same issue exists within the "Plans" search box:
Screenshot from 2024-06-17 12-20-51
Screenshot from 2024-06-17 12-20-56

briri:
Yes, if you see it elsewhere then it should also be patched. I am not going to have time to do that in the next day or so, so feel free to update in this branch if you want. :)

aaronskiba:
> Yes, if you see it elsewhere then it should also be patched. I am not going to have time to do that in the next day or so, so feel free to update in this branch if you want. :)

Cool. @johnpinto1 would you want me to do that? Or would you prefer to take care of it yourself?

johnpinto1:
> > Yes, if you see it elsewhere then it should also be patched. I am not going to have time to do that in the next day or so, so feel free to update in this branch if you want. :)
> 
> Cool. @johnpinto1 would you want me to do that? Or would you prefer to take care of it yourself?

Please do @aaronskiba. Cheers!

aaronskiba:
> > > Yes, if you see it elsewhere then it should also be patched. I am not going to have time to do that in the next day or so, so feel free to update in this branch if you want. :)
> > 
> > 
> > Cool. @johnpinto1 would you want me to do that? Or would you prefer to take care of it yourself?
> 
> Please do @aaronskiba. Cheers!

I was working on this and discovered a bit of an edge-case that complicates this PR a bit:

The issue is when the string being searched (User.firstname + User.surname, Org.name, etc.) actually has double spaces built-in.

```SQL
SELECT firstname, surname
FROM users
WHERE firstname LIKE '% '
OR surname LIKE ' %';
```
When I execute the above query on a recent DMP Assistant DB dump from production, I get 307 returned results.

As a single example, there is a User with first name "Susan  " and last name "Brown".

Searches for both `"Susan Brown"` (one space) or `"Susan  Brown"` (two spaces) both return no results:
Screenshot from 2024-06-26 10-44-19
Screenshot from 2024-06-26 10-44-28

However, when I revert the changes from this PR, the User is returned when I search `"Susan  Brown"` (two spaces):
Screenshot from 2024-06-26 10-45-00

I like the idea behind this PR, but to avoid this edge-case, I'm thinking that in addition to the search string, we'll have to apply the same `.squish` to the records being queried.

johnpinto1:
> I like the idea behind this PR, but to avoid this edge-case, I'm thinking that in addition to the search string, we'll have to apply the same `.squish` to the records being queried.

@aaronskiba Well spotted,  sounds good to me.

PR #3420, opened 2024-05-21
Fix handling of template.visibility

### LINKED ISSUE #3419: Template.visibililty Checks Not Always Working as Expected

- `v4.2.0`

**Expected behaviour:**
- `app/views/org_admin/templates/_form.html.erb` includes the following:
 ```ruby
 <%= f.check_box(:visibility, checked: f.object.visibility == 'organisationally_visible') %>
```
-  `app/views/org_admin/templates/_show.html.erb` includes the following:
```ruby
<% if template.visibility == 'organisationally_visible' %>
```
You'd expect these checks to verify the value of template.visibility

**Actual behaviour:**
```
3.0.5 :001 > Plan.visibilities
 => {"organisationally_visible"=>0, "publicly_visible"=>1, "is_test"=>2, "privately_visible"=>3} 
3.0.5 :002 > Plan.first.visibility
  Plan Load (1.2ms)  SELECT "plans".* FROM "plans" ORDER BY "plans"."id" ASC LIMIT $1  [["LIMIT", 1]]
 => "privately_visible" 
3.0.5 :003 > Template.visibilities
 => {"organisationally_visible"=>0, "publicly_visible"=>1} 
3.0.5 :004 > Template.first.visibility
  Template Load (1.2ms)  SELECT "templates".* FROM "templates" ORDER BY "templates"."id" ASC LIMIT $1  [["LIMIT", 1]]
 => 0 
```
While `plan.visibility` outputs the string representation of the enum value, `template.visibilty` outputs the integer. As a result, the aforementioned `== 'organisationally_visibile'` checks will always return false.

aaronskiba:
Fixes #3419

Changes proposed in this PR:
- app/views/org_admin/templates/_show.html.erb
  - Prior to this commit, `if template.visibility == 'organisationally_visible'`, would always evaluate to false. This is because `template.visibility` returns an integer value.

- app/views/org_admin/templates/_form.html.erb
  - `f.object.visibility == 'organisationally_visible'` always evaluates to false. Thus, prior to this commit, the checkbox would always initially render as unchecked.
  - Also, prior to this commit, the default checked/unchecked values were used (i.e. "1" would be returned when checked, and "0" would be returned when unchecked), and the box is meant to be checked when selecting 'organisationally_visible' ('for internal %{org_name} use only'), which makes the default checked/unchecked values opposite to the mapping of our enums (i.e. `{"organisationally_visible"=>0, "publicly_visible"=>1}`).

### DIFF
--- a/app/views/org_admin/templates/_form.html.erb
+++ b/app/views/org_admin/templates/_form.html.erb
@@ -20,7 +20,11 @@
                          placement: 'right' }%>
     <div class="form-check">
       <%= f.label(:visibility) do %>
-        <%= f.check_box(:visibility, checked: f.object.visibility == 'organisationally_visible') %>
+        <%= f.check_box(:visibility,
+                        {},
+                        f.object.class.visibilities[:organisationally_visible],
+                        f.object.class.visibilities[:publicly_visible])
+                        %>
 
         <%= _('for internal %{org_name} use only') % { org_name: f.object.org.name } %>
       <% end %>
--- a/app/views/org_admin/templates/_show.html.erb
+++ b/app/views/org_admin/templates/_show.html.erb
@@ -31,7 +31,7 @@
     <!-- If the Org is a funder and another org type then allow then to set the visibility -->
     <dt><%= _('Visibility') %></dt>
     <dd>
-      <% if template.visibility == 'organisationally_visible' %>
+      <% if template.organisationally_visible? %>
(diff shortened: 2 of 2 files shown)


aaronskiba:
Closing this because the behaviour of `Template.visibility` will change with the upgrade to Rails 7. The changes within this PR have been adapted to the following PR: https://github.com/DMPRoadmap/roadmap/pull/3435.

PR #3428, opened 2024-06-17
Update to answers and plans controllers and added an index to the set…

briri:
Some relatively minor changes that provide significant performance improvements:
- Add an index to the `settings` table
- Update queries on the plans_controller and answers_controller.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -505,6 +505,7 @@ ActiveRecord::Schema.define(version: 2022_03_15_104737) do
     t.string "target_type"
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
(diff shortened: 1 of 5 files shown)


db/schema.rb
```diff
@@ -505,6 +505,7 @@
     t.string "target_type"
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
+    t.index ["target_id", "target_type"], name: "settings_target"
```
aaronskiba, line 508: Should there be a corresponding migration file for this?

briri: yes 💯

briri: just added one

PR #3441, opened 2024-07-29
Fix for bug in Plan duplication results in original plan identifier b…

johnpinto1:
…eing copied.

Change in class method Plan.deep_copy:
  - Firstly, on duplicating the Plan we set the plan identifier to nil and save.
  - Then we fill the identifier variable with the Plan id that was regenerated when the duplicate copy was save in the previous.
  - We then persist this change by saving the Plan again.


app/models/plan.rb
```diff
@@ -254,7 +254,12 @@ def self.deep_copy(plan)
     plan_copy = plan.dup
     plan_copy.title = "Copy of #{plan.title}"
     plan_copy.feedback_requested = false
+    # Don't copy the identifier as it will be regenerated
+    plan_copy.identifier = nil
     plan_copy.save!
+    # Copy newly generated Id to the identifier
+    plan_copy.identifier = plan_copy.id
+    plan.save!
```
aaronskiba, line 262: Looks good. I tested the PR and it all appears to be behaving as it should. :) I just have a couple of thoughts as well.

The last two lines are having the final say in the saved value of `plan_copy.identifier`. So assigning  `plan_copy.identifier = nil` prior to the first save doesn't seem necessary.

Also, `plan_copy.identifier` is a String, but `plan_copy.id` is an integer. I guess we have similar code in https://github.com/DMPRoadmap/roadmap/blob/main/app/controllers/plans_controller.rb#L140  (`@plan.identifier = @plan.id.to_s`). However, I guess Rails is automatically typecasting the value to a String here.

johnpinto1: Thanks @aaronskiba. I have made the changes suggested in a separate commit.

aaronskiba (approved):
Looks good. 👍

benjaminfaure (approved):

PR #3449, opened 2024-09-04
Address Deprecation Warnings

aaronskiba:
This PR addresses the following deprecation warnings (the warnings were encountered when executing `bundle exec rspec`):

1. ```
    DEPRECATION WARNING: Setting action_dispatch.show_exceptions to false is deprecated. Set to :none instead. 
    (called from rescue in call at /usr/share/rvm/gems/ruby-3.0.5@upstream/gems/actionpack- 
    7.1.3.4/lib/action_dispatch/middleware/debug_exceptions.rb:43)
    ```
2.  Added `coder:` and `type:` keywords in various places

     Example warning (before adding `type: ` keyword in `app/models/user.rb`:
     ```
     Please pass the class as a keyword argument:

     serialize :prefs, type: Hash
     (called from <class:User> at /path/to/app/models/user.rb:73)
     DEPRECATION WARNING: Passing the class as positional argument is deprecated and will be removed in Rails 
     7.2.
     ```

### DIFF
--- a/app/models/condition.rb
+++ b/app/models/condition.rb
@@ -28,9 +28,9 @@
 class Condition < ApplicationRecord
   belongs_to :question
   enum action_type: %i[remove add_webhook]
-  serialize :option_list, Array
-  serialize :remove_data, Array
-  serialize :webhook_data, JSON
+  serialize :option_list, type: Array
+  serialize :remove_data, type: Array
+  serialize :webhook_data, coder: JSON
 
   # Sort order: Number ASC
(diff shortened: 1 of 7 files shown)


aaronskiba:
Some of the PostgreSQL tests are failing, but that is independent of these PR changes. Those failing tests are addressed here in the following PR: https://github.com/DMPRoadmap/roadmap/pull/3451

benjaminfaure (approved):

PR #3455, opened 2024-09-24
Updated Rack Attack configuration to address vulnerabilities in password updates

johnpinto1:
Changes:
    The fix involves adding a new Rack Attack rule "profile_updates/ip" and
    rewriting the body of the rules "password_resets/ip" and "logins/ip" so
    the the request ip is returned if the rule is triggered.

To Test:

- Make 4+  Forgot Password requests in less than 30 secs.
Selection_219

- Make 4+ Password updates using the User Profile in 30 secs.

Selection_218

You should get the following message with each test above.
Selection_217

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Changelog
 
+- Updated Rack Attack configuration to address vulnerabilities in password updates.
+
 ## v4.2.0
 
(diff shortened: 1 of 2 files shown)


johnpinto1:
Please hold on this I see I must fix tests and I found a bug.

johnpinto1:
Withdrawing until fixed issues mentioned earlier.

PR #3461, opened 2024-10-25
Fix for bugs in the Conditional Questions functionality:

johnpinto1:
- In case of conditional question with checkbox answers the removed
       questions were not being removed from view, nor was the answer to
    these questions (which persisted in the db.

    Changes:
    - Fixed the broken functionality in the method remove_answers_list in app/helpers/conditions_helper.rb.
    - Removed and destroyed the answers of the removed questions.
    - Updated RSpec tests for Conditional Questions functionality.

Reason for this fix: 

> An university have reported that conditional questions don't work as expected in template for one of their templates.
> 
> What happens:
> 
>     Conditions for question 18 don't work.
>     In the past, conditions for questions Conditions for questions 8a failed too. On the day we met with them, conditions for 8a worked.
> 
> What should happen:
> 
>     User says: 'When Q.18 is answered ‘None of the above applies’, Q19, 20 & 21 should go, but they are still there'. Note that the behaviour of the questions is very random. On the day met with them questions 20 and 21 had gone, only 19 was still present.
> 
> Note: the architecture of conditions within the template is that some conditions include questions that also have conditions attached to them (conditional questions within conditional questions).
> 
> We asked them to send us a list of all the conditions within their template so that we can examine the behaviour. Document attached.
> 
>

### DIFF
--- a/app/controllers/answers_controller.rb
+++ b/app/controllers/answers_controller.rb
@@ -99,8 +99,15 @@ class AnswersController < ApplicationController
       template = @section.phase.template
 
       remove_list_after = remove_list(@plan)
-
       all_question_ids = @plan.questions.pluck(:id)
+
+      # Destroy all answers for removed questions
+      remove_list_after.each do |id|
+        Answer.where(question_id: id, plan: @plan).each do |a|
+          Answer.destroy(a.id)
+        end
+      end
+
       # rubocop pointed out that these variable is not used
       # all_answers = @plan.answers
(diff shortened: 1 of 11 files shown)


johnpinto1:
@aaronskiba & @benjaminfaure My bad. I did not check my code changes with Rubocop. Will fix.

johnpinto1:
Will fix issues and create a new PR.

PR #3462, opened 2024-10-25
Fix for bugs in the Conditional Questions functionality:

johnpinto1:
In the case of a conditional question with answers that removed questions, any answers of removed questions was not removed. Nor were the removed answers deleted in the database.

    **Changes:**
    - Fixed the broken functionality in the method remove_answers_list in app/helpers/conditions_helper.rb.
    - Removed and destroyed the answers of the removed questions.
    - Updated RSpec tests for Conditional Questions functionality.

**The context for why this fix was created:**

An university have reported that conditional questions don't work as expected in template for one of their templates.
**What happens:**
    Conditions for question 18 don't work.
    In the past, conditions for questions Conditions for questions 8a failed too. On the day we met with them, conditions for 8a worked.
**What should happen:**
    User says: 'When Q.18 is answered ‘None of the above applies’, Q19, 20 & 21 should go, but they are still there'. Note that the behaviour of the questions is very random. On the day met with them questions 20 and 21 had gone, only 19 was still present.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+- Fix for bug in Conditional questions functionality. In the case of a conditional question with answers that removed questions, any answers of removed questions was not removed. Nor were the removed answers deleted in the database. The Rspec tests were also updated to reflect this change.
+
+
 ## v4.2.0
 
 **Note this upgrade is mainly a migration from Bootstrap 3 to Bootstrap 5.** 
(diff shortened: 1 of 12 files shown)


johnpinto1:
I noticed github-action suggest breaking PR. That is not sensible in this case. Separating the Rspec tests from the main code changes would break the previous RSpec tests.

johnpinto1:
Hold on this PR as I have found a bug that needs fixing. So closing.

PR #3466, opened 2024-10-31
Remove Auto-Generated TinyMCE Skins and Add `public/tinymce/skins/` to `.gitignore`

aaronskiba:
Changes proposed in this PR:
- Remove auto-generated TinyMCE skins from `public/tinymce/skins/`
- Add `public/tinymce/skins/` to `.gitignore`


johnpinto1 (approved):
I really like this fix. I am surprised no one thought about it before. Those tinymce files  removed created noise in codebase that one had to be careful to ignore.

PR #3470, opened 2024-12-16
The V1 API Docs link in Edit profile > API access tab was broken.

johnpinto1:
Updated URL in config.x.application.api_documentation_urls property in config/initializers/_dmproadmap.rb.
Selection_029


johnpinto1:
BTW the link property  is also used in the app/mailers/user_mailer.rb.
https://github.com/DMPRoadmap/roadmap/blob/fc047ac0ae1b10f9fd451909e11a2d31e50b3ada/app/mailers/user_mailer.rb#L210

PR #3496, opened 2025-03-26
Match `config.load_defaults` Version to Rails Version

aaronskiba:
Changes proposed in this PR:
- Updated `config.load_defaults` value to `7.1` to match our Rails version.
- Added explicit `coder: YAML` to `serialize` calls to conform with Rails 7.1 changes
  - (Prior to Rails 7.1, `config.active_record.default_column_serializer = YAML` was implicit, so these explicit calls were unnecessary.)

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -70,7 +70,7 @@ class User < ApplicationRecord
 
   ##
   # User Notification Preferences
-  serialize :prefs, type: Hash
+  serialize :prefs, type: Hash, coder: YAML
 
   # default user language to the default language
   attribute :language_id, :integer, default: -> { Language.default&.id }
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -68,10 +68,6 @@ class User < ApplicationRecord
          :rememberable, :trackable, :validatable, :omniauthable,
          omniauth_providers: %i[shibboleth orcid]
 
-  ##
-  # User Notification Preferences
-  serialize :prefs, type: Hash, coder: YAML
-
   # default user language to the default language
   attribute :language_id, :integer, default: -> { Language.default&.id }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-- Updated app to Rails 7 [#3426](https://github.com/DMPRoadmap/roadmap/pull/3426)
+- Updated app to Rails 7 [#3426](https://github.com/DMPRoadmap/roadmap/pull/3426), [#3496](https://github.com/DMPRoadmap/roadmap/pull/3496)
 - Address Some Bullet Warnings / Optimise Mean Request Times [#3440](https://github.com/DMPRoadmap/roadmap/pull/3440)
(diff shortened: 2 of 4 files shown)


johnpinto1:
@aaronskiba I wonder if this file  https://github.com/rails/rails/blob/v7.0.0/railties/lib/rails/generators/rails/app/templates/config/initializers/new_framework_defaults_7_0.rb.tt (with .tt remove) should be included in config/initializers for completenes. So config commented out config can be switched on if necessary.

johnpinto1:
@aaronskiba The Condition helper coder: YAML will need to be be overridden by PR https://github.com/CDLUC3/dmptool/pull/667 (or the Conditional questions will fail as we moved from YAML to JSON.
https://github.com/DMPRoadmap/roadmap/blob/76cc7eb71fd998fcb7f5d2427e9c9a3a1e822f28/app/models/condition.rb#L29-L33
Also this change is incorrect it should 
**serialize :prefs, coder: JSON**
https://github.com/DMPRoadmap/roadmap/blob/76cc7eb71fd998fcb7f5d2427e9c9a3a1e822f28/app/models/user.rb#L72-L74

aaronskiba:
> @aaronskiba The Condition coder: YAML will need to be be overridden by PR [CDLUC3#667](https://github.com/CDLUC3/dmptool/pull/667) (or the Conditional questions will fail as we moved from YAML to JSON.
> 
> https://github.com/DMPRoadmap/roadmap/blob/76cc7eb71fd998fcb7f5d2427e9c9a3a1e822f28/app/models/condition.rb#L29-L33
> 
> 
> Also this change is incorrect it should (as is currently)
> **serialize :prefs, type: Hash, coder: JSON**
> https://github.com/DMPRoadmap/roadmap/blob/76cc7eb71fd998fcb7f5d2427e9c9a3a1e822f28/app/models/user.rb#L72-L74

Thank you. Yes, the PR you mentioned will migrate those serialized columns from YAML to JSON. We'll have to be sure we end up with the JSON type in the end.

I'm wondering what to do with the other code you mentioned:
```ruby
 # User Notification Preferences 
 serialize :prefs, type: Hash, coder: YAML
```
The `prefs` column isn't actually part of the `users` table. Based on that, I don't think the code serves any purpose.

Here's a bit of additional information too:
```
rb(main):0015> user.pref
  Pref Load (0.4ms)  SELECT "prefs".* FROM "prefs" WHERE "prefs"."user_id" = $1 LIMIT $2
```
```ruby
class Pref < ApplicationRecord
  ##
  # Serialize prefs to JSON
  serialize :settings, coder: JSON
```
`has_one  :pref` exists in the User model. `user.pref` loads the data from that table and `settings` is correctly set to JSON.

app/models/user.rb
```diff
@@ -70,7 +70,7 @@ class User < ApplicationRecord
 
   ##
   # User Notification Preferences
-  serialize :prefs, type: Hash
+  serialize :prefs, type: Hash, coder: YAML
```
aaronskiba, line 73: Hi @briri, does it look like this code could be removed entirely? I can see an older migration that moved prefs to its own table, so I'm thinking this code no longer serves a purpose. (https://github.com/DMPRoadmap/roadmap/blob/main/db/migrate/20170712084314_move_prefs_to_table.rb).

PR #3476, opened 2025-01-30
Fix issues with Conditional question serialization (offered by @briri from DMPTool)

johnpinto1:
Fix issues with Conditional question serialization (offered by @briri from DMPTool) in PR https://github.com/CDLUC3/dmptool/pull/667

Changes proposed in this PR (text from cited PR):

-  There is a migration file with code for MySQL and Postgres to update the Conditions table to convert JSON Arrays in string format records in the conditions table so that they are JSON Arrays.
- Updated the form partials in app/views/org_admin/conditions/_form.erb.rb to properly send condition data to the controller.      
- Removed all JSON.parse calls in the app/helpers/conditions.rb helper 
- The user can no longer edit a condition. They need to remove it and create a new condition. This applies to the email for 'add notifications' too.

Made the following changes to simplify this patch and to make it a little more user friendly:
- The "Add Conditions" button will now say "Edit Conditions" if there are any conditions for a given question.
- Updated the column heading over the "thing that happens when the condition is met" from "Remove" to "Target" since the content of the column can either be questions being removed or an email notification being sent. 
- Conditions of any action type can be added or removed (not updated anymore). 
- Hovering over the email of an existing condition displays a tooltip that shows the email message, subject, etc. The email content can no longer be edited once saved. It will need to be removed and re-created.   
- We allow only one question option to be selected when adding a Condition unlike inthe DMPTool version because experience with multiple options chosen has been problematic and buggy when used by users in DMPOnline. 
 - To add a condition you must have selected an Option and Action together with: 
    -- If Action is 'remove', you need to select one or more choices in Target. 
    -- If Action is 'add notification', you need to fill in all the fields in the 'Send email' popup. 
    Otherwise, the condition will not be saved.
    
Some screenshots:
Selection_073
Selection_075
Selection_076
Selection_077

PR #3497, opened 2025-03-27
Fix issues with Conditional question serialization (offered by @briri from DMPTool) in PR https://github.com/CDLUC3/dmptool/pull/667

johnpinto1:
_Previous PR closed as it was easier to add code to a new pull of development. Old PR https://github.com/DMPRoadmap/roadmap/pull/3476 closed_

Fix issues with Conditional question serialization (offered by @briri from DMPTool) in PR https://github.com/CDLUC3/dmptool/pull/667

Changes proposed in this PR (text from cited PR):

-  There is a migration file with code for MySQL and Postgres to update the Conditions table to convert JSON Arrays in string format records in the conditions table so that they are JSON Arrays.
- Updated the form partials in app/views/org_admin/conditions/_form.erb.rb to properly send condition data to the controller.      
- Removed all JSON.parse calls in the app/helpers/conditions.rb helper 
- The user can no longer edit a condition. They need to remove it and create a new condition. This applies to the email for 'add notifications' too.

Made the following changes to simplify this patch and to make it a little more user friendly:
- The "Add Conditions" button will now say "Edit Conditions" if there are any conditions for a given question.
- Updated the column heading over the "thing that happens when the condition is met" from "Remove" to "Target" since the content of the column can either be questions being removed or an email notification being sent. 
- Conditions of any action type can be added or removed (not updated anymore). 
- Hovering over the email of an existing condition displays a tooltip that shows the email message, subject, etc. The email content can no longer be edited once saved. It will need to be removed and re-created.   
- We allow only one question option to be selected when adding a Condition unlike inthe DMPTool version because experience with multiple options chosen has been problematic and buggy when used by users in DMPOnline. 
 - To add a condition you must have selected an Option and Action together with: 
    -- If Action is 'remove', you need to select one or more choices in Target. 
    -- If Action is 'add notification', you need to fill in all the fields in the 'Send email' popup. 
    Otherwise, the condition will not be saved.
    
Some screenshots:
Selection_073
Selection_075
Selection_076
Selection_077

### DIFF
--- a/app/controllers/org_admin/questions_controller.rb
+++ b/app/controllers/org_admin/questions_controller.rb
@@ -211,17 +211,21 @@ module OrgAdmin
 
     private
 
-    # param_conditions looks like:
-    #   [
-    #     {
-    #       "conditions_N" => {
-    #         name: ...
-    #         subject ...
-    #         ...
-    #       }
-    #       ...
-    #     }
-    #   ]
+    # param_conditions is a hash of a hash like this (example where
+    # action_types is "remove" and "add_webhook" respectively):
+    # Parameters:
+    # {"0"=>{"question_option"=>["212159"],
+    #        "action_type"=>"remove",
+    #        "remove_question_id"=>["191471 191472"],
+    #        "number"=>"0"},
+    # "1"=>{"question_option"=>["212160"],
+    #       "action_type"=>"add_webhook",
+    #        "number"=>"1",
+    #        "webhook-name"=>"DMP Admin",
+    #        "webhook-email"=>"dmp-admin@example.com",
+    #        "webhook-subject"=>"Woodcote cillum quis elit consectetur epsom",
+    #        "webhook-message"=>"Labore ut epsom downs exercitation ...."}
+    #  }
     def sanitize_hash(param_conditions)
       return {} if param_conditions.nil?
       return {} if param_conditions.empty?
--- a/app/controllers/org_admin/questions_controller.rb
+++ b/app/controllers/org_admin/questions_controller.rb
@@ -227,14 +227,12 @@ module OrgAdmin
       return {} if param_conditions.empty?
 
       res = {}
-      hash_of_hashes = param_conditions[0]
-      hash_of_hashes.each do |cond_name, cond_hash|
+      param_conditions.each do |cond_id, cond_hash|
         sanitized_hash = {}
         cond_hash.each do |k, v|
-          v = ActionController::Base.helpers.sanitize(v) if k.start_with?('webhook')
-          sanitized_hash[k] = v
+          sanitized_hash[k] = k.start_with?('webhook') ? ActionController::Base.helpers.sanitize(v) : v
         end
-        res[cond_name] = sanitized_hash
+        res[cond_id] = sanitized_hash
       end
       res
     end
--- a/app/models/condition.rb
+++ b/app/models/condition.rb
@@ -30,7 +30,7 @@ class Condition < ApplicationRecord
   enum :action_type, { remove: 0, add_webhook: 1 }
   serialize :option_list, type: Array, coder: JSON
   serialize :remove_data, type: Array, coder: JSON
-  serialize :webhook_data, type: Hash, coder: JSON
+  serialize :webhook_data, coder: JSON
 
   # Sort order: Number ASC
   default_scope { order(number: :asc) }
--- a/app/models/condition.rb
+++ b/app/models/condition.rb
@@ -27,10 +27,10 @@
 # Object that represents a condition of a conditional question
 class Condition < ApplicationRecord
   belongs_to :question
-  enum action_type: %i[remove add_webhook]
-  serialize :option_list, type: Array
-  serialize :remove_data, type: Array
-  serialize :webhook_data, coder: JSON
+  enum :action_type, { remove: 0, add_webhook: 1 }
+  serialize :option_list, type: Array, coder: JSON
+  serialize :remove_data, type: Array, coder: JSON
+  serialize :webhook_data, type: Hash, coder: JSON
 
   # Sort order: Number ASC
   default_scope { order(number: :asc) }
--- a/app/models/question.rb
+++ b/app/models/question.rb
@@ -145,7 +145,7 @@ class Question < ApplicationRecord
     guidances = {}
     if theme_ids.any?
       GuidanceGroup.includes(guidances: :themes)
-                   .where(org_id: org.id).each do |group|
+                   .where(org_id: org.id).find_each do |group|
         group.guidances.each do |g|
           g.themes.each do |theme|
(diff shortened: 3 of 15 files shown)


johnpinto1:
This was added as Rails 7 update was breaking for editing template and adding conditional question
Selection_007
Selection_006

app/models/condition.rb
```diff
@@ -27,10 +27,10 @@
...
 class Condition < ApplicationRecord
   belongs_to :question
-  enum action_type: %i[remove add_webhook]
-  serialize :option_list, type: Array
-  serialize :remove_data, type: Array
-  serialize :webhook_data, coder: JSON
+  enum :action_type, { remove: 0, add_webhook: 1 }
+  serialize :option_list, type: Array, coder: JSON
+  serialize :remove_data, type: Array, coder: JSON
+  serialize :webhook_data, type: Hash, coder: JSON
```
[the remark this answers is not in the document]
johnpinto1: @aaronskiba My bad @briri had removed the type: Hash that I missed. I have pushed the change.

app/controllers/org_admin/questions_controller.rb
aaronskiba, line 1: This code change now treats `param_conditions` itself as the hash of hashes. It seems to work, but we should probably update the code comment on line 214 to the following to reflect that:

```ruby
    # param_conditions looks like:
    #     {
    #       "conditions_N" => {
    #         name: ...
    #         subject ...
    #         ...
    #       }
    #       ...
    #     }
```

johnpinto1: @aaronskiba Update comment as requested.

app/views/org_admin/questions/_form.html.erb
```diff
@@ -45,16 +45,18 @@
         <%= render "/org_admin/question_options/option_fields", f: f, q: question %>
         <!--display for selecting comment box when multiple choice is being used-->
       </div>
+    
       <% if question.id != nil && question.question_options[0].text != nil %>
-        <%= link_to _('Add Conditions'), org_admin_question_open_conditions_path(question_id: question.id, conditions: conditions), class: "add-logic btn btn-secondary", 'data-loaded': (conditions.size > 0).to_s, remote: true %>
+        <% cond_lbl = (!conditions.nil? && conditions.any?) ? 'Edit Conditions' : 'Add Conditions' %>
```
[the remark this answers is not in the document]
johnpinto1: @aaronskiba Updated code as requested https://github.com/DMPRoadmap/roadmap/pull/3497/commits/a58017c934aaf742e8be9b9e0843f0192bd119fe

app/views/org_admin/conditions/_form.html.erb
```diff
@@ -1,35 +1,88 @@
+<% condition = condition ||= nil %>
```
aaronskiba, line 1: We could just refactor this to `<% condition ||= nil %>`

johnpinto1: @aaronskiba Update code with change requested https://github.com/DMPRoadmap/roadmap/pull/3497/commits/06b0db5326c50be6bfcb0ac9a8d8401f29f41971

app/models/question.rb
```diff
@@ -221,28 +221,47 @@ def save_condition(value, opt_map, question_id_map)
...
-        opt_map[qopt]
+
+    if opt_map.present?
+      new_question_options = []
+      c.option_list.map do |qopt|
+        new_question_options << opt_map[qopt]
       end
-      c.option_list = new_question_options || []
+      c.option_list = new_question_options
     end
```
aaronskiba, line 231: This refactor is an improvement, but I think it could even be taken further. Maybe it's over the top but I will publish a PR that attempts to do a bigger refactor of this `save_condtion` method.

aaronskiba: https://github.com/DMPRoadmap/roadmap/pull/3501

app/views/org_admin/conditions/_form.html.erb
```diff
@@ -1,35 +1,88 @@
...
-      <%= select_tag(:remove_question_id, remove_question_group, {name: name_start + "[remove_question_id][]", class: 'form-select regular', multiple: true, 'data-bs-style': 'dropdown-toggle bg-white px-4 py-3'}) %>
+  
+  <%# If this is a new condition then display the interactive controls. otherwise just display the logic %>
+  <% if condition.nil? %>
+  <div class="form-label bold">Add condition</div>
+  <div class="row  mb-3">
+    <div class="col-md-9 pe-2">
+      <div class="form-label bold"><%= _('Option') %></div>
+      <%= select_tag(:question_option, options_from_collection_for_select(question.question_options.sort_by(&:number), "id", "text",
+          condition_exists ? condition[:question_option_id] : question.question_options.sort_by(&:number)[0]), {class: 'form-select regular', 'data-bs-style': 'dropdown-toggle bg-white px-4 py-3', name: name_start + "[question_option][]"}) %>
```
aaronskiba, line 25:
```ruby
condition_exists ? condition[:question_option_id] : question.question_options.sort_by(&:number)[0])
```
I am wondering about this. This code is only executed when `if condition.nil?` is true. Is there there possibility of a NoMethodError here? Or can we be sure `condition_exists` is always false and replace `condition_exists ? condition[:question_option_id] : question.question_options.sort_by(&:number)[0])` with `question.question_options.sort_by(&:number)[0])`?

aaronskiba: Like the Question model, there's a bit to refactor here. I might do one more PR.

app/views/org_admin/conditions/_form.html.erb
```diff
@@ -1,35 +1,88 @@
+<% condition ||= nil %>
```
aaronskiba, line 1: `<% condition ||= nil %>` is a good addition and I think the code could be updated to use it better.

`condition_exists = local_assigns.has_key? :condition` now seems redundant as a result. We can just use condition.presence? instead.

app/views/org_admin/conditions/_new_condition_form.erb
```diff
@@ -0,0 +1,34 @@
+  <%
+    action_type_arr = [["removes", :remove], ["adds notification", :add_webhook]]
+    remove_question_collection = later_question_list(question)
+    remove_question_group = grouped_options_for_select(remove_question_collection)
+  %>
+
+  <div class="form-label bold">Add condition</div>
```
aaronskiba, line 7: We should enable translations here too.

app/views/org_admin/conditions/_add.html.erb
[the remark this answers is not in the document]
johnpinto1: @aaronskiba Pls add code and merge.

aaronskiba (approved):
My last commit enables a few more translations. I am approving this PR.

PR #3503, opened 2025-04-08
Fix failing eslint workflow / upgrade `actions/checkout` & `actions/setup-node` to v3

aaronskiba:
Changes proposed in this PR:
- This change is intended to fix the following GitHub Action failure: https://github.com/DMPRoadmap/roadmap/actions/runs/14087197077
- Upgraded upgrade actions/checkout & actions/setup-node to v3


aaronskiba:
The eslint workflow appears to be working again without this change.

PR #3507, opened 2025-04-14
Re-Implement Email Confirmation

aaronskiba:
Changes proposed in this PR:
- Added `:confirmable` module to User model, which implements email confirmation via Devise.

- Added the rake task, `email_confirmation:clear_all` (see `lib/tasks/email_confirmation.rake`)
  - The rake task sets the values of all email confirmation-related columns (`confirmed_at`, `confirmation_token`, and `confirmation_sent_at`) to nil for all users. It then proceeds to confirm all superusers within the app.

- Streamline the email confirmation process for existing users
  - By default, Devise's `:confirmable` module generates a `confirmation_token` and auto-sends confirmation instructions when a new user is created.
  - Because we are only implementing `:confirmable` now, existing users can't receive these autosent instructions. However, this PR implements autosent confirmation instructions in the following manner:
    1. A user attempts to sign into the app. (Note, this sign-in can either be performed via the system sign-in, or via the Shibboleth sign-in).
    2. The `return if confirmation_instructions_missing_and_handled?(user)` line of code is executed. This method belongs to the `EmailConfirmationHandler` concern, and works as follows:
         i) returns `false` if the user is either already confirmed or has an outstanding confirmation_token
         ii) Else (the user is unconfirmed AND has no outstanding confirmation_token). Generate the confirmation token and auto-send the confirmation instructions email. (Note: on subsequent sign-in attempts, attempts, the method will return `false`, preventing redundant emails.)

- Customise `devise.failure.unconfirmed` value in various `config/locales/*.yml` files. The customised value includes an embedded link to `/users/confirmation/new`. The following is a screenshot of the customised value for `:"en-CA"`:
  - Screenshot from 2025-04-23 13-43-21

- Updated existing tests
  - Added `confirmed_at { Time.current }` to User factory
  - Added `config.action_mailer.default_options = { from: 'noreply@example.org' }` to enable email confirmation tests

- Add new tests (spec/features/email_confirmation_spec.rb)
  - These tests verify the streamlined email confirmation behaviour for existing users
    - Quite a bit of config was needed to enable testing of behaviour with shibboleth (see `spec/support/helpers/omniauth_helper.rb` and additions to `spec/rails_helper.rb`)
  - They also test the clickable link to `/users/confirmation/new` embedded in the customised `devise.failure.unconfirmed` flash message.

- Refactoring
  - Addressed some rubocop offences within `SessionsController#create` and `Users::OmniauthCallbacksController#handle_omniauth`

- https://github.com/DMPRoadmap/roadmap/pull/3511
  - Update `spec/support/faker.rb` to replace `'en'` with `I18n.default_locale` for assigning locales. This change should allow for more accurate testing by using the application's specified locale.
  - Prior to this PR, there was a commit in the aforementioned file that read `"Keep this as :en. Faker doesn't have :en-GB"`. However, `I18n.default_locale` evaluates to `:"en-GB"` for this codebase, and all of the tests appear to be passing. Additionally, the branch that this PR is pointed at (`upstream/aaron/add-email-confirmation`), uses `I18n.t(...)` for several tests, and this change is required for those tests to pass.

### DIFF
--- a/config/locales/translation.fr-CA.yml
+++ b/config/locales/translation.fr-CA.yml
@@ -242,7 +242,8 @@ fr-CA:
       not_found_in_database: "%{authentication_keys} ou mot de passe incorrect."
       timeout: Votre session est expirée. Veuillez vous reconnecter pour continuer.
       unauthenticated: Vous devez vous connecter ou vous inscrire pour continuer.
-      unconfirmed: Vous devez confirmer votre adresse courriel pour continuer.
+      unconfirmed: Vous devez confirmer votre compte avant de continuer. <a href="/users/confirmation/new"
+        class="a-orange">(cliquez pour demander un nouveau courriel de confirmation)</a>
       invited: Vous avez une invitation en attente. Acceptez-la pour terminer la création
         de votre compte.
     mailer:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
 - Fix failing eslint workflow / upgrade `actions/checkout` & `actions/setup-node` to v3 [#3503](https://github.com/DMPRoadmap/roadmap/pull/3503)
 - Fix rendering of `confirm_merge` partial [#3515](https://github.com/DMPRoadmap/roadmap/pull/3515)
 - Remove Auto-Generated TinyMCE Skins and Add `public/tinymce/skins/` to `.gitignore` [#3466](https://github.com/DMPRoadmap/roadmap/pull/3466)
+- Re-implement email confirmation [#3507](https://github.com/DMPRoadmap/roadmap/pull/3507)
+- Update `spec/support/faker.rb` to use `I18n.default_locale` [#3507](https://github.com/DMPRoadmap/roadmap/pull/3507)
 
 ## v5.0.0
 
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -19,15 +19,7 @@ class SessionsController < Devise::SessionsController
       end
 
       # Until ORCID login is supported
-      unless session['devise.shibboleth_data'].nil?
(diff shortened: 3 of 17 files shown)


config/locales/translation.fr-CA.yml
```diff
@@ -242,12 +242,13 @@ fr-CA:
...
       unauthenticated: Vous devez vous connecter ou vous inscrire pour continuer.
-      unconfirmed: Vous devez confirmer votre adresse courriel pour continuer.
+      unconfirmed: Vous devez confirmer votre compte avant de continuer. <a href="/users/confirmation/new"
+        class="a-orange">(cliquez pour demander un nouveau courriel de confirmation)</a>
       invited: Vous avez une invitation en attente. Acceptez-la pour terminer la création
         de votre compte.
     mailer:
       confirmation_instructions:
-        subject: Consignes de confirmation
+        subject: Validez votre compte Assistant PGD
```
aaronskiba, line 251: This value is currently hardcoded for DMP Assistant. It will have to be updated.

spec/integration/email_confirmation_spec.rb
```diff
@@ -0,0 +1,86 @@
+# frozen_string_literal: true
```
aaronskiba, line 1: This can be moved to `spec/integration/email_confirmation_spec.rb`

lib/tasks/dmp_assistant_upgrade.rake
```diff
@@ -0,0 +1,56 @@
+# frozen_string_literal: true
```
aaronskiba, line 1: We'll have to update the naming and versioning in here.

johnpinto1 (approved):
I am unable to check Shibb. But if @aaronskiba has it working, then I suggest merging PR.

PR #3513, opened 2025-04-21
Add codespell support (config, workflow to detect/not fix) and make it fix some typos

yarikoptic:
More about codespell: https://github.com/codespell-project/codespell .

I personally introduced it to dozens if not hundreds of projects already and so far only positive feedback.

CI workflow has 'permissions' set only to 'read' so also should be safe.

Inspired by initial attempt in
- https://github.com/CDLUC3/dmptool/pull/428

while listening to a presentation about dmptool.org .


aaronskiba:
Hi [yarikoptic](https://github.com/yarikoptic), thank you for the PR. :) This is interesting and could prove to be useful for us. I suppose the renaming of variables and existing strings could pose issues though. I might point this at the our `development` branch and revert the actual code changes if I end up approving. Hope that's all okay and thank you again.

yarikoptic:
just to make clear: do you want me to rebase/redo these fixes on top of `development` branch instead?

as for variables names etc -- we could whitelist them or annotate inline if only few specific lines, see https://github.com/codespell-project/codespell?tab=readme-ov-file#inline-ignore . Please point which ones you would like?

PR #3529, opened 2025-05-30
Fix handling of `attrs[:managed]` + Refactor `OrgsController#admin_update`

### LINKED ISSUE #3528: BUG: Updating Org Fields Can Set `Org.managed` to false
This issue is documented at https://github.com/portagenetwork/roadmap/issues/1061 but also exists within this codebase.

aaronskiba:
Fixes #3528

Changes proposed in this PR:
- This change resolves issue #3528 by preventing the updating of `@org.managed` when `attrs[:managed]` is missing. Prior to this change, the `if attrs.key?(:managed)` was not present. But it was necessary, because that key is not present when a super user clicks "Save" on the "Request Feedback" page.
- As a result, if `attrs[:managed]` was not present,  then `attrs[:managed] = (attrs[:managed] == '1')` would evaluate false, and `if @org.update(attrs)` (line 81) would subsequently update `@org.managed` to false in the db.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## Unreleased
+- Fix Handling of `attrs[:managed]` + Refactor `OrgsController#admin_update` [#3529](https://github.com/DMPRoadmap/roadmap/pull/3529)
+
 ## v5.0.2
 - Bump Ruby to v3.1.4 and use `.ruby-version` in CI
   - [#3566](https://github.com/DMPRoadmap/roadmap/pull/3566)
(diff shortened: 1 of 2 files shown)


aaronskiba:
Setting as draft. I'd like to refactor this to try to handle some of these validations outside of the controller.

johnpinto1 (approved):
**Super admin's view an Organisation details:**
Refactoring and fix working fine for missing "Managed".

Not sure why "Managed" is cited as a mandatory field. (Is it worthwhile removing the asterisk?).
Also it looks like you can never save if the mandatory field "Contact email" is missing, but you can if "Link text" is missing. I guess this is something added in past, super admins are fine with this convention.

Issue #3561, opened 2025-09-16
Discrepancy in Org Admin view of the Organisation Plans and the related CSV Download


johnpinto1:
There is a discrepancy in Org Admin view of the Organisation Plans and the related CSV Download.
The CSV download shows de-activated plans whilst the view only displays active plans.

The code for both use the same method **org_admin_plans** in the **Org** model
https://github.com/DMPRoadmap/roadmap/blob/bfbb5183a86d9db12edb0a46bde2b8620fddbaca/app/models/org.rb#L287-L302

The paginable method **org_admin** for Org Admin plans by including the filter condition `where(Role.creator_condition)` ensures the view for Org Admins and Super Admins only see active plans (as we set Role of the creator (owner or co-owner who removes plan) to active: false).
https://github.com/DMPRoadmap/roadmap/blob/bfbb5183a86d9db12edb0a46bde2b8620fddbaca/app/controllers/paginable/plans_controller.rb#L40-L59

We think discrepancy between view and download data should be fixed in **Org** model's method  **org_admin_plans** rather than having it done in the paginable method **org_admin** as the former method is used in several places:

**app/models/plan.rb**
172:    plan_ids = user.org.**org_admin_plans**.where(complete: true).pluck(:id).uniq
 
**app/controllers/paginable/plans_controller.rb**
48:      plans = @super_admin ? Plan.all : current_user.org.**org_admin_plans**
 
**app/controllers/api/v0/plans_controller.rb**
75:        org_admin_plans = @user.org.**org_admin_plans**
 
**app/controllers/org_admin/plans_controller.rb**
23:         current_user.org.**org_admin_plans**.page(1)
71:        org.**org_admin_plans**.includes(template: :org).order(updated_at: :desc).each do |plan|

PR #3023, opened 2021-09-02
Issue #496 - Plan search issues.

johnpinto1:
Fixes #496 .
The specification for Plans search and display was cited by @mariapraetzellis  in comment
https://github.com/DMPRoadmap/roadmap/pull/3000#issuecomment-901207072
```
We discussed this in our team meeting. @johnpinto1

My Dashboard - shows Plans the current user's currently has an 'active' role on (creator, co-owner, editor, commenter)

    agreed this is the correct behavior

My Dashboard - shows the organisationally visible plans for the current user
-- agreed this is the correct behavior

Admin -> Plans - shows all of the plans owned/co-owned by one of the Org's users

    agreed this is the correct behavior
    -there is an exception to this in the code that are two configurable settings that allows an instance to set this so private plans do not display for admins or super admins

Admin -> Users - shows the Plans for the user being edited (not sure if these are creator or would show if the user is just an editor or commenter)

    This is the correct behavior but we do have an additional feature to add as described below. We will make a new issue for this additional feature.
    We should add a new column that shows the user's role in each plan. This would also need to adhere to the configuration settings that allows an instance to set this so private plans do not display for admins or super admins.

Public Plans - shows any Plan marked as publicly_visible

    This is the correct behavior

We'd have to look at the API v0 and v1 to see if they use that search method (I think v0 might)

    Please check API v0 & V1 to make sure they're following the configuration settings
```

Further comments by @raycarrick-ed and @briri  are linked here:
https://github.com/DMPRoadmap/roadmap/pull/3000#issuecomment-901706504
https://github.com/DMPRoadmap/roadmap/pull/3000#issuecomment-901975844

Changes proposed in this PR:

     - Replaced .where(Role.creator_condition) with .where(roles: { active: true })
     in model Plan search scope.
     - Added .where(roles: { active: true }) to Plan retrieval in
    org_admin_plans() method of the Org model.  (@briri does this address comment https://github.com/DMPRoadmap/roadmap/pull/3000#issuecomment-901975844, as aPI only available to Org Admins.)

PR #3564, opened 2025-09-25
    Issue #3561 - Fix for bug "Discrepancy in Org Admin view of the

johnpinto1:
Organisation Plans and the related CSV Download".

    **Changes:**
    - where(Role.creator_condition) condition added to Plans retrieved in the Org model's org_admin_plans method. This ensures that Plans returned are only active plans. Note: if an owner or co-owner de-activates a plan the Role of the user is set to active: false, so any plan with creator with role (having active: false) would mean the Plan is removed.
    - To avoid duplication we removed .where(Role.creator_condition) in org_admin method in app/controllers/paginable/plans_controller.rb from line plans = plans.joins(:template, roles: [user::org]).where(Role.creator_condition).
    - Updated RSpec tests for Org method org_admin_plans() in spec/models/org_spec.rb.
    - Update CHANGELOG.

Under the current code an Org plan that is removed by owner or coowner will be visible by the Org admins for the Org in the CSV. The Org Admin, correctly, does not show plan. With this change in code the removed plan is no longer present in the CSV.
**To test:**
Look at a given removed plan before and after code change.

### DIFF
--- a/app/controllers/paginable/plans_controller.rb
+++ b/app/controllers/paginable/plans_controller.rb
@@ -46,13 +46,8 @@ module Paginable
       @super_admin = current_user.can_super_admin?
       @clicked_through = params[:click_through].present?
 
-      plans = if @super_admin
-                Plan.all.joins(:template, roles: [user: :org])
-                    .where(Role.creator_condition)
-              else
-                current_user.org.org_admin_plans
-                            .joins(:template, roles: [user: :org])
-              end
+      plans = @super_admin ? Plan.where(Role.creator_condition) : current_user.org.org_admin_plans
+      plans = plans.joins(:template, roles: [user: :org])
 
       paginable_renderise(
         partial: 'org_admin',
--- a/app/controllers/paginable/plans_controller.rb
+++ b/app/controllers/paginable/plans_controller.rb
@@ -45,8 +45,14 @@ module Paginable
       # check if current user if super_admin
       @super_admin = current_user.can_super_admin?
       @clicked_through = params[:click_through].present?
-      plans = @super_admin ? Plan.all : current_user.org.org_admin_plans
-      plans = plans.joins(:template, roles: [user: :org]).where(Role.creator_condition)
+
+      plans = if @super_admin
+                Plan.all.joins(:template, roles: [user: :org])
+                    .where(Role.creator_condition)
+              else
+                current_user.org.org_admin_plans
+                            .joins(:template, roles: [user: :org])
+              end
 
       paginable_renderise(
         partial: 'org_admin',
--- a/app/models/org.rb
+++ b/app/models/org.rb
@@ -293,11 +293,13 @@ class Org < ApplicationRecord
     if Rails.configuration.x.plans.org_admins_read_all
       Plan.includes(:template, :phases, :roles, :users).where(id: combined_plan_ids)
           .where(roles: { active: true })
+          .where(Role.creator_condition)
     else
       Plan.includes(:template, :phases, :roles, :users).where(id: combined_plan_ids)
           .where.not(visibility: Plan.visibilities[:privately_visible])
           .where.not(visibility: Plan.visibilities[:is_test])
           .where(roles: { active: true })
+          .where(Role.creator_condition)
     end
   end
   # rubocop:enable Metrics/AbcSize
--- a/app/models/org.rb
+++ b/app/models/org.rb
@@ -288,19 +288,17 @@ class Org < ApplicationRecord
   # This replaces the old plans method. We now use the native plans method and this.
   # rubocop:disable Metrics/AbcSize
   def org_admin_plans
-    combined_plan_ids = (native_plan_ids + affiliated_plan_ids).flatten.uniq
+    scope = Plan.includes(:template, :phases, :roles, :users)
+                .where(id: (native_plan_ids + affiliated_plan_ids).uniq)
+                .where(roles: { active: true })
+                .where(Role.creator_condition)
 
-    if Rails.configuration.x.plans.org_admins_read_all
-      Plan.includes(:template, :phases, :roles, :users).where(id: combined_plan_ids)
-          .where(roles: { active: true })
-          .where(Role.creator_condition)
-    else
-      Plan.includes(:template, :phases, :roles, :users).where(id: combined_plan_ids)
-          .where.not(visibility: Plan.visibilities[:privately_visible])
-          .where.not(visibility: Plan.visibilities[:is_test])
-          .where(roles: { active: true })
-          .where(Role.creator_condition)
+    unless Rails.configuration.x.plans.org_admins_read_all
+      scope = scope.where.not(visibility: Plan.visibilities[:privately_visible])
+                   .where.not(visibility: Plan.visibilities[:is_test])
     end
+
+    scope
   end
   # rubocop:enable Metrics/AbcSize
 
--- a/spec/models/org_spec.rb
+++ b/spec/models/org_spec.rb
@@ -344,79 +344,124 @@ RSpec.describe Org, type: :model do
 
   describe '#org_admin_plans' do
     Rails.configuration.x.plans.org_admins_read_all = true
-    let!(:org) { create(:org) }
-    let!(:plan) { create(:plan, org: org, visibility: 'publicly_visible') }
-    let!(:user) { create(:user, org: org) }
+    # Two Orgs
+    let!(:org1) { create(:org) }
+    let!(:org2) { create(:org) }
 
-    subject { org.org_admin_plans }
+    # Plans for org1
+    let!(:org1plan1) { create(:plan, :creator, :organisationally_visible, org: org1) }
+    let!(:org1plan2) { create(:plan, :creator, :privately_visible, org: org1) }
+    let!(:org1plan3) { create(:plan, :creator, :publicly_visible, org: org1) }
+    let!(:org1plan4) { create(:plan, :creator, :organisationally_visible, org: org1) }
+
+    # Plans for org2
+    let!(:org2plan1) { create(:plan, :creator, :organisationally_visible, org: org2) }
+
+    subject { org1.org_admin_plans }
 
     context 'when user belongs to Org and plan owner with role :creator' do
       before do
-        create(:role, :creator, user: user, plan: plan)
-        plan.add_user!(user.id, :creator)
+        Rails.configuration.x.plans.org_admins_read_all = true
       end
-
-      it { is_expected.to include(plan) }
+      it { is_expected.to include(org1plan1, org1plan2, org1plan3, org1plan4) }
+      it { is_expected.not_to include(org2plan1) }
     end
 
-    context 'when user belongs to Org and plan user with role :administrator' do
+    context 'when user belongs to Org and a plan removed by creator assuming there are no coowners' do
       before do
-        plan.add_user!(user.id, :administrator)
+        Rails.configuration.x.plans.org_admins_read_all = true
+        org1plan4.roles.map { |r| r.update(active: false) if r.user_id == org1plan4.owner.id }
+      end
+
+      it { is_expected.to include(org1plan1, org1plan2, org1plan3) }
+      it { is_expected.not_to include(org1plan4) }
+    end
+
+    context 'when user belongs to Org and a plan removed by creator, but cowner still active.' do
+      before do
+        Rails.configuration.x.plans.org_admins_read_all = true
+        coowner = create(:user, org: org1)
+        org1plan4.add_user!(coowner.id, :coowner)
+        owner_id = org1plan4.owner.id
+        org1plan4.roles.map { |r| r.update(active: false) if r.user_id == owner_id }
+      end
+
+      it { is_expected.to include(org1plan1, org1plan2, org1plan3) }
+      it { is_expected.not_to include(org1plan4) }
+    end
+
+    context 'when user belongs to Org, plan user with role :administrator, but plan creator from a different Org' do
+      before do
+        Rails.configuration.x.plans.org_admins_read_all = true
+        # Creator belongs to different org
+        @plan = create(:plan, :creator, :organisationally_visible, org: create(:org))
+        @plan.add_user!(create(:user, org: org1).id, :administrator)
       end
 
       it {
-        is_expected.to include(plan)
+        is_expected.to include(org1plan1, org1plan2, org1plan3, org1plan4, @plan)
       }
     end
 
     context 'user belongs to Org and plan user with role :editor, but not :creator and :admin' do
       before do
-        plan.add_user!(user.id, :editor)
+        Rails.configuration.x.plans.org_admins_read_all = true
+        # Creator and admin belongs to different orgs
+        @plan = create(:plan, :creator, :organisationally_visible, org: create(:org))
+        @plan.add_user!(create(:org).id, :administrator)
+        # Editor belongs to org1
+        @plan.add_user!(create(:user, org: org1).id, :editor)
       end
 
-      it { is_expected.to include(plan) }
+      it { is_expected.not_to include(@plan) }
     end
 
     context 'user belongs to Org and plan user with role :commenter, but not :creator and :admin' do
       before do
-        plan.add_user!(user.id, :commenter)
+        Rails.configuration.x.plans.org_admins_read_all = true
+        # Creator and admin belongs to different orgs
+        @plan = create(:plan, :creator, :organisationally_visible, org: create(:org))
+        @plan.add_user!(create(:org).id, :administrator)
+        # Commenter belongs to org1
+        @plan.add_user!(create(:user, org: org1).id, :commentor)
       end
 
-      it { is_expected.to include(plan) }
+      it { is_expected.not_to include(@plan) }
     end
 
     context 'user belongs to Org and plan user with role :reviewer, but not :creator and :admin' do
       before do
-        plan.add_user!(user.id, :reviewer)
+        Rails.configuration.x.plans.org_admins_read_all = true
+        # Creator and admin belongs to different orgs
+        @plan = create(:plan, :creator, :organisationally_visible, org: create(:org))
+        @plan.add_user!(create(:org).id, :administrator)
+        # Reviewer belongs to org1
+        @plan.add_user!(create(:user, org: org1).id, :reviewer)
       end
 
-      it { is_expected.to include(plan) }
+      it { is_expected.not_to include(@plan) }
     end
 
     context 'read_all is false, visibility private and user org_admin' do
       before do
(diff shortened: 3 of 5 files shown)


app/controllers/paginable/plans_controller.rb
```diff
@@ -46,7 +46,7 @@ def org_admin
       @super_admin = current_user.can_super_admin?
       @clicked_through = params[:click_through].present?
       plans = @super_admin ? Plan.all : current_user.org.org_admin_plans
-      plans = plans.joins(:template, roles: [user: :org]).where(Role.creator_condition)
+      plans = plans.joins(:template, roles: [user: :org])
```
aaronskiba, line 50: Given these changes, it makes sense to remove the redundant `.where(Role.creator_condition)` when `@super_admin` evaluates to false.
But what about when `@super_admin` evaluates to true? The previous code applied `.where(Role.creator_condition)` for super admins, but now it does not.

aaronskiba:
`Note: if an owner or co-owner de-activates a plan the Role of the user is set to active: false, so any plan with creator with role (having active: false) would mean the Plan is removed.`

I'm not 100% sure of that statement. Here is the `deactivate!` method in the Plan model.

```ruby
app/models/plan.rb
  # Deactivates the plan (sets all roles to inactive and visibility to :private)
  #
  # Returns Boolean
  def deactivate!
    # If no other :creator, :administrator or :editor is attached
    # to the plan, then also deactivate all other active roles
    # and set the plan's visibility to :private
    if authors.empty?
      roles.where(active: true).update_all(active: false)
      self.visibility = Plan.visibilities[:privately_visible]
      save!
    else
      false
    end
  end
```

Based on this code, it seems that rather than just the creator role, ALL roles associated with the plan are set to `active = false`. However, maybe this change is still needed?

Just to clarify, is this code change only menat to affect the CSV results?

app/controllers/paginable/plans_controller.rb
johnpinto1: I have mage the changes to ensure super admins plans has .where(Role.creator_condition). Always good to have code checked by a person not involved in change.
Selection_038

johnpinto1:
> `Note: if an owner or co-owner de-activates a plan the Role of the user is set to active: false, so any plan with creator with role (having active: false) would mean the Plan is removed.`
> 
> I'm not 100% sure of that statement. Here is the `deactivate!` method in the Plan model.
> 
> ```ruby
> app/models/plan.rb
>   # Deactivates the plan (sets all roles to inactive and visibility to :private)
>   #
>   # Returns Boolean
>   def deactivate!
>     # If no other :creator, :administrator or :editor is attached
>     # to the plan, then also deactivate all other active roles
>     # and set the plan's visibility to :private
>     if authors.empty?
>       roles.where(active: true).update_all(active: false)
>       self.visibility = Plan.visibilities[:privately_visible]
>       save!
>     else
>       false
>     end
>   end
> ```
> 
> Based on this code, it seems that rather than just the creator role, ALL roles associated with the plan are set to `active = false`. However, maybe this change is still needed?
> 
> Just to clarify, is this code change only menat to affect the CSV results?

I have removed the clearly false comment in my commit.

aaronskiba (commented):
NOTE: Updated this comment for accuracy.

```ruby
.where(roles: { active: true }).where(Role.creator_condition)
```
```ruby
irb(main):001> Role.creator_condition
=> "(\"roles\".\"access\" in (1,3,5,7,9,11,13,15,17,19,21,23,25,27,29,31))"
```

So it is verifying that there is at least one active creator on the plan. As an edge case, there do in fact appear to be plans with a completely absent creator role. I believe the following SQL query can be used to check your db:

```sql
SELECT p.*
FROM plans p
WHERE NOT EXISTS (
    SELECT 1
    FROM roles r
    WHERE r.plan_id = p.id
      AND r.access IN (1,3,5,7,9,11,13,15,17,19,21,23,25,27,29,31)
);
```
These plans mostly appear to be quite old in DMP Assistant's db. I can look further, but I assume these plans were created prior to the current bit field implementation in the Role model:
```ruby
  ##
  # Define Bit Field Values
  # Column access
  has_flags 1 => :creator,            # 1
            2 => :administrator,      # 2
            3 => :editor,             # 4
            4 => :commenter,          # 8
            5 => :reviewer,           # 16
            column: 'access',
            check_for_column: !Rails.env.test?
```

Are we sure we want to filter out plans with absent creator roles?

johnpinto1:
> Are we sure we want to filter out plans with absent creator roles?

This PR code change aligns the CSV for the Org admin with the current Org admin plans view. 

Any further changes to will need to thought through. Do we want to do this here or address this in a separate issue?

app/models/org.rb
```diff
@@ -293,11 +293,13 @@ def org_admin_plans
     if Rails.configuration.x.plans.org_admins_read_all
       Plan.includes(:template, :phases, :roles, :users).where(id: combined_plan_ids)
           .where(roles: { active: true })
+          .where(Role.creator_condition)
     else
       Plan.includes(:template, :phases, :roles, :users).where(id: combined_plan_ids)
           .where.not(visibility: Plan.visibilities[:privately_visible])
           .where.not(visibility: Plan.visibilities[:is_test])
           .where(roles: { active: true })
+          .where(Role.creator_condition)
```
aaronskiba: I wonder if a refactor like the following would be a nice improvement:
```ruby
def org_admin_plans
  scope = Plan.includes(:template, :phases, :roles, :users)
              .where(id: (native_plan_ids + affiliated_plan_ids).uniq)
              .where(roles: { active: true })
              .where(Role.creator_condition)

  unless Rails.configuration.x.plans.org_admins_read_all
    scope = scope.where.not(visibility: Plan.visibilities[:privately_visible])
                 .where.not(visibility: Plan.visibilities[:is_test])
  end

  scope
end
```

app/controllers/paginable/plans_controller.rb
```diff
@@ -45,8 +45,14 @@ def org_admin
       # check if current user if super_admin
       @super_admin = current_user.can_super_admin?
       @clicked_through = params[:click_through].present?
-      plans = @super_admin ? Plan.all : current_user.org.org_admin_plans
-      plans = plans.joins(:template, roles: [user: :org]).where(Role.creator_condition)
```
aaronskiba, line 49: How about the following for a refactor?
```ruby
plans = @super_admin ? Plan.where(Role.creator_condition) : current_user.org.org_admin_plans
plans = plans.joins(:template, roles: [user: :org])
```

spec/models/org_spec.rb
```diff
@@ -344,79 +344,124 @@
...
+      it { is_expected.not_to include(org1plan4) }
     end
 
-    context 'when user belongs to Org and plan user with role :administrator' do
+    context 'when user belongs to Org and a plan removed by creator, but cowner still active.' do
       before do
-        plan.add_user!(user.id, :administrator)
+        Rails.configuration.x.plans.org_admins_read_all = true
+        coowner = create(:user, org: org1)
+        org1plan4.add_user!(coowner.id, :coowner)
```
aaronskiba: You're trying to add a `:coowner` here, but it looks like a `:commenter` is being added instead:

```ruby
  # Creates a role for the specified user (will update the user's
  # existing role if it already exists)
  #
  # Expects a User.id and access_type from the following list:
  #  :creator, :administrator, :editor, :commenter
  #
  # Returns Boolean
  def add_user!(user_id, access_type = :commenter)
    user = User.where(id: user_id).first
    if user.present?
      role = Role.find_or_initialize_by(user_id: user_id, plan_id: id)

      # Access is cumulative, so set the appropriate flags
      # (e.g. an administrator can also edit and comment)
      case access_type
      when :creator
        role.creator = true
        role.administrator = true
        role.editor = true
      when :administrator
        role.administrator = true
        role.editor = true
      when :editor
        role.editor = true
      end
      role.commenter = true
      role.save
    else
      false
    end
  end
```

```
   386:         org1plan4.roles.map { |r| r.update(active: false) if r.user_id == owner_id }
   387:         byebug
=> 388:       end
   389: 
   390:       it { is_expected.to include(org1plan1, org1plan2, org1plan3) }
   391:       it { is_expected.not_to include(org1plan4) }
   392:     end
(byebug) roles = org1plan4.roles
#<ActiveRecord::Associations::CollectionProxy [#<Role id: 4, user_id: 4, plan_id: 7, created_at: "2025-12-03 17:37:57.852288000 +0000", updated_at: "2025-12-03 17:37:57.855932000 +0000", access: 15, active: false>, #<Role id: 6, user_id: 6, plan_id: 7, created_at: "2025-12-03 17:37:57.987364000 +0000", updated_at: "2025-12-03 17:37:57.987364000 +0000", access: 8, active: true>]>
(byebug) roles.last.creator?
false
(byebug) roles.last.administrator?
false
(byebug) roles.last.editor?
false
(byebug) roles.last.commenter?
true
(byebug) roles.last.reviewer?
false
(byebug) 
```

app/models/org.rb
johnpinto1: Thanks for refactor advice. Nice and neat. Will add commit with change.

spec/models/org_spec.rb
johnpinto1: My bad. Only allowed access_types ":creator, :administrator, :editor, :commenter". Will update code.

app/controllers/paginable/plans_controller.rb
johnpinto1: This code change in a new commit with another bit of re-factored code cited above.

spec/models/org_spec.rb
johnpinto1: Updated tests to only add access_type :administrator for a coowner test.

aaronskiba:
> > Are we sure we want to filter out plans with absent creator roles?
> 
> This PR code change aligns the CSV for the Org admin with the current Org admin plans view.
> 
> Any further changes to will need to thought through. Do we want to do this here or address this in a separate issue?

Sorry, I've updated that previous comment I made, as it was partly inaccurate. (However, there are in fact active plans with non-existent creator roles.)

I am thinking of an alternate way to align the CSV with the Org Admin plans. This PR currently does so by applying `.where(Role.creator_condition)` in `def org_admin_plans`. Is removing `.where(Role.creator_condition)` altogether another way to accomplish that? I wish I knew why `.where(Role.creator_condition)` was first introduced, because I'm not sure why active plans with inactive creators would not be returned by `org_admin_plans`.

aaronskiba:
> > > Are we sure we want to filter out plans with absent creator roles?
> > 
> > 
> > This PR code change aligns the CSV for the Org admin with the current Org admin plans view.
> > Any further changes to will need to thought through. Do we want to do this here or address this in a separate issue?
> 
> Sorry, I've updated that previous comment I made, as it was partly inaccurate. (However, there are in fact active plans with non-existent creator roles.)
> 
> I am thinking of an alternate way to align the CSV with the Org Admin plans. This PR currently does so by applying `.where(Role.creator_condition)` in `def org_admin_plans`. Is removing `.where(Role.creator_condition)` altogether another way to accomplish that? I wish I knew why `.where(Role.creator_condition)` was first introduced, because I'm not sure why active plans with inactive creators would not be returned by `org_admin_plans`.

I tried digging back into the `app/controllers/paginable/plans_controller.rb` history and found this: https://github.com/DMPRoadmap/roadmap/pull/2434
Sorry @martaribeiro, this is going way back, but do you have any thoughts or recollections on why `.where(Role.creator_condition)` may have been introduced here.

johnpinto1:
> > > Are we sure we want to filter out plans with absent creator roles?
> > 
> > 
> > This PR code change aligns the CSV for the Org admin with the current Org admin plans view.
> > Any further changes to will need to thought through. Do we want to do this here or address this in a separate issue?
> 
> Sorry, I've updated that previous comment I made, as it was partly inaccurate. (However, there are in fact active plans with non-existent creator roles.)
> 
> I am thinking of an alternate way to align the CSV with the Org Admin plans. This PR currently does so by applying `.where(Role.creator_condition)` in `def org_admin_plans`. Is removing `.where(Role.creator_condition)` altogether another way to accomplish that? I wish I knew why `.where(Role.creator_condition)` was first introduced, because I'm not sure why active plans with inactive creators would not be returned by `org_admin_plans`.

 > I tried digging back into the app/controllers/paginable/plans_controller.rb history and found this: https://github.com/DMPRoadmap/roadmap/pull/2434
Sorry @martaribeiro, this is going way back, but do you have any thoughts or recollections on why .where(Role.creator_condition) may have been introduced here.

Let us research the history of the code evolution to answer this intelligently. However, we in DMPOnline focused on aligning what the Org admins saw in the UI with the CSV download for this PR. I suspect we need a new issue for further changes.

aaronskiba:
> > > > Are we sure we want to filter out plans with absent creator roles?
> > > 
> > > 
> > > This PR code change aligns the CSV for the Org admin with the current Org admin plans view.
> > > Any further changes to will need to thought through. Do we want to do this here or address this in a separate issue?
> > 
> > 
> > Sorry, I've updated that previous comment I made, as it was partly inaccurate. (However, there are in fact active plans with non-existent creator roles.)
> > I am thinking of an alternate way to align the CSV with the Org Admin plans. This PR currently does so by applying `.where(Role.creator_condition)` in `def org_admin_plans`. Is removing `.where(Role.creator_condition)` altogether another way to accomplish that? I wish I knew why `.where(Role.creator_condition)` was first introduced, because I'm not sure why active plans with inactive creators would not be returned by `org_admin_plans`.
> 
> > I tried digging back into the app/controllers/paginable/plans_controller.rb history and found this: #2434
> > Sorry @martaribeiro, this is going way back, but do you have any thoughts or recollections on why .where(Role.creator_condition) may have been introduced here.
> 
> Let us research the history of the code evolution to answer this intelligently. However, we in DMPOnline focused on aligning what the Org admins saw in the UI with the CSV download for this PR. I suspect we need a new issue for further changes.

I guess the two will be aligned whether we add or remove `.where(Role.creator_condition)` from `org_admin_plans`. Maybe keeping it for now, as you are doing here, errs on the safer side.

Maybe an example scenario like the following would help: Suppose Marta is the org_admin, and John and I belong to her org. I create a plan, mark it organisationally visible, and add John as a co-owner. Marta can see this plan. Later I delete myself from the plan, leaving John as an active co-owner but with no remaining active “creator” role on the record. Should Marta still see it in the Org Admin plans page and CSV?

I don't know if this PR provides any further insights. It made a change that replaced `.where(Role.creator_condition)` with `.where(roles: { active: true })` in `scope :search` 
https://github.com/DMPRoadmap/roadmap/pull/3023

johnpinto1:
> > > > > Are we sure we want to filter out plans with absent creator roles?
> > > > 
> > > > 
> > > > This PR code change aligns the CSV for the Org admin with the current Org admin plans view.
> > > > Any further changes to will need to thought through. Do we want to do this here or address this in a separate issue?
> > > 
> > > 
> > > Sorry, I've updated that previous comment I made, as it was partly inaccurate. (However, there are in fact active plans with non-existent creator roles.)
> > > I am thinking of an alternate way to align the CSV with the Org Admin plans. This PR currently does so by applying `.where(Role.creator_condition)` in `def org_admin_plans`. Is removing `.where(Role.creator_condition)` altogether another way to accomplish that? I wish I knew why `.where(Role.creator_condition)` was first introduced, because I'm not sure why active plans with inactive creators would not be returned by `org_admin_plans`.
> > 
> > 
> > > I tried digging back into the app/controllers/paginable/plans_controller.rb history and found this: #2434
> > > Sorry @martaribeiro, this is going way back, but do you have any thoughts or recollections on why .where(Role.creator_condition) may have been introduced here.
> > 
> > 
> > Let us research the history of the code evolution to answer this intelligently. However, we in DMPOnline focused on aligning what the Org admins saw in the UI with the CSV download for this PR. I suspect we need a new issue for further changes.
> 
> I guess the two will be aligned whether we add or remove `.where(Role.creator_condition)` from `org_admin_plans`. Maybe keeping it for now, as you are doing here, errs on the safer side.
> 
> Maybe an example scenario like the following would help: Suppose Marta is the org_admin, and John and I belong to her org. I create a plan, mark it organisationally visible, and add John as a co-owner. Marta can see this plan. Later I delete myself from the plan, leaving John as an active co-owner but with no remaining active “creator” role on the record. Should Marta still see it in the Org Admin plans page and CSV?
> 
> I don't know if this PR provides any further insights. It made a change that replaced `.where(Role.creator_condition)` with `.where(roles: { active: true })` in `scope :search` #3023

This PR as I have said was to address the discrepancy between view and CSV because the CSV was using a different call. This just aligned the call. The users were finding it strange that the CSV had extra plans not visible in UI.

aaronskiba:
> This PR as I have said was to address the discrepancy between view and CSV because the CSV was using a different call. This just aligned the call. The users were finding it strange that the CSV had extra plans not visible in UI.

Sounds good. Sorry, yes, my review/questions might be a bit overkill. This only changes/reduces the CSV results to align with the org admin view.

aaronskiba (commented):
Glad you added some test coverage for `org_admin_plans`. I was going to add some comments for that as well as some refactoring recommendations in `spec/factories/plans.rb`.

I ended up creating a branch off of yours and tried to code things up instead. Hope that's okay and that I didn't go too far with my changes.

I made two commits on this branch: https://github.com/DMPRoadmap/roadmap/tree/aaron/issue/3561

Let me know what you think. If it works, maybe we could just cherry-pick those commits onto this PR's branch.

johnpinto1:
> Glad you added some test coverage for `org_admin_plans`. I was going to add some comments for that as well as some refactoring recommendations in `spec/factories/plans.rb`.
> 
> I ended up creating a branch off of yours and tried to code things up instead. Hope that's okay and that I didn't go too far with my changes.
> 
> I made two commits on this branch: https://github.com/DMPRoadmap/roadmap/tree/aaron/issue/3561
> 
> Let me know what you think. If it works, maybe we could just cherry-pick those commits onto this PR's branch.

That looks like nice concise re-factoring of my code. So merge.

johnpinto1:
> That looks like nice concise re-factoring of my code. So merge.

@aaronskiba Nice that is merged.

aaronskiba (commented):
This PR resolves the discrepancy between the org admin view and the outputted CSV, but it also applies the additional `.where(Role.creator_condition)` filtering to plans in the following places:

1. Plans returned via `GET api/v0/plans` (See `def index` in `app/controllers/api/v0/plans_controller.rb`)

2. Plans displayed in the "#{org.name} Plans" section of the `/plans` page (See `scope :organisationally_or_publicly_visible` in `app/models/plan.rb`)

Are we wanting to perform the extra filtering in these sections as well? If not, then rather solving the discrepancy by adding `.where(Role.creator_condition)` to `org_admin_plans`, we could instead remove `.where(Role.creator_condition)`  when querying plans for the CSV.

Issue #3555, opened 2025-08-22
Org admins can show/hide the ethical_issues field on plans


andreadavanzo:
We need to permit the org admins to show/hide the ethical_issues field on plans (Project details tab).

PR #3578, opened 2025-10-17
DRAFT PR: Added feature "Org admins can show/hide the ethical_issues field on p…

johnpinto1:
…lans" #3555 
Changes:

(1) Added a property Rails.configuration.x.madmp.enable_ethical_issues_per_template to configuration which works as follows: If Rails.configuration.x.madmp.enable_ethical_issues is true, this flag hide_ethical_issues_question_per_template comes into play. It will determine if the ethical_issues question is hidden/shown based on the template selected for the plan. If the template has the hide_ethical_issues attribute set to true, then the question will be hidden. If the template has the hide_ethical_issues is set to false, then the question will be shown.

(2) Added to Template model
    attribute :hide_ethical_issues_question, :boolean, default: false
with migration.

(3) Added a safe param :hide_ethical_issues_question to the org admin Template controller.

(4) Updated the app/views/org_admin/templates/_form.html.erb to display a checkbox for "Hide question about ethical issues". This checkbox will only appear if Rails.configuration.x.madmp.enable_ethical_issues and Rails.configuration.x.madmp.enable_ethical_issues_per_template are both set to true in config.

(5) Updated the conitional for display of the ethical issues question in the Project details page as follows:
<% if Rails.configuration.x.madmp.enable_ethical_issues &&
      (!Rails.configuration.x.madmp.enable_hide_ethical_issues_per_template  ||
       (Rails.configuration.x.madmp.enable_hide_ethical_issues_per_template  && plan.template.present? && plan.template.hide_ethical_issues_question.present? && !plan.template.hide_ethical_issues_question)) %>

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.1].define(version: 2025_01_15_102816) do
+ActiveRecord::Schema[7.1].define(version: 2025_10_16_134521) do
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
 
@@ -548,6 +548,7 @@ ActiveRecord::Schema[7.1].define(version: 2025_01_15_102816) do
     t.integer "family_id"
     t.boolean "archived"
     t.text "links"
+    t.boolean "hide_ethical_issues_question"
     t.index ["family_id", "version"], name: "index_templates_on_family_id_and_version", unique: true
     t.index ["family_id"], name: "index_templates_on_family_id"
     t.index ["org_id", "family_id"], name: "template_organisation_dmptemplate_index"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Changelog
 
+- Feature for Org admins can show/hide the ethical_issues field on plans [#3555](https://github.com/DMPRoadmap/roadmap/pull/3555) if config enabled.
+
 ## v5.0.2
 - Bump Ruby to v3.1.4 and use `.ruby-version` in CI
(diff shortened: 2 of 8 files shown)


db/schema.rb
```diff
@@ -10,10 +10,25 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.1].define(version: 2025_01_15_102816) do
+ActiveRecord::Schema[7.1].define(version: 2025_10_16_134521) do
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
 
+  create_table "admin_users", id: :serial, force: :cascade do |t|
```
momo3404: There seems to be a lot of unintended changes being made to the DB schema here. Could a different DB have been used when the migration was made?

johnpinto1: @momo3404 Thanks for spotting that. I will remove the schema.rb from commit.

johnpinto1: @momo3404 I have removed the schema.rb from commit as suggested.

johnpinto1: @momo3404 I think I need to load schema.rb again without the unnecessary change to fix test on Github.

johnpinto1: @momo3404 After a bit of a struggle I managed to create schema.rb with only the current migration change.
After restoring old db/schema.rb
git co db/schema.rb
rails db:schema:load
rails db:migrate VERSION=20251016134521

johnpinto1: Test failing for another reason now.

johnpinto1:
Withdrawing PR, as there was some misunderstanding on my part on the requirements that DCC implemented.

johnpinto1:
Thanks @aaronskiba and @momo3404  for comments and advice.

PR #3579, opened 2025-10-22
Issue #3555 - Added feature for Org admins can show/hide the

johnpinto1:
Ethical_Issues fragment in plans project details page.
Co-authored-by: don-stuckey <dstuckey@ed.ac.uk>

This feature was contributed to DMPonline by @don-stuckey.

**Changes:**
- These changes only kick in if enable_ethical_issues is enable in the application config.
- Added to Org model attribute in a migration add_ro_ethical_concerns  :boolean  default(true), not null
- Added a checkbox '_Add "Research outputs may have ethical concerns" field to "Project details" tab in plans_' to app/views/orgs/_profile_form.html.erb.
- Added code to the Plan Project Details view to show or hide the Ethical Issues fragment.
- Added :add_ro_ethical_concerns to the strong params org_params for the Org controller.

Here is the spec and tests compiled by @don-stuckey who contributed this feature to DMPonline:

**Specification**
For each organisation’s profile in DMPonline, the “Organisation Permission” section now includes the checkbox “Add "Research outputs may have ethical concerns" field to "Project details" tab in plans”, and the default value of this checkbox is true.

**Test 1: Ensuring that the option to enable / disable the "Research outputs may have ethical concerns" checkbox for an organisation is available for super-admins** 

Test steps: 

    tester signs in as a super-admin and opens their organisation, e.g. “Organisation 1” (Org1) in edit mode 

Expected results: 

    tester sees the option to enable / disable the "Research outputs may have ethical concerns" checkbox for the organisation 

**Test 2: Ensuring that the option to enable / disable the "Research outputs may have ethical concerns" checkbox for an organisation is available for org-admins for their organisation** 

Test steps: 

    tester signs in as a org-admin and opens their organisation, e.g. “Organisation 1” (Org1) in edit mode 

Expected results: 

    tester sees the option to enable / disable the "Research outputs may have ethical concerns" checkbox for their organisation 

**Test 3: Disabling "Research outputs may have ethical concerns" checkbox for an organisation** 

Test steps: 

    tester signs in as a org-admin and opens their organisation, e.g. “Organisation 1” (Org1) in edit mode 

    tester de-selects the checkbox “Organisation Permissions > “Add "Research outputs may have ethical concerns" field to "Project details" tab in plans” and selects “Save” 

    tester signs in as a standard user associated with Org1 

    tester creates a plan and saves a plan 

Expected results: 

    tester does not see the "Research outputs may have ethical concerns" field in the “Project details” page on creating a plan or on editing that plan they just created 

**Test 4: Ensuring that existing plans that have the "Research outputs may have ethical concerns" checkbox checked still have that field as visible if the organisation disables the "Research outputs may have ethical concerns" checkbox** 

Test steps: 

    tester signs in as a org-admin and opens an organisation, e.g. “Organisation 1” (Org1) in edit mode, and ensures that  

    the “Organisation Permissions > “Add "Research outputs may have ethical concerns" field to "Project details" tab in plans” field is checked 

    tester signs in as a standard user associated with Org1 

    tester creates a plan, checks the  "Research outputs may have ethical concerns" field, and saves the plan 

    tester signs in as a org-admin and opens up Org1 in edit mode 

    tester then deselects the “Organisation Permissions > “Add "Research outputs may have ethical concerns" field to "Project details" tab in plans” checkbox and saves the change 

    tester signs in as a standard user  

Expected results: 

    tester should still see the "Research outputs may have ethical concerns" checkbox in the plan they created even though that field has been disabled for all new (not existing) plans going forward 

**Test 5: Ensuring that existing plans that have the do not have the "Research outputs may have ethical concerns" checkbox checked subsequently do not show that field when the organisation disables the "Research outputs may have ethical concerns" checkbox** 

Test steps: 

    tester signs in as a org-admin and opens an organisation, e.g. “Organisation 1” (Org1) in edit mode, and ensures that  

    the “Organisation Permissions > “Add "Research outputs may have ethical concerns" field to "Project details" tab in plans” checkbox is checked 

    tester signs in as a standard user associated with Org1 

    tester creates a plan, and does not check the "Research outputs may have ethical concerns" checkbox, and saves the plan 

    tester signs in as a org-admin and opens up Org1 in edit mode 

    tester then de-selects the “Organisation Permissions > “Add "Research outputs may have ethical concerns" field to "Project details" tab in plans” checkbox and saves the change 

    tester signs in as a standard user and opens the plan they just created 

Expected results: 

    tester should not see the "Research outputs may have ethical concerns" checkbox in the plan they created 

Selection_139

Selection_140

Selection_141

### DIFF
--- a/app/views/orgs/_profile_form.html.erb
+++ b/app/views/orgs/_profile_form.html.erb
@@ -180,7 +180,37 @@
       </div>
     <% end %>
   </div>
+  <hr>
+  <%# Organisation Permissions block where super admin can add:
+    "Project details" tab in a plan
+  %>
 
+  <% org_perms_header_block = capture do %>
+    <h3>
+      <%= _('Organisation Permissions') %>
+    </h3>
+  <% end %>
+
+  <% add_ro_ethical_concerns_block = capture do %>
+    <div class="checkbox">
+      <%= f.label :add_ro_ethical_concerns do %>
+        <%= f.check_box :add_ro_ethical_concerns, { class: 'org_types', value: org.add_ro_ethical_concerns?  }, "true", "false" %>
+        <%= _('Add "Research outputs may have ethical concerns" field to "Project details" tab in plans') %>
+      <% end %>
+    </div>
+  <% end %>
+
+  <div class="row">
+    <% if current_user.can_super_admin? %>
+      <fieldset class="col-sm-8">
+        <%= org_perms_header_block %>
+        <%= add_ro_ethical_concerns_block %>
+      </fieldset>
+    <% elsif current_user.can_org_admin? %>
+      <%= org_perms_header_block %>
+      <%= add_ro_ethical_concerns_block %>
+    <% end %>
+  </div>
   <hr>
 
   <% if !org.new_record? %>
--- a/app/views/orgs/_profile_form.html.erb
+++ b/app/views/orgs/_profile_form.html.erb
@@ -200,18 +200,15 @@
     </div>
   <% end %>
 
-  <div class="row">
-    <% if current_user.can_super_admin? %>
+  <% if current_user.can_org_admin? %>
+     <div class="row">
       <fieldset class="col-sm-8">
         <%= org_perms_header_block %>
         <%= add_ro_ethical_concerns_block %>
       </fieldset>
-    <% elsif current_user.can_org_admin? %>
-      <%= org_perms_header_block %>
-      <%= add_ro_ethical_concerns_block %>
-    <% end %>
-  </div>
-  <hr>
+    </div>
+    <hr>
+  <% end %>
 
   <% if !org.new_record? %>
     <%= render partial: "orgs/external_identifiers",
--- a/app/views/plans/_project_details.html.erb
+++ b/app/views/plans/_project_details.html.erb
@@ -102,7 +102,7 @@ ethics_report_tooltip = _("Link to a protocol from a meeting with an ethics comm
   </div>
 <% end %>
 
-<% if Rails.configuration.x.madmp.enable_ethical_issues %>
+<% ethical_issues_block = capture do %>
   <conditional>
     <div class="form-control mb-3">
       <div class="col-lg-8">
@@ -141,6 +141,16 @@ ethics_report_tooltip = _("Link to a protocol from a meeting with an ethics comm
       </div>
     </div>
   </conditional>
+<% end %>
+
+<% if Rails.configuration.x.madmp.enable_ethical_issues %>
+    <% if plan.ethical_issues == true %>
+      <%= ethical_issues_block %>
+    <% else %>
+      <% if current_user.org.add_ro_ethical_concerns? %>
+        <%= ethical_issues_block %>
+      <% end %>
+    <% end %>
 <% end %>
 
 <conditional>
--- a/db/migrate/20251027100520_add_add_ro_ethical_concerns_to_orgs.rb
+++ b/db/migrate/20251027100520_add_add_ro_ethical_concerns_to_orgs.rb
@@ -1,4 +1,4 @@
-class AddAddRoEthicalConcernsToOrgs < ActiveRecord::Migration[6.1]
+class AddAddRoEthicalConcernsToOrgs < ActiveRecord::Migration[7.1]
   def change
     add_column :orgs, :add_ro_ethical_concerns, :boolean, default: true, null: false
     Org.update_all(add_ro_ethical_concerns: true)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Changelog
 
+- Feature for Org admins can show/hide the ethical_issues field on plans [#3555](https://github.com/DMPRoadmap/roadmap/pull/3555).
+
 ## v5.0.2
 - Bump Ruby to v3.1.4 and use `.ruby-version` in CI
   - [#3566](https://github.com/DMPRoadmap/roadmap/pull/3566)
--- a/app/controllers/orgs_controller.rb
+++ b/app/controllers/orgs_controller.rb
@@ -222,6 +222,7 @@ class OrgsController < ApplicationController
           .permit(:name, :abbreviation, :logo, :contact_email, :contact_name,
                   :remove_logo, :managed, :feedback_enabled, :org_links,
                   :funder, :institution, :organisation,
+                  :add_ro_ethical_concerns,
                   :feedback_msg, :org_id, :org_name, :org_crosswalk,
                   :helpdesk_email,
                   identifiers_attributes: %i[identifier_scheme_id value],
--- a/app/models/org.rb
+++ b/app/models/org.rb
@@ -25,6 +25,7 @@
 #  region_id              :integer
 #  managed                :boolean          default(false), not null
 #  helpdesk_email         :string
+#  add_ro_ethical_concerns  :boolean        default(true), not null
 #
 # Foreign Keys
 #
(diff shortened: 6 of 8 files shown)


gjacob24:
Looks good @johnpinto1. I used the same test cases that we used in DMPonline to test this branch.

gjacob24 (approved):

db/migrate/20251027100520_add_add_ro_ethical_concerns_to_orgs.rb
aaronskiba, line 1: It may be best to recreate this migration file (i.e. `rails generate migration AddAddRoEthicalConcernsToOrgs`) so that it uses the version of Rails that we currently on (i.e. 7.1 rather than 6.1).

johnpinto1: The migration has been re-created for Rails version 7.1 as suggested.

app/views/plans/_project_details.html.erb
```diff
@@ -143,6 +143,16 @@ ethics_report_tooltip = _("Link to a protocol from a meeting with an ethics comm
   </conditional>
 <% end %>
 
+<% if Rails.configuration.x.madmp.enable_ethical_issues %>
```
aaronskiba, line 146: What do you think of the following refactor?

```ruby
<% if Rails.configuration.x.madmp.enable_ethical_issues && (plan.ethical_issues? || current_user.org.add_ro_ethical_concerns?) %>
  <%= ethical_issues_block %>
<% end %>
```

app/views/orgs/_profile_form.html.erb
```diff
@@ -180,7 +180,37 @@
...
+    <h3>
+      <%= _('Organisation Permissions') %>
+    </h3>
+  <% end %>
+
+  <% add_ro_ethical_concerns_block = capture do %>
+    <div class="checkbox">
+      <%= f.label :add_ro_ethical_concerns do %>
+        <%= f.check_box :add_ro_ethical_concerns, { class: 'org_types', value: org.add_ro_ethical_concerns?  }, "true", "false" %>
+        <%= _('Add "Research outputs may have ethical concerns" field to "Project details" tab in plans') %>
```
aaronskiba, line 198: I wonder if these labels are a bit too literal and developer-y. What do we think of something more along the following lines?:
h3 title: "Organisation Customisations"
Checkbox caption: "Allow users to describe potential ethical issues in plans"

app/views/orgs/_profile_form.html.erb
```diff
@@ -180,7 +180,37 @@
...
+  <div class="row">
+    <% if current_user.can_super_admin? %>
+      <fieldset class="col-sm-8">
+        <%= org_perms_header_block %>
+        <%= add_ro_ethical_concerns_block %>
+      </fieldset>
+    <% elsif current_user.can_org_admin? %>
+      <%= org_perms_header_block %>
+      <%= add_ro_ethical_concerns_block %>
+    <% end %>
```
aaronskiba: Is there a reason why one uses `<fieldset>` tags but the other doesn't? If not, could we just refactor to the following?

```ruby
    <% if current_user.can_org_admin? %>
      <fieldset class="col-sm-8">
        <%= org_perms_header_block %>
        <%= add_ro_ethical_concerns_block %>
      </fieldset>
    <% end %>
```

NOTE: We could omit `current_user.can_super_admin?` in the following case because `can_org_admin?` is implemented as follows:

```ruby
  def can_org_admin?
    return true if can_super_admin?

    # Automatically false if the user has no Org or the Org is not managed
    return false unless org.present? && org.managed?

    can_grant_permissions? || can_modify_guidance? ||
      can_modify_templates? || can_modify_org_details? ||
      can_review_plans?
  end
  # rubo
```

aaronskiba (commented):
I'm considering the research outputs ethical concerns handling we already have prior to this change. We have `Rails.configuration.x.madmp.enable_ethical_issues` to enable/disable the option globally. Users also have the "Research outputs may have ethical concerns" checkbox to enable/disable it on a per plan level. I wonder how necessary it is to also have the same enable/disable option at the org level.

There is also the issue of users from different orgs collaborating on the same plan. `current_user.org.add_ro_ethical_concerns?` could evaluate differently for different users working on the same plan. If we go with this approach, would we also need to address that scenario?

Also, I'm not sure if DMP Assistant would even want org admins to have the ability to customise the visibility of this field. Hopefully, this is a conversation we can have at the next all-team monthly meeting.

johnpinto1:
> I'm considering the research outputs ethical concerns handling we already have prior to this change. We have `Rails.configuration.x.madmp.enable_ethical_issues` to enable/disable the option globally. Users also have the "Research outputs may have ethical concerns" checkbox to enable/disable it on a per plan level. I wonder how necessary it is to also have the same enable/disable option at the org level.
> 
> There is also the issue of users from different orgs collaborating on the same plan. `current_user.org.add_ro_ethical_concerns?` could evaluate differently for different users working on the same plan. If we go with this approach, would we also need to address that scenario?
> 
> Also, I'm not sure if DMP Assistant would even want org admins to have the ability to customise the visibility of this field. Hopefully, this is a conversation we can have at the next all-team monthly meeting.

I think the way forward with this is you don't make this option available in the Org Admin view. In a DSpace based  project I work we customise the base code to our requirements. Otherwise, we won't be able add features because it is unlikely that some features will be universally required. Many European and British users want this feature as they feel having the Ethical Questions in a DMP confuses researchers.

app/views/orgs/_profile_form.html.erb
```diff
@@ -180,8 +180,35 @@
...
+  <% add_ro_ethical_concerns_block = capture do %>
+    <div class="checkbox">
+      <%= f.label :add_ro_ethical_concerns do %>
+        <%= f.check_box :add_ro_ethical_concerns, { class: 'org_types', value: org.add_ro_ethical_concerns?  }, "true", "false" %>
+        <%= _('Add "Research outputs may have ethical concerns" field to "Project details" tab in plans') %>
+      <% end %>
+    </div>
+  <% end %>
+
+  <% if current_user.can_org_admin? %>
```
aaronskiba, line 203: I had commented on a refactor before, but now I'm wondering if even the `<% if current_user.can_org_admin? %>` is redundant -- aren't `org_admin` privileges required to access this page in the first place?

aaronskiba: Maybe it should be changed to `<% if Rails.configuration.x.madmp.enable_ethical_issues %>`? That way the option is hidden if the global config value is set to false.

db/migrate/20251027100520_add_add_ro_ethical_concerns_to_orgs.rb
```diff
@@ -0,0 +1,6 @@
+class AddAddRoEthicalConcernsToOrgs < ActiveRecord::Migration[7.1]
+  def change
+    add_column :orgs, :add_ro_ethical_concerns, :boolean, default: true, null: false
+    Org.update_all(add_ro_ethical_concerns: true)
```
aaronskiba, line 4: I wonder if `Org.update_all(add_ro_ethical_concerns: Rails.configuration.x.madmp.enable_ethical_issues)` would be even better.

Also, I am wondering, do we want org admins to be able to set the visibilty of the "Research outputs may have ethical concerns" field regardless of whether `Rails.configuration.x.madmp.enable_ethical_issues` is set to true or false?

johnpinto1: Worth discussing further at meeting.

Issue #3534, opened 2025-06-30
Redesign plan-creation page to enforce template access rules and simplify UI


don-stuckey:
## Problem
In the current plan-creation page, it is possible for a user to use an organisation's template even though the user is not a verified member of that organisation.

For example, suppose a user goes to the plan-creation page and then selects an organisation, say organisation X, that they are not a verified member of in the "Select the primary research organisation" field. Then, if the user checks the "No funder associated with this plan or my funder is not listed" box, the user can choose from one of the selected organisation's templates in the now-visible "Which DMP template would you like to use?" field. However, only verified members of an organisation should be able to use that organisation's templates.

## Solution
To solve this problem, we will change the plan-creation page. The plan-creation page will consist of a grouped dropdown menu, and the following wireframes show the dropdown collapsed (on the left) and expanded (on the right):

<img width="1075" alt="Image" src="https://github.com/user-attachments/assets/33aaa41b-15ce-49cc-94a7-eb91fb93e585" />

In the expanded dropdown, there are three groups, with the first group being "<institution_name> templates". Importantly, it is **only** the user's organisation's templates that are displayed in this group (and the user is a verified member of that organisation). The second group in the dropdown contains the global templates (new feature). And the third group contains the funder templates, which are considered public and so can be used by any user.

In addition, the new plan-creation page will benefit users by having a smaller and simplified field set of three fields (whereas the current create-a-new-plan page has effectively seven fields).

don-stuckey:
We have done a first iteration of the new create-plan page in issue branch `issues/3534`, and these screenshots show the iteration:

<img width="770" height="384" alt="Image" src="https://github.com/user-attachments/assets/7aefe0c8-02b4-4fd6-ab9c-d2d7fc6151d7" />

<img width="770" height="385" alt="Image" src="https://github.com/user-attachments/assets/a6279542-9204-40b4-9045-f675835727e2" />

PR #3580, opened 2025-10-27
Redesigned plan-creation page to enforce template access rules and simplify UI

johnpinto1:
Redesigned plan-creation page to enforce template access rules and simplify UI [#3534](https://github.com/DMPRoadmap/roadmap/issues/3534)
Co-authored-by: don-stuckey dstuckey@ed.ac.uk

This feature was contributed to DMPonline by @don-stuckey.

Changes:
- Complete re-write of 'Create a new plan' view _app/views/plans/new.html.erb_.
- The Plans controller app/controllers/plans_controller.rb has be greatly simplified to get three types og templates for selection in view: The global templates, the user's org templates and the funder templates available.
- The seeds.rb file has been updated to include templates for the three groups.
- The tests has been updated.

New Create a new plan page
Selection_142

Screenshot when running RSpec test in _spec/features/plans_spec.rb_
screenshot1

### DIFF
--- a/app/controllers/plans_controller.rb
+++ b/app/controllers/plans_controller.rb
@@ -29,62 +29,15 @@ class PlansController < ApplicationController
   # rubocop:enable Metrics/AbcSize
 
   # GET /plans/new
-  # rubocop:disable Metrics/AbcSize, Metrics/MethodLength, Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
   def new
     @plan = Plan.new
     authorize @plan
-    @plan.org_id = current_user.org&.id
-    # looks into the list of families of templates
-    customizations = Template.latest_customized_version_per_org(@plan.org_id)
-    customization_ids = customizations.select(&:published?).collect(&:customization_of)
+    org_id = current_user.org&.id
 
-    # get templates of user's own org
-    user_org_own_templates = Template.organisationally_visible
-                                     .where(org_id: @plan.org_id, customization_of: nil)
-                                     .published
-                                     .uniq.sort_by(&:title)
-
-    # get templates of user's customised org
-    user_org_custom_templates = Template.latest_customized_version_per_org(@plan.org_id)
-                                        .published
-                                        .uniq.sort_by(&:title)
-
-    # get funder templates no customised templates
-    funder_non_customised_templates = Template.published
-                                              .joins(:org)
-                                              .where(orgs: { org_type: Org.org_type_values_for(:funder) })
-                                              # The next line removes templates that belong to a family that
-                                              # has customised templates
-                                              .where.not(family_id: customization_ids)
-                                              .uniq.sort_by(&:title)
-
-    # get global templates
-    global_templates = Template.published
-                               .where(is_default: true)
-                               # The next line removes templates that belong to a family that
-                               # has customised templates
-                               .where.not(family_id: customization_ids)
-                               .uniq.sort_by(&:title)
-
-    # create templates-grouped hash
-    @templates_grouped = {
-      _("Your Organisation's Templates:") => user_org_own_templates.map do |t|
-        [t.title, t.id]
-      end,
-      _("Your Organisation's Customised Templates:") => user_org_custom_templates.map do |t|
-        [t.title, t.id]
-      end,
-      _('Global Templates:') => global_templates.map do |t|
-        [t.title, t.id]
-      end,
-      _('Funder Templates:') => funder_non_customised_templates.map do |t|
-        [t.title, t.id]
-      end
-    }.reject { |_, val| val.empty? }
-
+    # Get templates grouped hash
+    @templates_grouped = templates_available_to_org_user(org_id)
     respond_to :html
   end
-  # rubocop:enable Metrics/AbcSize, Metrics/MethodLength, Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
 
   # POST /plans
   # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
@@ -102,6 +55,15 @@ class PlansController < ApplicationController
         format.html { redirect_to new_plan_path }
       end
     else
+      template_id = plan_params[:template_id].to_i
+      unless validate_template_available_to_org_user?(template_id, current_user.org_id)
+        respond_to do |format|
+          flash[:alert] = _('The selected template is not available to your organisation.')
+          format.html { redirect_to new_plan_path }
+        end
+        return
+      end
+
       @plan.visibility = if plan_params['visibility'].blank?
                            Rails.configuration.x.plans.default_visibility
                          else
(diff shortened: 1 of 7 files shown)


app/controllers/plans_controller.rb
```diff
@@ -29,31 +29,43 @@ def index
...
+
+    # get global templates
+    global_templates = Template.published
+                               .where(is_default: true)
+                               .distinct
+
+    # get templates of user's org
+    user_org_templates = Template.published
+                                 .where(org: current_user.org)
+                                 .distinct
```
aaronskiba: I'm comparing this simplified logic with what we were previously using (https://github.com/DMPRoadmap/roadmap/blob/main/app/controllers/template_options_controller.rb).

One thing I am concerned about with this new logic is the same template being listed in two, or all three sections. For example, what if `Org.funder` has a default template (which is the case with DMP Assistant). Then that same template is listed in both funder_templates and global_templates. And if `user.org` == `Org.funder` (again the case with DMP Assistant), then `funder_templates` and `user_org_templates` will have identical lists.

`TemplateOptionsController` also performed special handling for template customisations. Specifically, if an org had a customisation of the most recent version of a funder_template, then only that customisation would appear as an template option. With this logic, it looks like the most recent customisation will appear in `user_org_templates`, but the non-customised version of it will also be listed as an option in `funder_templates`.

app/views/plans/new.html.erb
```diff
@@ -1,125 +1,90 @@
...
+    </div>
+  </div>
+  </fieldset>
+      <fieldset class="form-group">
+        <legend>
+          <%= _('Select a DMP template') %>
+        </legend>
+  <div class="row form-row">
+    <div class="col-12">
+        <% @templates_grouped.each do |group_label, group_templates| %>
```
aaronskiba, line 51: I like the label grouping. Maybe we should also be ordering the templates by title?

momo3404:
I just wanted to point out the fun coincidence that I'm also working on a similar redesign for the templates dropdown in DMP Assistant. This is [the PR](https://github.com/portagenetwork/roadmap/pull/1215) if you're curious, and here's what it looks like right now.

[Screencast from 2025-10-28 11-16-28.webm](https://github.com/user-attachments/assets/ada063f6-29f3-47ef-924c-b836b7189eee)

For more details about our approach, we initially edited `template_options_controller.rb` to organize the templates appropriately in [this PR](https://github.com/portagenetwork/roadmap/pull/1199). Then, in the PR I tagged above, we're editing the styling of the dropdown (with the new order) to include headers and a "Show more templates" button due to how many funder templates we have publicly available. We're only editing code in  `template_options_controller.rb` and `app/javascript/src/plans/new.js`, so it is a bit of a simpler implementation.

If you want, I'm interested on collaborating on this if you want to match our approach.

app/controllers/plans_controller.rb
aaronskiba: If we are no longer fetching all of the possible org options, maybe  `PlansController#new` should use the existing logic in `TemplateOptionsController`?

johnpinto1:
@aaronskiba @momo3404 Just saw your 2 comments. Our version is live and working as per our requirements and @don-stuckey did a lot of work on this. Maybe we can work out a way that we can chose which feature to use in config? Worth discussing and finding a solution we can go forward with for other features.

aaronskiba:
> @aaronskiba @momo3404 Just saw your 2 comments. Our version is live and working as per our requirements and @don-stuckey did a lot of work on this. Maybe we can work out a way that we can chose which feature to use in config? Worth discussing and finding a solution we can go forward with for other features.

Sounds good. This PR enforces the org selection to user.org, which is something we'll need at some point (maybe we should add a PR for the feature flags in the near future too?). @momo3404's PR maintains the current template options logic via `template_options_controller.rb`, which I think this PR would benefit from.

app/controllers/plans_controller.rb
johnpinto1: @aaronskiba Given that you are implementing this differently, maybe we keep this change in one place that only affects us. What do you think. We can discuss at Developers' meeting on Monday. I am off tomorrow.

johnpinto1:
> > @aaronskiba @momo3404 Just saw your 2 comments. Our version is live and working as per our requirements and @don-stuckey did a lot of work on this. Maybe we can work out a way that we can chose which feature to use in config? Worth discussing and finding a solution we can go forward with for other features.
> 
> Sounds good. This PR enforces the org selection to user.org, which is something we'll need at some point (maybe we should add a PR for the feature flags in the near future too?). @momo3404's PR maintains the current template options logic via `template_options_controller.rb`, which I think this PR would benefit from.

I will fix this on Monday or after meeting.

app/controllers/plans_controller.rb
aaronskiba: > I'm comparing this simplified logic with what we were previously using (https://github.com/DMPRoadmap/roadmap/blob/main/app/controllers/template_options_controller.rb).
> 
> One thing I am concerned about with this new logic is the same template being listed in two, or all three sections. For example, what if `Org.funder` has a default template (which is the case with DMP Assistant). Then that same template is listed in both funder_templates and global_templates. And if `user.org` == `Org.funder` (again the case with DMP Assistant), then `funder_templates` and `user_org_templates` will have identical lists.
> 
> `TemplateOptionsController` also performed special handling for template customisations. Specifically, if an org had a customisation of the most recent version of a funder_template, then only that customisation would appear as an template option. With this logic, it looks like the most recent customisation will appear in `user_org_templates`, but the non-customised version of it will also be listed as an option in `funder_templates`.

During the Monday meeting, I was informed that your team had recently implemented a fix for the duplicate templates I mentioned, and that it could be added to this PR.

johnpinto1:
@aaronskiba @momo3404 Updated the code with @martaribeiro and @gjacob24 fixes mentioned in last roadmap meeting. 
BTW I will be missing the next Roadmap meeting as we are on strike on Nov 17-19.

app/controllers/plans_controller.rb
```diff
@@ -29,31 +29,62 @@ def index
...
+                                              .uniq.sort_by(&:title)
+
+    # get global templates
+    global_templates = Template.published
+                               .where(is_default: true)
+                               # The next line removes templates that belong to a family that
+                               # has customised templates
+                               .where.not(family_id: customization_ids)
+                               .uniq.sort_by(&:title)
+
```
aaronskiba: The newly added commits appear to address some of my earlier concerns regarding duplicate templates, so thank you for that.

That said, I still notice that the `upgrade_customization?` handling from `TemplateOptionsController` isn’t being used here, and it’s unclear what other behavioural differences now exist compared to that controller.

I am wondering if, instead of a complete rework of the template selection logic, we could instead pass `user.org` to `TemplateOptionsController`. And if there is in fact any required behavioural changes (e.g. always return default/global templates, grouping the templates into the three categories, etc.) then the code could be changed there. Re-using the existing logic could also help mitigate any further regressions or follow-up fixes, and keep template-selection behaviour consistent.

johnpinto1:
@aaronskiba I think your suggestion on chat for server side validation is a good idea. Will sort that soon.

app/controllers/plans_controller.rb
johnpinto1: Will look at this.

spec/features/plans_spec.rb
aaronskiba, line 1: This comment isn't directly related to this PR, but I am surprised that the codebase doesn't include `spec/controllers/plans_controller_spec.rb`. The `PlansController` is so central to the app, that not having that Spec file makes me worry that we are missing some key code coverage. I will add a GitHub issue for this.

johnpinto1: Good point. Will address this and validation issue.

aaronskiba: I created a related issue [here](https://github.com/DMPRoadmap/roadmap/issues/3583). I'd be happy to get started on this one.

johnpinto1: I am current writing tests. Not got it right yet. I will push what I have got once I get it working correctly.

johnpinto1: Just pushed code for an RSpec test of methods introduced. Review and let me know if you think more tests are needed.

aaronskiba: Cool. I like that you added the validation and accompanying tests.

aaronskiba (commented):
There’s substantial work here. Regarding the update/removal of the org-selection logic in the “create plan” form, I recognize the need for these changes in DMP Online. This creates a divergence between DMP Online and DMP Assistant that we’ll need to accommodate.

My main hesitation is around the template-selection changes. The new implementation replaces the existing TemplateOptionsController, which introduces a larger split in how the two applications handle this area. Since the underlying template-selection requirements still seem quite similar, keeping the shared logic aligned would be preferable. I’m concerned that merging this approach could make that more difficult going forward.

aaronskiba:
> There’s substantial work here. Regarding the update/removal of the org-selection logic in the “create plan” form, I recognize the need for these changes in DMP Online. This creates a divergence between DMP Online and DMP Assistant that we’ll need to accommodate.
> 
> My main hesitation is around the template-selection changes. The new implementation replaces the existing TemplateOptionsController, which introduces a larger split in how the two applications handle this area. Since the underlying template-selection requirements still seem quite similar, keeping the shared logic aligned would be preferable. I’m concerned that merging this approach could make that more difficult going forward.

If you'd like, I could point a new PR at this one that attempts to refactor these changes so that the `TemplateOptionsController` logic is still utilised.

johnpinto1:
> There’s substantial work here. Regarding the update/removal of the org-selection logic in the “create plan” form, I recognize the need for these changes in DMP Online. This creates a divergence between DMP Online and DMP Assistant that we’ll need to accommodate.
> 
> My main hesitation is around the template-selection changes. The new implementation replaces the existing TemplateOptionsController, which introduces a larger split in how the two applications handle this area. Since the underlying template-selection requirements still seem quite similar, keeping the shared logic aligned would be preferable. I’m concerned that merging this approach could make that more difficult going forward.
>
> if you'd like, I could point a new PR at this one that attempts to refactor these changes so that the TemplateOptionsController logic is still utilised.

Agree, I think we should resolve this problem as you suggest with a new PR.

PR #3590, opened 2025-12-19
[DRAFT] Add question and answer flag to V2 API

momo3404:
Changes proposed in this PR:
- Add the `complete` flag to the V2 API to enable accessing all questions and answers belonging to a plan in the API response.
- Per the V2 API Documentation: "The idea is to have a flag which defaults to false and thereby is compliant with the RDA metadata standard, but can be passed in with a value of true to include the questions and answers of the plan."
- The flag is optional and works with the `api/v2/plans` and `api/v2/plans/[plan_id]` endpoints. For example: `api/v2/plans/12345?complete=true`.
- The schema the flag returns is as follows:
- ```
        "complete_plan": [
            {
              "title": "Question 1",
              "question": "<p>What considerations will you take into account with respect to ethical, legal, or commercial issues?</p>\r\n<p><br>Describe any applicable ethical, legal, or commercial considerations related to your project and data. This includes research involving Indigenous communities and knowledges, human subjects, legal and commercial considerations/agreements, partnerships or data with a high level of risk associated with it.</p>",
              "answer": "<p>test</p>"
            },
            {
              "title": "Question 2",
              "question": "<p>What data will you collect or otherwise bring into your project under this plan?</p>\r\n<p><br>Describe the data that will be collected, generated, and/or acquired.</p>",
              "answer": "<p>test</p>"
            },
          ]
- Questions with no answers are not included in the response.

### DIFF
--- a/app/controllers/api/v2/plans_controller.rb
+++ b/app/controllers/api/v2/plans_controller.rb
@@ -4,6 +4,7 @@ module Api
   module V2
     class PlansController < BaseApiController # rubocop:todo Style/Documentation
       respond_to :json
+      before_action :set_complete_param, only: %i[show index]
 
       # GET /api/v2/plans/:id
       def show
@@ -17,7 +18,6 @@ module Api
         raise Pundit::NotAuthorizedError unless plans_policy.show?
 
         @items = [@plan]
-        @complete = ActiveModel::Type::Boolean.new.cast(params[:complete])
 
         render '/api/v2/plans/index', status: :ok
       end
@@ -28,9 +28,12 @@ module Api
 
         @plans = PlansPolicy::Scope.new(@resource_owner).resolve
         @items = paginate_response(results: @plans)
-        @complete = ActiveModel::Type::Boolean.new.cast(params[:complete])
 
         render '/api/v2/plans/index', status: :ok
+      end
+
+      def set_complete_param
+        @complete = params[:complete].to_s.downcase == 'true'
       end
     end
   end
--- a/app/controllers/api/v2/plans_controller.rb
+++ b/app/controllers/api/v2/plans_controller.rb
@@ -17,6 +17,8 @@ module Api
         raise Pundit::NotAuthorizedError unless plans_policy.show?
 
         @items = [@plan]
+        @question_and_answer = ActiveModel::Type::Boolean.new.cast(params[:question_and_answer])
+
         render '/api/v2/plans/index', status: :ok
       end
 
@@ -26,6 +28,8 @@ module Api
 
         @plans = PlansPolicy::Scope.new(@resource_owner).resolve
         @items = paginate_response(results: @plans)
+        @question_and_answer = ActiveModel::Type::Boolean.new.cast(params[:question_and_answer])
+
         render '/api/v2/plans/index', status: :ok
       end
     end
--- a/app/controllers/api/v2/plans_controller.rb
+++ b/app/controllers/api/v2/plans_controller.rb
@@ -17,7 +17,7 @@ module Api
         raise Pundit::NotAuthorizedError unless plans_policy.show?
 
         @items = [@plan]
-        @question_and_answer = ActiveModel::Type::Boolean.new.cast(params[:question_and_answer])
+        @complete = ActiveModel::Type::Boolean.new.cast(params[:complete])
 
         render '/api/v2/plans/index', status: :ok
       end
@@ -28,7 +28,7 @@ module Api
 
         @plans = PlansPolicy::Scope.new(@resource_owner).resolve
         @items = paginate_response(results: @plans)
-        @question_and_answer = ActiveModel::Type::Boolean.new.cast(params[:question_and_answer])
+        @complete = ActiveModel::Type::Boolean.new.cast(params[:complete])
 
         render '/api/v2/plans/index', status: :ok
       end
--- a/app/presenters/api/v2/plan_presenter.rb
+++ b/app/presenters/api/v2/plan_presenter.rb
@@ -4,9 +4,9 @@ module Api
   module V2
     # Helper class for the API V2 project / DMP
     class PlanPresenter
-      attr_reader :data_contact, :contributors, :costs
+      attr_reader :data_contact, :contributors, :costs, :complete_plan_data
 
-      def initialize(plan:)
+      def initialize(plan:, complete: false)
         @contributors = []
         return unless plan.present?
 
@@ -22,6 +22,8 @@ module Api
         end
 
         @costs = plan_costs(plan: @plan)
+
+        @complete_plan_data = fetch_all_q_and_a if complete
       end
 
       # Extract the ARK or DOI for the DMP OR use its URL if none exists
@@ -34,6 +36,8 @@ module Api
         # if no DOI then use the URL for the API's 'show' method
         Identifier.new(value: Rails.application.routes.url_helpers.api_v2_plan_url(@plan))
       end
+
+      private
 
       # Fetch all questions and answers from a plan, regardless of theme
       def fetch_all_q_and_a
@@ -50,8 +54,6 @@ module Api
           }
         end
       end
-
-      private
 
       # Retrieve the answers that have the Budget theme
       def plan_costs(plan:)
--- a/app/presenters/api/v2/plan_presenter.rb
+++ b/app/presenters/api/v2/plan_presenter.rb
@@ -36,12 +36,12 @@ module Api
       end
 
       # Fetch all questions and answers from a plan, regardless of theme
-      def fetch_all_q_and_a # rubocop:disable Metrics/AbcSize,Metrics/CyclomaticComplexity
-        return [] unless @plan&.questions.present?
+      def fetch_all_q_and_a
+        return [] unless @plan.questions.present?
 
         @plan.questions.filter_map do |q|
           a = @plan.answers.find { |ans| ans.question_id == q.id }
-          next unless a.present? && !a.blank?
+          next unless a.present?
 
           {
             title: "Question #{q.number || q.id}",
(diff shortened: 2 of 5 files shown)


app/views/api/v2/plans/_show.json.jbuilder
```diff
@@ -57,6 +57,24 @@ unless @minimal
...
+        json.set! output.id.to_s do
+          json.array! q_and_a do |item|
+            json.title item[:title]
+            json.question item[:question]
+            json.answer item[:answer]
+          end
+        end
+      end
+    end
+  end
```
aaronskiba: I'm looking at the comment at the top of this file:
```
# Note the symbol of the dmproadmap json object
# nested in extensions which is the container for the json template object, etc.
# A JSON representation of a Data Management Plan in the
# RDA Common Standard format
```
With the exception of `json.extension`, I think all of the keys here map onto the [RDA-DMP-Common-Standard structure](https://github.com/RDA-DMP-Common/RDA-DMP-Common-Standard?tab=readme-ov-file#structure). Maybe this additional data should also be placed inside of `json.extension`?

app/views/api/v2/plans/_show.json.jbuilder
```diff
@@ -57,6 +57,24 @@ unless @minimal
 
   outputs = plan.research_outputs.any? ? plan.research_outputs : [plan]
 
+  if @question_and_answer
+    json.questions_and_answers do
+      outputs.each do |output|
+        presenter = Api::V2::ResearchOutputPresenter.new(output: output)
```
aaronskiba: Maybe you did this because `app/presenters/api/v2/research_output_presenter.rb` already has some other q_and_a methods, but I'm not sure about this approach.

It looks like `Api::V2::ResearchOutputPresenter` is currently only instantiated in `app/views/api/v2/datasets/_show.json.jbuilder`, which is responsible for the `dataset` section of the RDA DMP Common Standard structure. Because `fetch_all_q_and_a(plan:)` isn't being used for that section, I don't think we should define it there.

I worry about directly calling the private fetch_all_q_and_a method from here as well.

Since this view is already instantiating `Api::V2::PlanPresenter`, maybe defining `fetch_all_q_and_a` there might better fit the existing architecture?

momo3404: I agree, I initially kept `fetch_all_q_and_a` in the `research output presenter` since I thought that's where we'd want a function like that. I moved it to `plan_presenter`.

app/views/api/v2/plans/_show.json.jbuilder
momo3404: We'll probably have to get Marcus to sign off on what the schema should look like in the end, but I agree additional data like this should be inside the `extension` category. I moved it there.

app/presenters/api/v2/plan_presenter.rb
```diff
@@ -35,6 +35,22 @@ def identifier
         Identifier.new(value: Rails.application.routes.url_helpers.api_v2_plan_url(@plan))
       end
 
+      # Fetch all questions and answers from a plan, regardless of theme
+      def fetch_all_q_and_a(plan:) # rubocop:disable Metrics/AbcSize,Metrics/CyclomaticComplexity
```
aaronskiba: `plan` should be the same as `@plan`, which is already initialised in this presenter. We could simplify things here by removing the `plan:` parameter, and simply referencing `@plan`. It would also make it consistent with the rest of the class (e.g. `def identifier` uses `@plan`).

app/views/api/v2/plans/_show.json.jbuilder
```diff
@@ -68,5 +68,22 @@ unless @minimal
...
+      json.questions_and_answers do
+        outputs.each do |output|
+          q_and_a = presenter.send(:fetch_all_q_and_a)
+          next if q_and_a.blank?
+
+          json.set! output.id.to_s do
+            json.array! q_and_a do |item|
+              json.title item[:title]
+              json.question item[:question]
+              json.answer item[:answer]
```
aaronskiba: You gave the following as an example of the JSON response structure:
```json
      "questions_and_answers": {
        "15476": [
          {
            "title": "Question 1",
            "question": "<p>What considerations will you take into account with respect to ethical, legal, or commercial issues?</p>\r\n<p><br>Describe any applicable ethical, legal, or commercial considerations related to your project and data. This includes research involving Indigenous communities and knowledges, human subjects, legal and commercial considerations/agreements, partnerships or data with a high level of risk associated with it.</p>",
            "answer": "<p>test</p>"
          },
          {
            "title": "Question 2",
            "question": "<p>What data will you collect or otherwise bring into your project under this plan?</p>\r\n<p><br>Describe the data that will be collected, generated, and/or acquired.</p>",
            "answer": "<p>test</p>"
          },
        ]
      }
```

However, the following is a bit simpler and seems more consistent with the rest of the response structure:
```json
      "questions_and_answers":  [
          {
            "title": "Question 1",
            "question": "<p>What considerations will you take into account with respect to ethical, legal, or commercial issues?</p>\r\n<p><br>Describe any applicable ethical, legal, or commercial considerations related to your project and data. This includes research involving Indigenous communities and knowledges, human subjects, legal and commercial considerations/agreements, partnerships or data with a high level of risk associated with it.</p>",
            "answer": "<p>test</p>"
          },
          {
            "title": "Question 2",
            "question": "<p>What data will you collect or otherwise bring into your project under this plan?</p>\r\n<p><br>Describe the data that will be collected, generated, and/or acquired.</p>",
            "answer": "<p>test</p>"
          },
        ]
```
`plan.id` at least exists in some other places in the JSON response (i.e. `source: "GET /api/v2/plans/15476"` and `identifier: "http://localhost/api/v2/plans/15476"`).

momo3404: Just added the fix to remove the plan id from the JSON.

app/views/api/v2/plans/_show.json.jbuilder
momo3404: Marcus signed off on putting it into `extension`, and he also said the flag should be titled `complete` and the JSON should have `complete_plan`, like this:

```
        "complete_plan": [
            {
              "title": "Question 1",
              "question": "<p>What considerations will you take into account with respect to ethical, legal, or commercial issues?</p>\r\n<p><br>Describe any applicable ethical, legal, or commercial considerations related to your project and data. This includes research involving Indigenous communities and knowledges, human subjects, legal and commercial considerations/agreements, partnerships or data with a high level of risk associated with it.</p>",
              "answer": "<p>test</p>"
            },
            {
              "title": "Question 2",
              "question": "<p>What data will you collect or otherwise bring into your project under this plan?</p>\r\n<p><br>Describe the data that will be collected, generated, and/or acquired.</p>",
              "answer": "<p>test</p>"
            },
          ]
```          
I added those changes to this PR.

app/presenters/api/v2/plan_presenter.rb
```diff
@@ -35,6 +35,22 @@ def identifier
         Identifier.new(value: Rails.application.routes.url_helpers.api_v2_plan_url(@plan))
       end
 
+      # Fetch all questions and answers from a plan, regardless of theme
+      def fetch_all_q_and_a # rubocop:disable Metrics/AbcSize,Metrics/CyclomaticComplexity
+        return [] unless @plan&.questions.present?
+
+        @plan.questions.filter_map do |q|
+          a = @plan.answers.find { |ans| ans.question_id == q.id }
+          next unless a.present? && !a.blank?
```
aaronskiba: I think `a.present?` and `!a.blank?` are logically equivalent.

app/presenters/api/v2/plan_presenter.rb
```diff
@@ -35,6 +35,22 @@ def identifier
         Identifier.new(value: Rails.application.routes.url_helpers.api_v2_plan_url(@plan))
       end
 
+      # Fetch all questions and answers from a plan, regardless of theme
+      def fetch_all_q_and_a # rubocop:disable Metrics/AbcSize,Metrics/CyclomaticComplexity
+        return [] unless @plan&.questions.present?
```
aaronskiba: I'd just remove the safe navigation operator from @plan.

app/controllers/api/v2/plans_controller.rb
```diff
@@ -17,6 +17,8 @@ def show
         raise Pundit::NotAuthorizedError unless plans_policy.show?
 
         @items = [@plan]
+        @complete = ActiveModel::Type::Boolean.new.cast(params[:complete])
```
aaronskiba: Maybe we want to make this stricter? A lot of `params[:complete]`  values can result in `@complete` evaluating to true:
```
irb(main):001> ActiveModel::Type::Boolean.new.cast('x')
=> true
irb(main):002> ActiveModel::Type::Boolean.new.cast('False')
=> true
```

ActiveModel might be a bit overkill too. Something like `@complete = params[:complete].to_s.downcase == 'true'` might be sufficient.

You could even DRY things up with a before_action:
```ruby
before_action :set_complete_param, only: [:show, :index]
```

And then assign `@complete` within `def set_complete_param`.

momo3404: I just pushed that change, let me know if it looks right.

app/views/api/v2/plans/_show.json.jbuilder
```diff
@@ -68,5 +68,18 @@ unless @minimal
         json.title template.title
       end
     end
+
+    if @complete
```
aaronskiba, line 72: We've managed to simplify the `.json.jbuilder` file a good deal, but it might be best to have `PlanPresenter` own even more of the logic. Here's a suggestion on how to go about that:
```ruby
# app/views/api/v2/plans/_show.json.jbuilder
presenter = Api::V2::PlanPresenter.new(plan: plan, complete: @complete)
```

```ruby
# app/presenters/api/v2/plan_presenter.rb
attr_reader :data_contact, :contributors, :costs, :complete_plan_data
def initialize(plan:, complete: false)
  ...
  @complete_plan_data = fetch_all_q_and_a if complete
```

From there, `app/views/api/v2/plans/_show.json.jbuilder` would call `presenter.complete_plan_data` and `fetch_all_q_and_a` could be set as a private method.

PR #3595, opened 2026-02-06
Add complete plan flag to V2 API

momo3404:
Changes proposed in this PR:
- Add the `complete` flag to the V2 API to enable accessing all questions and answers belonging to a plan in the API response.
- Per the V2 API Documentation: "The idea is to have a flag which defaults to false and thereby is compliant with the RDA metadata standard, but can be passed in with a value of true to include the questions and answers of the plan."
- The flag is optional and works with the `api/v2/plans` and `api/v2/plans/[plan_id]` endpoints. For example: `api/v2/plans/12345?complete=true`.
- The schema the flag returns is as follows:
- ```
        "complete_plan": [
            {
              "title": "Question 1",
              "question": "<p>What considerations will you take into account with respect to ethical, legal, or commercial issues?</p>\r\n<p><br>Describe any applicable ethical, legal, or commercial considerations related to your project and data. This includes research involving Indigenous communities and knowledges, human subjects, legal and commercial considerations/agreements, partnerships or data with a high level of risk associated with it.</p>",
              "answer": "<p>test</p>"
            },
            {
              "title": "Question 2",
              "question": "<p>What data will you collect or otherwise bring into your project under this plan?</p>\r\n<p><br>Describe the data that will be collected, generated, and/or acquired.</p>",
              "answer": "<p>test</p>"
            },
          ]
- Questions with no answers are not included in the response.

### DIFF
--- a/app/controllers/api/v2/plans_controller.rb
+++ b/app/controllers/api/v2/plans_controller.rb
@@ -4,12 +4,13 @@ module Api
   module V2
     class PlansController < BaseApiController # rubocop:todo Style/Documentation
       respond_to :json
+      before_action :set_complete_param, only: %i[show index]
 
       # GET /api/v2/plans/:id
       def show
         raise Pundit::NotAuthorizedError unless @scopes.include?('read')
 
-        @plan = Plan.find_by(id: params[:id])
+        @plan = Plan.includes(roles: :user).find_by(id: params[:id])
 
         raise Pundit::NotAuthorizedError unless @plan.present?
 
@@ -27,6 +28,13 @@ module Api
         @plans = PlansPolicy::Scope.new(@resource_owner).resolve
         @items = paginate_response(results: @plans)
         render '/api/v2/plans/index', status: :ok
+      end
+
+      private
+
+      # GET /api/v2/plans?complete=true and  /api/v2/plans/:id?complete=true
+      def set_complete_param
+        @complete = params[:complete].to_s.downcase == 'true'
       end
     end
   end
--- a/app/controllers/api/v2/plans_controller.rb
+++ b/app/controllers/api/v2/plans_controller.rb
@@ -26,6 +26,7 @@ module Api
         raise Pundit::NotAuthorizedError unless @scopes.include?('read')
 
         @plans = PlansPolicy::Scope.new(@resource_owner).resolve
+        @plans = @plans.includes(answers: { question: :section }) if @complete
         @items = paginate_response(results: @plans)
         render '/api/v2/plans/index', status: :ok
       end
--- a/app/presenters/api/v2/plan_presenter.rb
+++ b/app/presenters/api/v2/plan_presenter.rb
@@ -68,6 +68,7 @@ module Api
           next unless q.present?
 
           {
+            id: q.id,
             title: "Question #{q.number || q.id}",
             section: q.section&.title,
             question: q.text.to_s,
--- a/app/presenters/api/v2/plan_presenter.rb
+++ b/app/presenters/api/v2/plan_presenter.rb
@@ -60,7 +60,7 @@ module Api
 
       # Fetch all questions and answers from a plan, regardless of theme
       def fetch_all_q_and_a
-        answers = @plan.answers.includes(question: [:section])
+        answers = @plan.answers
         return [] unless answers.present?
 
         answers.filter_map do |answer|
--- a/app/presenters/api/v2/plan_presenter.rb
+++ b/app/presenters/api/v2/plan_presenter.rb
@@ -4,9 +4,9 @@ module Api
   module V2
     # Helper class for the API V2 project / DMP
     class PlanPresenter
-      attr_reader :data_contact, :contributors, :costs
+      attr_reader :data_contact, :contributors, :costs, :complete_plan_data
 
-      def initialize(plan:)
+      def initialize(plan:, complete: false)
         @contributors = []
         return unless plan.present?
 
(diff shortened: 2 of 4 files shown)


app/views/api/v2/plans/_show.json.jbuilder
```diff
@@ -68,5 +68,18 @@ unless @minimal
         json.title template.title
       end
     end
+
+    if @complete
+      json.complete_plan do
+        q_and_a = presenter.complete_plan_data
+        next if q_and_a.blank?
+
+        json.array! q_and_a do |item|
```
aaronskiba, line 77: Up to you, but this might be a tiny bit easier to read:
```ruby
if @complete && presenter.complete_plan_data.present?
  json.complete_plan do
    json.array! presenter.complete_plan_data do |item|
```

app/presenters/api/v2/plan_presenter.rb
```diff
@@ -55,6 +57,24 @@ def plan_costs(plan:)
...
+      def fetch_all_q_and_a
+        answers = @plan.answers.includes(question: [:section])
+        return [] unless answers.present?
+
+        answers.filter_map do |answer|
+          q = answer.question
+          next unless q.present?
+
+          {
+            title: "Question #{q.number || q.id}",
```
aaronskiba, line 72: Did we end up agreeing on the unique identifier (i.e. `q.id`) for this?

momo3404: I think we agreed that it wasn't possible as is to have a unique identifier for a question that would remain unchanged if the question was edited, and that it was shippable as is.

aaronskiba: Sorry, maybe I'm misunderstanding, but it might be possible.
When I create a plan, it has a plan.template associated with it. And that template has several questions associated with it. That template should always be associated with that plan, and those questions should always be associated with that template.
If an admin were to update a template question, a new template would be created. However, no plans with the prior template or prior questions would be affected.

Despite all that, we're running out of time. :P You can leave this to ship as-is for now if you'd like.

app/controllers/api/v2/plans_controller.rb
```diff
@@ -28,6 +29,11 @@ def index
         @items = paginate_response(results: @plans)
         render '/api/v2/plans/index', status: :ok
       end
+
+      # GET /api/v2/plans?complete=true and  /api/v2/plans/:id?complete=true
+      def set_complete_param
```
aaronskiba, line 37: set_complete_param might be better as a private method.

app/presenters/api/v2/plan_presenter.rb
```diff
@@ -55,6 +57,24 @@ def plan_costs(plan:)
             currency_code: 'usd', value: answer.text }
         end
       end
+
+      # Fetch all questions and answers from a plan, regardless of theme
+      def fetch_all_q_and_a
+        answers = @plan.answers.includes(question: [:section])
```
aaronskiba: It's good you added the section here. We still have a bit of an N+1 issue here though.
We are first fetching all of the plans, and then fetching answers + the includes queries for each plan individually. We could optimize this further by pre-fetching the answers + questions + sections as part of the initial plans query (but only when we are in fact fetching all of the questions and answers).

There's still some more work to be done in the main API v2 branch, so we could merge this as is for now if we'd like and then address that with other related changes.

aaronskiba: Maybe the following would be an easy way to address it, though.

```ruby
# V2::PlansController#index
@plans = PlansPolicy::Scope.new(@resource_owner).resolve
@plans = @plans.includes(answers: { question: :section }) if @complete
@items = paginate_response(results: @plans)

# V2::PlanPresenter
def fetch_all_q_and_a
  answers = @plan.answers
```
I ran `curl -v -H "Authorization: Bearer X" http://127.0.0.1:3000/api/v2/plans?complete=true` followed by
`ab -n 10 -c 1   -H "Authorization: Bearer X"   "http://127.0.0.1:3000/api/v2/plans?complete=true"` before and after this change. It sped up the mean request time by close to 25%:
BEFORE:
```
Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.0      0       0
Processing:   618  729  99.7    725     884
Waiting:      618  729  99.7    725     884
Total:        618  729  99.7    725     884
```
AFTER:
```
Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.0      0       0
Processing:   471  561 112.2    529     860
Waiting:      471  561 112.2    529     860
Total:        471  561 112.2    530     860
```

app/presenters/api/v2/plan_presenter.rb
momo3404: Oh I see, I misunderstood. I can add the `q.id`

app/views/api/v2/plans/_show.json.jbuilder
```diff
@@ -68,5 +68,20 @@ unless @minimal
...
+          json.id item[:id]
+          json.title item[:title]
+          json.section item[:section]
+          json.question item[:question]
+          json.answer item[:answer]
```
aaronskiba: I like this idea. Maybe just having `json.question_id item[:id]` would help in clarifying that the id is in regard to the question itself. With the current response structure, it seems a bit ambiguous:
```json
  {
    "id": 69711,
    "title": "Question 1",
    "section": "Data Collection",
    "question": "What types of data will you collect, create, link to, acquire and/or record?",
    "answer": "<p>trying to udate this</p>"
  }
```

aaronskiba (approved):
This looks good. Feel free to merge.

PR #3597, opened 2026-02-09
Add Internal v2 API Access Token Generation for Users

aaronskiba:
# Changes proposed in this PR:

## Summary
This PR introduces support for issuing internal, user-scoped v2 API access tokens, managed entirely within the application (first-party tokens) for internal users.

## Key Changes
[Create internal Doorkeeper app via rake task](https://github.com/DMPRoadmap/roadmap/pull/3597/commits/e46519f5f36ba1631e8d0381d902dd9ec75649cb)
- Executed via `bundle exec rake doorkeeper:ensure_internal_app`
- Creates a first-party Doorkeeper client for issuing internal v2 API tokens
- Ensures the internal application exists in all environments before token service is used

[Create Api::V2::InternalUserAccessTokenService](https://github.com/DMPRoadmap/roadmap/pull/3597/commits/ccf0f928847631dd985f4f38286158ab2974d6b5)
- This service manages user-scoped v2 API access tokens for internal app users.
  - Tokens are equivalent to first-party Personal Access Tokens (PATs) and are issued directly to authenticated users, bypassing the full OAuth 2.0 authorization_code flow.
  - Supports token creation, rotation, and revocation.
  - Uses `Doorkeeper::AccessToken` records for consistent scoping, expiry, and revocation handling.
  - Designed strictly for internal usage; third-party OAuth clients are not supported.

[Add "POST /api/v2/internal_user_access_token" action & route](https://github.com/DMPRoadmap/roadmap/pull/3597/commits/797aeb619af043bb4190872b03c5cd176544ee26)
- Adds `Api::V2::InternalUserAccessTokensController#create` with Pundit authorization and routing. Also reuses the existing `users/refresh_token.js.erb` response to update the UI via JS.

- `@success` is read by `app/views/users/refresh_token.js.erb` (similar approach as `UsersController#refresh_token`)

[Add API v2 section to /users/edit#api-details](https://github.com/DMPRoadmap/roadmap/pull/3597/commits/965896b9c2f8ec197a395e2f6d3127795b84d781)
- This change updates `app/views/devise/registrations/_api_token.html.erb` to include support for the v2 API access token. Existing v0/v1 token support is retained.
  - Introduce V2 token lookup via `Api::V2::InternalUserAccessTokenService`
  - Display a dedicated V2 API access token section with its own regeneration action

[Expose API Access tab to all users / restrict legacy token rendering](https://github.com/DMPRoadmap/roadmap/pull/3597/commits/3782f0268fcdc597139c08e221e53696d88a4173)
- The API Access tab is now visible to all users to support the new v2 API token,
which is accessible to everyone.

[Add test coverage for internal v2 token generation](https://github.com/DMPRoadmap/roadmap/pull/3597/commits/8ef06e2c10f8cd1e436d9dde1a5af1e39d7e4d2c)
- Add request specs for InternalUserAccessTokensController
  - Include both authenticated & unauthenticated user scenarios
  - Include both present & absent internal OAuth app scenarios

- Add service specs for `InternalUserAccessTokenService`
  - Test token retrieval, rotation, and OAuth app presence
  - Verify old token revocation when rotating

- Add view specs for API token partials
  - Test legacy partial rendering based on `user.can_use_api?`
  - Test OAuth application availability scenarios

### DIFF
--- a/app/controllers/api/v2/internal_user_access_tokens_controller.rb
+++ b/app/controllers/api/v2/internal_user_access_tokens_controller.rb
@@ -1,0 +1,20 @@
+# frozen_string_literal: true
+
+module Api
+  module V2
+    # Controller for managing the current user's internal V2 API access token.
+    # Provides token rotation for authenticated internal users.
+    # See Api::V2::InternalUserAccessTokenService for token implementation details.
+    class InternalUserAccessTokensController < ApplicationController
+      # POST "/api/v2/internal_user_access_token"
+      def create
+        authorize current_user, :internal_user_v2_access_token?
+        @token = Api::V2::InternalUserAccessTokenService.rotate!(current_user)
+        @success = true
+        respond_to do |format|
+          format.js { render 'users/refresh_token' }
+        end
+      end
+    end
+  end
+end
--- a/app/controllers/api/v2/internal_user_access_tokens_controller.rb
+++ b/app/controllers/api/v2/internal_user_access_tokens_controller.rb
@@ -9,7 +9,7 @@ module Api
       # POST "/api/v2/internal_user_access_token"
       def create
         authorize current_user, :internal_user_v2_access_token?
-        @token = Api::V2::InternalUserAccessTokenService.rotate!(current_user)
+        @v2_token = Api::V2::InternalUserAccessTokenService.rotate!(current_user)
         @success = true
         respond_to do |format|
           format.js { render 'users/refresh_token' }
--- a/app/services/api/v2/internal_user_access_token_service.rb
+++ b/app/services/api/v2/internal_user_access_token_service.rb
@@ -21,9 +21,7 @@ module Api
     # This service does NOT support third-party OAuth clients or delegated consent flows.
     class InternalUserAccessTokenService
       READ_SCOPE = 'read'
-      APPLICATION = Doorkeeper::Application.find_by(
-        name: Rails.application.config.x.application.internal_oauth_app_name
-      )
+      INTERNAL_OAUTH_APP_NAME = Rails.application.config.x.application.internal_oauth_app_name
 
       class << self
         def for_user(user)
@@ -46,7 +44,26 @@ module Api
           )
         end
 
+        # Used by views (e.g. devise/registrations/_v2_api_token.html.erb) to safely
+        # gate token UI if the internal OAuth application is missing.
+        def application_present?
+          application!
+          true
+        rescue StandardError => e
+          Rails.logger.error(e.message)
+          false
+        end
+
         private
+
+        def application!
+          Doorkeeper::Application.find_by(name: INTERNAL_OAUTH_APP_NAME) ||
+            raise(
+              StandardError,
+              "Required Doorkeeper application '#{INTERNAL_OAUTH_APP_NAME}' not found. " \
+              'Please ensure the application exists in the database.'
+            )
+        end
 
         def revoke_existing!(user)
           Doorkeeper::AccessToken.revoke_all_for(application!.id, user)
--- a/app/services/api/v2/internal_user_access_token_service.rb
+++ b/app/services/api/v2/internal_user_access_token_service.rb
@@ -24,24 +24,16 @@ module Api
       INTERNAL_OAUTH_APP_NAME = Rails.application.config.x.application.internal_oauth_app_name
 
       class << self
-        def for_user(user)
-          Doorkeeper::AccessToken.find_by(
-            application_id: application!.id,
-            resource_owner_id: user.id,
-            scopes: READ_SCOPE,
-            revoked_at: nil
-          )
-        end
-
         def rotate!(user)
           revoke_existing!(user)
 
-          Doorkeeper::AccessToken.create!(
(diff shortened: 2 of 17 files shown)


app/controllers/api/v2/internal_user_access_tokens_controller.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+module Api
+  module V2
+    class InternalUserAccessTokensController < ApplicationController
```
momo3404, line 8: Nitpick, but the class title has `Internal User Access Token*s*` which doesn't match `internal_user_access_token`, would be good to just standardize it.

app/controllers/api/v2/internal_user_access_tokens_controller.rb
```diff
@@ -0,0 +1,16 @@
...
+      # POST "/api/v2/internal_user_access_token"
+      def create
+        authorize current_user, :internal_user_v2_access_token?
+        @token = Api::V2::InternalUserAccessTokenService.rotate!(current_user)
+        respond_to do |format|
+          format.js { render 'users/refresh_token' }
```
momo3404, line 15: Would it be possible for this endpoint to be hit without JS?

app/services/api/v2/internal_user_access_token_service.rb
```diff
@@ -0,0 +1,60 @@
...
+    class InternalUserAccessTokenService
+      READ_SCOPE = 'read'
+      APPLICATION = Doorkeeper::Application.find_by(name: 'Internal v2 API Client')
+
+      class << self
+        def for_user(user)
+          Doorkeeper::AccessToken.find_by(user_token_filter(user))
+        end
+
+        def rotate!(user)
```
momo3404, line 27: Would it be possible for this to fail in the creation stage? If so, I'm just wondering if we have ample error handling for that scenario.

app/services/api/v2/internal_user_access_token_service.rb
```diff
@@ -0,0 +1,60 @@
...
+
+          Doorkeeper::AccessToken.create!(
+            user_token_filter(user)
+            .merge(expires_in: nil) # Overrides Doorkeeper's `access_token_expires_in`
+          )
+        end
+
+        private
+
+        def revoke_existing!(user)
```
momo3404, line 60: Just double checking, do revoked tokens remain in the DB?

momo3404:
There's two documentation related Rubocop errors 

`app/controllers/api/v2/internal_user_access_tokens_controller.rb:5:5: C: Style/Documentation: Missing top-level documentation comment for class Api::V2::InternalUserAccessTokensController.`
`app/services/api/v2/internal_user_access_token_service.rb:23:5: C: Style/Documentation: Missing top-level documentation comment for class Api::V2::InternalUserAccessTokenService.`

app/services/api/v2/internal_user_access_token_service.rb
aaronskiba: Yes, cleaning up expired tokens is something we'll have to address even beyond this PR.

app/controllers/api/v2/internal_user_access_tokens_controller.rb
aaronskiba: Good to nitpick, I certainly do my share...

I was a little unsure about how to do this.
```ruby
# config/routes.rb
resource :internal_user_access_token, only: :create
```
Here, the singular route `resource :internal_user_access_token` is meant to signal that this is managing "the user's token" (one at a time), not a collection of tokens. 

However, even for singular resources, the Rails convention is to pluralise controller names (e.g. `InternalUserAccessTokensController` rather than `InternalUserAccessTokenController`). I could override that (pretty easy to do), but even things like `Devise::SessionsController` (rather than `Devise::SessionController`) are used, even though a user only has one session at a time; so it seems like the pluralisation convention is a strong one.

app/services/api/v2/internal_user_access_token_service.rb
aaronskiba: I will at least add some handling in case the internal OAuth application is not found.

app/controllers/api/v2/internal_user_access_tokens_controller.rb
momo3404: I see, that makes sense. Seems easier to just keep it as is for now then.

app/controllers/api/v2/internal_user_access_tokens_controller.rb
aaronskiba: Yes, it's technically possible via a direct POST (but that would fail without proper authentication).

Our UI only initiates this action through JS. This is consistent with the legacy token refresh pattern, where JS rendering is implicit. To reuse `refresh_token.js.erb`, we need to explicitly add `format.js { render 'users/refresh_token' }`.

aaronskiba:
Some most recent commits are currently not showing in this PR.
https://github.com/DMPRoadmap/roadmap/commits/aaron/feature/v2-api-token-for-internal-users/

